PR #17, opened 2015-03-16
add registered channels to "Join Channel" optionbox

fafafaf:
query needs to be filtered for invite only and passworded channels, but i dont know where the server stores those flags.

### DIFF
--- a/lobbyserver.py
+++ b/lobbyserver.py
@@ -2002,6 +2002,11 @@ Thanks,\n\
                 jsonToSend = {"command": "social", "autojoin": channels}
                 self.sendJSON(jsonToSend)
 
+                # registered channels
+                query = QSqlQuery("select name from anope_db_ChannelInfo", self.parent.db)
+                while query.next():
+                    channels.append(str(query.value(0)))
+
                 # for GW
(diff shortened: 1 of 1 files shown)


lobbyserver.py
```diff
@@ -2002,6 +2002,11 @@ def command_hello(self, message):
                 jsonToSend = {"command": "social", "autojoin": channels}
                 self.sendJSON(jsonToSend)
 
+                # registered channels
+                query = QSqlQuery("select name from anope_db_ChannelInfo", self.parent.db)
```
Sheeo: Would need to select the anope database first, or prepend it to the table name: `anope.anope_db_ChannelInfo`

Sheeo:
Not sure where they're stored either, but I'm not sure we want to rely on the anope database for core lobby functionality either, or that we want this feature implemented without a sort of approval process of registered channels showing up first.

JeroenDeDauw:
Cool!

Am I right in assuming that at present it is not possible to actually join these channels from within the FAF client? `/join #channelname` is not doing anything.

Sheeo:
@JeroenDeDauw `/join` was added with https://github.com/FAForever/client/pull/26, but it hasn't made it to deployment yet, since the 0.11 release of the client has stalled for a while.

PR #21, opened 2015-03-20
Simplify assignment

JeroenDeDauw:


Sheeo:
Thanks for pointing me at this and doing a PR :) :+1:

All responsibility for networking is being centralized in other classes than `Player` though, so this definition is also going away,

Sheeo:
That is, for the network update, `GameConnection` will maintain all this kind of state.

PR #23, opened 2015-03-21
Don't create user accounts until they are validated

ChrisKitching:
Steal the very obvious idea everyone missed from here:
https://neosmart.net/blog/2015/using-hmac-signatures-to-avoid-database-writes/

This depends on changes to the PHP code, which isn't public.
@Sheeo, you can review the PHP side of this at /opt/dev/www/app/faf on devserver, commit 2a7a645

Rather more (a nonzero amount of) testing would be wise before this is deployed, but the basic idea should be pretty clear at this point...


Sheeo:
Looking good. Let's get together and do the nonzero-amount of testing when you have the time.

JeroenDeDauw:
Why is the PHP not public? It's easier for me to contribute to that than to the Python stuff.

ChrisKitching:
I'm not sure what the original reason was, but right now I don't think we should make it public: at the moment security by obscurity is about the only security we've got.

Sheeo:
Ideally we're moving away from the PHP scripts in question -- we're aiming for a django-based web-app for this kind of stuff. So I don't think it makes much sense to publicize it until we're at that point.

IDragonfire:
+1 for the cool idea ...
We should use a similar approch for the vault, instead of this:
https://github.com/FAForever/client/blob/develop/src/vault/__init__.py#L68

IDragonfire:
You need to remove the old query check:
https://github.com/ChrisKitching/server/blob/cleverregistration/src/lobbyconnection.py#L703
 ```
 query = QSqlQuery(self.parent.db)
                query.prepare("SELECT id FROM `login` WHERE LOWER(`login`) = ?")
                query.addBindValue(login.lower())
                if not query.exec_():
                    self.log.debug("Error inserting login %s", login)
                    self.log.debug(query.lastError())
                    self.sendReply("LOGIN_AVAILABLE", "no", login)
                    return

                if query.size() != 0:
                    self.log.debug("Login not available: %s", login)
                    self.sendReply("LOGIN_AVAILABLE", "no", login)
                    return
```

**Comment by IDragonfire** [comment]
I leave the test of create_account to you @ChrisKitching 
```
@mock.patch('src.lobbyconnection.QSqlQuery')
def test_handle_action_create_account(mock_query, fa_server_thread):
    fa_server_thread.sendJSON = mock.Mock()
    fa_server_thread.sendReply = mock.Mock()
    stream = mock.Mock()
    stream.readQString.side_effect = ['myLogin', 'me@mail.de', 'MyHash']

    mock_query.return_value.exec_.return_value = True
    mock_query.return_value.size.return_value = 0

    fa_server_thread.handleAction('CREATE_ACCOUNT', stream)
    fa_server_thread.sendReply.assert_called_once_with( \
        "LOGIN_AVAILABLE", "yes", 'myLogin')
    fa_server_thread.sendJSON.assert_called_once_with(\
        dict(command="notice", style="info",
             text="A e-mail has been sent with the instructions to validate your account"))
    
```

PR #24, opened 2015-03-21
Abstract SQL more

ChrisKitching:
Let's wrap the rather cryptic SQL library in a nicer API.

This is a WIP. To illustrate the new API I've refactored command_hello, as well as a couple of other random odds and sods.
dbhelper.py is the interesting bit that does all the work.

I'm uneasy about the implementation of singlerow_query. Naturally this needs tests (and to actually be executed) before merging.

@Sheeo: I've surely done something stupid here: please tell me what it is before I go too far in the wrong direction.

### DIFF
--- a/src/dbhelper.py
+++ b/src/dbhelper.py
@@ -1,0 +1,107 @@
+from PySide import QtSql
+from PySide.QtSql import QSqlQuery
+from passwords import DB_SERVER, DB_PORT, DB_LOGIN, DB_PASSWORD, DB_TABLE
+import logging
+
+# Object to wrap a database connection and provide some vaguely handy utilities.
+_logger = logging.getLogger(__name__)
+
+# Open the database connection. when the module is first touched...
+_db = QtSql.QSqlDatabase("QMYSQL")
+_db.setHostName(DB_SERVER)
+_db.setPort(DB_PORT)
+
+_db.setDatabaseName(DB_TABLE)
+_db.setUserName(DB_LOGIN)
+_db.setPassword(DB_PASSWORD)
+
+_db.setConnectOptions("MYSQL_OPT_RECONNECT=1")
+
+if not _db.open():
+    _logger.error(_db.lastError().text())
+    raise RuntimeError
+
+# A cache of prepared queries, keyed by SQL string. Allows reuse of prepared queries.
+_prepared_queries = {}
+
+def get_query_object(query_string, *query_params):
+    """
+    Get a QSqlQuery object configured for the given query string and parameters. Used by the
+    various query function to do common setup work.
+    """
+    if query_string in _prepared_queries:
+        query = _prepared_queries[query_string]
+    else:
+        query = QSqlQuery(_db)
+
+        # This ostensibly is very good for performance, and iterating resultsets backwards is a
+        # pretty strong antipattern at the best of times...
+        query.setForwardOnly(True)
+
+        query.prepare(query_string)
+        _prepared_queries[query_string] = query
+
+    # Bind the arguments to the query.
+    for i, arg in query_params:
+        query.addBindValue(arg)
+
+    return query
+
+
+def exec(query_string, *query_params):
+    """
+    Run a query for which you don't care about the returned result (such as an insert).
+     """
+    with get_query_object(query_string, query_params) as query:
+        if not query.exec_():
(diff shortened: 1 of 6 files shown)


src/dbhelper.py
```diff
@@ -0,0 +1,107 @@
...
+        _prepared_queries[query_string] = query
+
+    # Bind the arguments to the query.
+    for i, arg in query_params:
+        query.addBindValue(arg)
+
+    return query
+
+
+def exec(query_string, *query_params):
```
Sheeo: I guess we should go with `execute()` instead.

The only real potential problem though, is if someone does `from dbhelper import *`, in which case this definition will override the built-in `exec()`. But I suppose it is best practice to avoid built-in names.

src/dbhelper.py
```diff
@@ -0,0 +1,107 @@
...
+from passwords import DB_SERVER, DB_PORT, DB_LOGIN, DB_PASSWORD, DB_TABLE
+import logging
+
+# Object to wrap a database connection and provide some vaguely handy utilities.
+_logger = logging.getLogger(__name__)
+
+# Open the database connection. when the module is first touched...
+_db = QtSql.QSqlDatabase("QMYSQL")
+_db.setHostName(DB_SERVER)
+_db.setPort(DB_PORT)
```
Sheeo: While it's handy to open the connection at module-import time, it is a bit problematic in some cases:

 - When running tests, we don't want to contact the database, and ensuring that this code doesn't run becomes a problematic manner of managing every `import` statement in every file imported from tests, to ensure that you monkeypatch this module first.
 - When compiling the code with e.g. nuitka, it will try to run this code at compile-time.

It would be preferable to establish the connection on the first query, instead.

src/dbhelper.py
```diff
@@ -0,0 +1,107 @@
...
+_db.setPassword(DB_PASSWORD)
+
+_db.setConnectOptions("MYSQL_OPT_RECONNECT=1")
+
+if not _db.open():
+    _logger.error(_db.lastError().text())
+    raise RuntimeError
+
+# A cache of prepared queries, keyed by SQL string. Allows reuse of prepared queries.
+_prepared_queries = {}
```
Sheeo: I'm not sure it's sensible to cache the `QSqlQuery` objects themselves, as these are `PySide`-owned C++ objects.

Generally, I do think Qt or even the underlying database handles caching of prepared queries as well -- but I would need to investigate further to be sure. Did you check that out?

ChrisKitching: I did, and seemingly this is the optimal approach (it gets funkier when you want to use multiple threads, but we don't, so we can be slightly dumb). This relies on the assumption that these QSqlQuery are semantically equivalent to the ones in QT, and there's not some extra magic hiding somewhere, which seems unlikely?

Prepared statements are cached on the database-side, associated with database connections. Kill the connection and all your prepared statements get wiped out.
Here's our first screwup: We're always opening and closing connections. Given that we only have one thread, we may as well have one connection to rule them all, and have associated with it a prepared statement for everything we ever do.

However, when you do a prepareStatement call it leads to an IPC call to the database (to prepare the statement (or, if already cached, to do nothing)). This is the expense you conveniently dodge by caching the actual objects: one fairly costly IPC transaction per query. (On top of whatever overheads are associated with creating/destroying QSqlQuery objects).
Some guy on StackOverflow advocated this approach, so it *must* be right! :P

Sheeo: Where exactly are we opening and closing the database? From my findings, we keep one persistent connection. At least I cannot find any calls to `db.close()`.

If the IPC call actually blocks the entire eventloop, this would be bad -- but I'm not convinced that that is the case. Given that it does not, it is a cost we can live with in terms of latency, and it can potentially improve overall responsiveness of the server by allowing other events to be processed in the meantime.

Sheeo:
Before we do more work here, @Eximius has suggested we use [peewee](https://peewee.readthedocs.org/en/latest/index.html) as an ORM.

I'd like your opinion @ChrisKitching :)

IDragonfire:
ORM are nice, but in general you lose performacne and/or control.
From my point of view this has a very low priority ;)

Sheeo:
I think we are having this superseded by the FAForever/db repository.

PR #26, opened 2015-03-21
Improve installation and test running instructions

JeroenDeDauw:
Now understandable by people that like me don't do python dev
and have negative IQ when it comes to system administration


Sheeo:
Thanks for this @JeroenDeDauw. I added a commet clarifying that the PySide wheel is optional.

PR #29, opened 2015-03-21
Pass argument along as apparently intended

JeroenDeDauw:
Given the code was broken before, it might be unused. So perhaps it's better to remove it?


Sheeo:
This got fixed on the refactor/rating branch https://github.com/FAForever/server/blob/refactor/rating/src/abc/base_player.py#L49

PR #35, opened 2015-03-22
Delete empty file

JeroenDeDauw:


Sheeo:
Thanks, this was meant to actually contain som tests though :) I'll write those up in just a bit.

PR #36, opened 2015-03-22
Simplify replayArmy.isPlayer

JeroenDeDauw:
I got several questions while working on this:

* Where should tests be put? It seems like currently `tests/unit_tests` mirrors `src`. What about tests for code not in `src`? Also, is the intent to move everything into `src`?
* How many newlines should there be between methods? I'm seeing different amounts used in different files.
* How should classes be named? Should `replayArmy` not be `ReplayArmy`?
* How should test functions be named? Is what I did fine?
* Where do I best ask these kinds of questions? On a random PR like this one does not seem good. The `#faf_dev` IRC channel appears rather dead. Should it be on the `#development` slack channel? If so, won't questions get lot amidst all the commit notifications?

### DIFF
--- a/replays/replayArmy.py
+++ b/replays/replayArmy.py
@@ -18,7 +18,7 @@
 
 class replayArmy(object):
 
-    def __init__(self) :
+    def __init__(self):
 
         self.color = 0
         self.civilian = False
@@ -30,7 +30,6 @@ class replayArmy(object):
 
 
     def populate(self, infos):
-        #print infos
         armyInfo = dict(infos)
         self.color = armyInfo['PlayerColor']
         self.civilian = armyInfo['Civilian']
@@ -42,11 +41,7 @@ class replayArmy(object):
         
         
     def isPlayer(self):
-        
-        if (self.civilian == False and self.human == True) :
-            return True
-        else :
-            return False
+        return not self.civilian and self.human
         
     def __str__(self):
         return "%s : team (%i)" % (self.id, self.team)
--- a/tests/unit_tests/test_replayArmy.py
+++ b/tests/unit_tests/test_replayArmy.py
@@ -1,0 +1,22 @@
+from replays.replayArmy import replayArmy
+
+def test_civiliansAreNotPlayers():
+    army = replayArmy()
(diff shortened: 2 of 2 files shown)


Sheeo:
* I would like all python source to eventually reside in `src`, which is why the `tests` directory reflects this. Ideally tests do not reference modules residing in `src` using that prefix, however. This is currently accomplished by manipulating `sys.path` directly, but I'm looking for a more elegant solution.
* PEP-8 mandates 2 newlines between method definitions, and I try to follow this, but I don't consider it extremely important
* Classes should be `UpperCamelCase`, modules, and function names (class, instance and regular) should be all lower case with underscores if it helps clarity (I tend to use them)
* Naming of test functions isn't something we have a convention for, but I have been following the other convention regarding casing
* These kinds of questions are suitably placed right here on github, so we have a publicly visible record of discussion on this

Sheeo:
This particular package under test (`replays`) is something we should consider refining and moving to the [faftools](https://github.com/FAForever/faftools) repository longer term, but for the moment I'm fine with keeping it here and merging this PR as-is. I'll leave it unmerged for a bit so we can discuss the points laid out in the OP.

JeroenDeDauw:
> function names (class, instance and regular) should be all lower case with underscores if it helps clarity (I tend to use them)

So they_should_be_like_this and notLikeThis?

JeroenDeDauw:
> These kinds of questions are suitably placed right here on github, so we have a publicly visible record of discussion on this

Then I will create dedicated issues for them in the future. Hard to find when it's on a PR, and very unsuitable for longer discussion when there is use for that.

JeroenDeDauw:
> I'll leave it unmerged for a bit so we can discuss the points laid out in the OP.

You can merge, I'll see new comments posted here.

Sheeo:
> So they_should_be_like_this and notLikeThis?

Yes

>Then I will create dedicated issues for them in the future. Hard to find when it's on a PR, and very unsuitable for longer discussion when there is use for that.

Of course, issues are better. Putting a note about this in `README.md` is also a good idea.

PR #40, opened 2015-03-22
Remove tailing semicolons;

JeroenDeDauw:


Sheeo:
More like semi-C :) :+1:

PR #41, opened 2015-03-22
Optimized imports

JeroenDeDauw:
Ran script that removes unused imports, does some ordering and normalizes newlines


Sheeo:
When I made pycharm do this I had to re-import several modules afterwards. Which IDE are you using?

JeroenDeDauw:
charm

Sheeo:
Right, I guess it's a misconfiguration my end.

JeroenDeDauw:
I've seen PhpStorm fail when optimizing imports in some cases, though usually it works fine. Perhaps you encountered this with an older version?

PR #42, opened 2015-03-22
Remove dead code, some of which appears to be broken

JeroenDeDauw:


Sheeo:
These are not dead functions.

They override `QTcpServer::incomingConnection` to initialize objects responsible for handling said connection.

PR #43, opened 2015-03-22
Style fixes in tournamentServerThread

JeroenDeDauw:


tournament/tournamentServerThread.py
```diff
@@ -222,10 +225,11 @@ def sendReply(self, action, *args, **kwargs) :
...
                 self.log.exception("Something awful happened when sending reply !")  
-  
-    def done(self) :
+
+
+    def done(self):
         self.parent.removeUpdater(self)
         if self.socket:
-            self.socket.deleteLater()
+            self.socket.deleteLater()
```
Sheeo: Would you mind setting up your IDE to preserve and insert the newline at EOL? This is done through `Settings->Editor->Other->Ensure line feed at file end on save.`

Sheeo:
Save for the newline all is good here :+1:

tournament/tournamentServerThread.py
JeroenDeDauw: Sure. Can you merge and just add the line again via the GH web UI? Is quickest

Sheeo: I can rewrite your commit, merging with github causes merge-commits, which we try to avoid for small branches :)

PR #44, opened 2015-03-22
Boolean simplification and style improvements in proxy.py

JeroenDeDauw:


Sheeo:
I find these classes a bit confusing and somewhat needless. Essentially what this is is a mapping of a couple of players to a local proxy-port number.

We can represent all the data that these classes handle, and all the associated operations, with something of the form:

```
{{p1,p2}=0, {p2,p3}=1}
```

I find that the `Game` class may aswell contain methods to manipulate such a data structure directly, instead of delegating this somewhat simple task to two dedicated classes.

JeroenDeDauw:
Why did coveralls fail for the last commit? It just adds a comment...

Sheeo:
It only ran coverage for a4b7ad4, presumably due to timing of amendment to the PR.

What do you think about my suggestion?

JeroenDeDauw:
What you wrote sounds reasonable to me. I'm yet looking at that kind of stuff yet though, since I need better understanding of the overall picture first.

One note of caution though: having simple value objects is nice. Yes, you can represent them as maps, lists and whatnot, though often this is unwise. http://c2.com/cgi/wiki?PrimitiveObsession

Sheeo:
Yeah keeping the proxy class, but renaming it to `ProxyMap` is a good idea.

I'm not sure long-term I'd like to keep it though, as how the proxy works overall is quite confusing and brittle. Dedicating better-named domain objects to managing the proxy state is something I'd like to do.

JeroenDeDauw:
Well, merge then?

Sheeo:
You don't want to get the refactor we talked about into this PR?

JeroenDeDauw:
I'm actually unsure what refactor you are talking about now. In any case, it's not what I set out to do here.

PR #52, opened 2015-03-24
Refactor/proxymap

Sheeo:
Implements the changes described in #44.


Sheeo:
Right, the whole premise for this thing is potentially something we can fix -- there shouldn't be a reason for players to have to use the same local proxy port when connecting to eachother.

Will just merge this tomorrow and forget about it until future releases touching the proxy.

PR #53, opened 2015-03-27
Feat/test

IDragonfire:
I used vagrant to setup my development environment.
Tested on Win7, should also work on Mac and Linux.


.gitignore
```diff
@@ -6,3 +6,5 @@ logs/
 .coverage
 *.bak
 py3env
+.vagrant
+*.iml
```
Sheeo: Please configure your editor to insert these, if you would :)

requirements.txt
```diff
@@ -1,8 +1,7 @@
 trueskill
-quamash
 mock
-PySide
```
Sheeo: Indeed PySide doesn't have a binary pip distribution, but it is a requirement, so it should be in here.

PR #55, opened 2015-03-29
Test/modvault

IDragonfire:
I added some basic tests for the modvault ...

Open: Add Comment
Fixed: Vote on invalid mod produce an db query

### DIFF
--- a/README.md
+++ b/README.md
@@ -68,7 +68,7 @@ GPLv3. See the [license](license.txt) file.
 
 # Network Protocol
 
-WIP: JSON Protocol Overview based on [QStreamData](http://doc.qt.io/qt-5/qdatastream.html) (UTF-16, BigEndian)
+WIP: JSON Protocol Overview based on [QDataStream](http://doc.qt.io/qt-5/qdatastream.html) (UTF-16, BigEndian)
 
 ## Incoming Packages
 
--- a/README.md
+++ b/README.md
@@ -74,5 +74,7 @@ WIP: JSON Protocol Overview based on [QStreamData](http://doc.qt.io/qt-5/qdatast
 
 ###### Mod Vault
 
-* `{command: modvault, type: start}`
-* `{command: modvault, type: like, uid: <uid>}`
+* `{command: modvault, type: start}`: show the last 100 mods
+* `{command: modvault, type: like, uid: <uid>}`: check if user liked the mod, otherwise increase the like counter
+* `{command: modvault, type: download, uid: <uid>}`: notify server about an download (for downlaod counter), does not start the download
+* `{command: modvault, type: addcomment}`: not implemented
--- a/README.md
+++ b/README.md
@@ -65,3 +65,14 @@ Use `py.test` to execute the unit tests.
 # License
 
 GPLv3. See the [license](license.txt) file.
+
+# Network Protocol
+
+WIP: JSON Protocol Overview based on [QStreamData](http://doc.qt.io/qt-5/qdatastream.html) (UTF-16, BigEndian)
+
+## Incoming Packages
+
+###### Mod Vault
+
+* `{command: modvault, type: start}`
+* `{command: modvault, type: like, uid: <uid>}`
--- a/src/lobbyconnection.py
+++ b/src/lobbyconnection.py
@@ -2275,6 +2275,9 @@ Thanks,\n\
 
 
     def command_modvault(self, message):
+        if "type" not in message:
+            # TODO: message, exception if invalid
+            return
         type = message["type"]
         if type == "start":
             query = QSqlQuery(self.parent.db)
--- a/src/lobbyconnection.py
+++ b/src/lobbyconnection.py
@@ -2275,9 +2275,6 @@ Thanks,\n\
 
 
     def command_modvault(self, message):
-        if "type" not in message:
-            # TODO: message, exception if invalid
-            return
         type = message["type"]
         if type == "start":
             query = QSqlQuery(self.parent.db)
@@ -2373,9 +2370,13 @@ Thanks,\n\
             query.prepare("UPDATE `table_mod` SET downloads=downloads+1 WHERE uid = ?")
             query.addBindValue(uid)
             query.exec_()
+            # TODO: add response message
 
         elif type == "addcomment":
-            pass
+            # TODO: implement
+            raise NotImplementedError('addcomment not implemented')
+        else:
+            raise ValueError('invalid type argument')
 
     def prepareBigJSON(self, data_dictionary):
         """
--- a/src/lobbyconnection.py
+++ b/src/lobbyconnection.py
@@ -2275,15 +2275,16 @@ Thanks,\n\
 
 
     def command_modvault(self, message):
-        if "type" not in message:
-            # TODO: message, exception if invalid
-            return
-        type = message["type"]
+        type = ''
+        if "type" in message:
+            type = message["type"]
         if type == "start":
             query = QSqlQuery(self.parent.db)
             query.prepare("SELECT * FROM table_mod ORDER BY likes DESC LIMIT 0, 100")
             query.exec_()
-            if query.size() != 0:
+            if query.size() == 0:
+                self.sendJSON(dict(command="notice", style="error", text="no mods"))
+            else:
                 while query.next():
                     uid = str(query.value(1))
                     name = str(query.value(2))
@@ -2321,7 +2322,9 @@ Thanks,\n\
             query.addBindValue(uid)
             if not query.exec_():
                 self.log.debug(query.lastError())
-            if query.size() == 1:
+            if query.size() == 0:
+                self.sendJSON(dict(command="notice", style="error", text="invalid ui"))
+            else:
                 query.first()
                 uid = str(query.value(1))
                 name = str(query.value(2))
@@ -2364,6 +2367,8 @@ Thanks,\n\
                     query.addBindValue(uid)
                     query.exec_()
                     self.sendJSON(out)
+                else:
+                    self.sendJSON(dict(command="notice", style="error", text="You cannot like this"))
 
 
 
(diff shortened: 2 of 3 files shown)


README.md
```diff
@@ -65,3 +65,16 @@ Use `py.test` to execute the unit tests.
 # License
 
 GPLv3. See the [license](license.txt) file.
+
+# Network Protocol
+
+WIP: JSON Protocol Overview based on [QStreamData](http://doc.qt.io/qt-5/qdatastream.html) (UTF-16, BigEndian)
```
Sheeo: It's QDataStream :)

IDragonfire: xD, if that is all :D

src/lobbyconnection.py
```diff
@@ -2275,6 +2275,9 @@ def command_game_host(self, message):
 
 
     def command_modvault(self, message):
+        if "type" not in message:
+            # TODO: message, exception if invalid
+            return
```
Sheeo: Related to FAForever/client#187 in that if we don't reply here, the client may end up killing the users disk space.

This command isn't executed through the offending `SecondaryServer` class, but we should still ensure a request always gets a reply.

README.md
```diff
@@ -65,3 +65,16 @@ Use `py.test` to execute the unit tests.
 # License
 
 GPLv3. See the [license](license.txt) file.
+
+# Network Protocol
```
Sheeo: So regarding detailing the network protocol in `README.md`, I think it's better left to a seperate document (Perhaps the wiki?)

tests/unit_tests/test_FaServerThread.py
```diff
@@ -75,3 +75,72 @@ def test_command_game_host_calls_host_game_invalid_title(connected_socket,
...
+@mock.patch('src.lobbyconnection.Config')
+def test_mod_vault_start(mock_config, mock_query, connected_socket, mock_lobby_server):
+    server_thread = LobbyConnection(connected_socket, mock_lobby_server)
+    server_thread.sendJSON = mock.Mock()
+    mock_query.return_value.size.return_value = 1
+    mock_query.return_value.next = one_time_true
+    server_thread.command_modvault({'type': 'start'})
+    server_thread.sendJSON.assert_called_once()
+    # call, method:attributes, attribute_index
+    assert server_thread.sendJSON.mock_calls[0][1][0]['command'] == 'modvault_info'
```
Sheeo: Indexing into mock_calls like this is pretty voodoo, even if you state the parameters just above.

tests/unit_tests/test_FaServerThread.py
```diff
@@ -75,3 +75,72 @@ def test_command_game_host_calls_host_game_invalid_title(connected_socket,
...
+    if OneTimeTrue:
+        return False
+    OneTimeTrue = True
+    return True
+
+@mock.patch('src.lobbyconnection.QSqlQuery')
+@mock.patch('src.lobbyconnection.Config')
+def test_mod_vault_start(mock_config, mock_query, connected_socket, mock_lobby_server):
+    server_thread = LobbyConnection(connected_socket, mock_lobby_server)
+    server_thread.sendJSON = mock.Mock()
```
Sheeo: Initialization of LobbyConnection in a fixture useable by this module seems like a good idea at this point :)

tests/unit_tests/test_FaServerThread.py
```diff
@@ -75,3 +75,72 @@ def test_command_game_host_calls_host_game_invalid_title(connected_socket,
...
+    mock_query.return_value.size.return_value = 1
+    mock_query.return_value.next = one_time_true
```
Sheeo: You should be able to just do `mock_query.side_effect = [True, False]`

tests/unit_tests/test_FaServerThread.py
```diff
@@ -75,3 +75,72 @@ def test_command_game_host_calls_host_game_invalid_title(connected_socket,
...
+# TODO: implement it
+def test_mod_vault_addcomment(connected_socket, mock_lobby_server):
+    server_thread = LobbyConnection(connected_socket, mock_lobby_server)
+    server_thread.command_modvault({'type': 'addcomment'})
+
+def test_mod_vault_invalid_type(connected_socket,
+                                mock_lobby_server):
+    server_thread = LobbyConnection(connected_socket, mock_lobby_server)
+    server_thread.command_modvault({'type': 'DragonfireNegativeTest'})
+    # TODO: what to check?
```
Sheeo: The server could be friendly and tell the client that this command does not exist, as I mentioned above, we probably should do that.

tests/unit_tests/test_FaServerThread.py
```diff
@@ -75,3 +75,72 @@ def test_command_game_host_calls_host_game_invalid_title(connected_socket,
...
+
+def test_mod_vault_no_type(connected_socket,
+                           mock_lobby_server):
+    server_thread = LobbyConnection(connected_socket, mock_lobby_server)
+    server_thread.command_modvault({'invalidKey': None})
+    # TODO: what to check?
```
Sheeo: This message should not reach here, as it will be caught in the JSON dispatcher [here](https://github.com/FAForever/server/blob/develop/src/lobbyconnection.py#L2408)

Sheeo:
Overall they look nice :)

README.md
IDragonfire: I choice a file into the repo to have a protocol version for each version.
A wiki is central ...
If we have a file we can discuss stuff and people can with PR's update the file.

Sheeo: Okay, good points, I suggest we use `PROTOCOL.md` and link it from here.

IDragonfire: okay, but why separate it in several files? I love the linking :D

Sheeo: Because the `README.md` file would grow out of proportion :)

IDragonfire: and? We can add a TOC, e.g.
https://github.com/jnicklas/capybara#capybara

src/lobbyconnection.py
IDragonfire: okay, currently an error occure on the server and also nothing is send ...

tests/unit_tests/test_FaServerThread.py
IDragonfire: Have you a better idea? 
I comment the indezies ^^

tests/unit_tests/test_FaServerThread.py
IDragonfire: It is reachable with `{command: modvault}`
The dispatcher checks only if the "command_modvault" function/attribute exists ...
Not the content of the data/message

tests/unit_tests/test_FaServerThread.py
```diff
@@ -75,3 +75,72 @@ def test_command_game_host_calls_host_game_invalid_title(connected_socket,
...
+
+@mock.patch('src.lobbyconnection.QSqlQuery')
+@mock.patch('src.lobbyconnection.Config')
+def test_mod_vault_start(mock_config, mock_query, connected_socket, mock_lobby_server):
+    server_thread = LobbyConnection(connected_socket, mock_lobby_server)
+    server_thread.sendJSON = mock.Mock()
+    mock_query.return_value.size.return_value = 1
+    mock_query.return_value.next = one_time_true
+    server_thread.command_modvault({'type': 'start'})
+    server_thread.sendJSON.assert_called_once()
```
Sheeo: This actually ends up as a call, not an assertion.

README.md
Sheeo: Gotcha :)

tests/unit_tests/test_FaServerThread.py
Sheeo: Yes, I think the following would be better:

```
(response, ), _ = server_thread.sendJSON.call_args
assert response['command'] == 'modvault_info'
```

tests/unit_tests/test_FaServerThread.py
Sheeo: Use the following instead: `server_thread.sendJSON.assert_called_once_with(mock.ANY)`

tests/unit_tests/test_FaServerThread.py
IDragonfire: I am a noob in python mock...  That look sa way better ;)

IDragonfire:
That is also possible, but then It is hard to give a good "error message" ...
Not my decision ... I am only the server tester :p

Sheeo:
It's not, you can add arbitrary messages to `KeyError` and `ValueError`.

IDragonfire:
Okay, if that is the preferred way I will change this ;)

IDragonfire:
What is with success messages?
e.g.
https://github.com/FAForever/server/blob/develop/src/lobbyconnection.py#L2372

Sheeo:
What do you mean?

IDragonfire:
You said we send json automatic, based on the exception. What should we send if the operation is successful?

IDragonfire:
Like the linked code... Download tracked... 
Maybe command should return a json object or exception?

Sheeo:
I'm not sure in this specific case, would need to review the client code

IDragonfire:
What did you think about a general return value (JSON data) for the command methods?

tests/unit_tests/test_FaServerThread.py
IDragonfire: I am not able to use a fixture: 

```@pytest.fixture
def my_server_thread(connected_socket, mock_lobby_server):
    return LobbyConnection(connected_socket, mock_lobby_server)```
I get:

![image](https://cloud.githubusercontent.com/assets/1217999/6891322/df55b8c8-d6b9-11e4-8dac-b664a2dc2152.png)

Sheeo: Add the `@pytest.fixture` decorator to your 'fixture' :)

IDragonfire: Markdown Syntax error ...
`@pytest.fixtur` was their, but behind the three ``` 
Complete file:
https://gist.github.com/IDragonfire/d753606dc43344871966

IDragonfire: found it ...
py.test error messages are bad :(

IDragonfire:
ll stuff should now be fixed ;)

PR #61, opened 2015-03-30
Update readme to match changes in c752d33

JeroenDeDauw:


Sheeo:
Woops good catch @JeroenDeDauw, thanks!

PR #62, opened 2015-03-30
Tests for handle_action

IDragonfire:
CREATE_ACCOUNT tests are missing and waiting on #23
Does the email sending working?
I had encoding problems on my first mock test here:
`base64.b64encode(str(uid))`

### DIFF
--- a/README.md
+++ b/README.md
@@ -94,14 +94,19 @@ Can be combined !, e.g. `{command: social, teaminvite: <...>, friends: <..>}`
 
 ##### Misc
 
-* `{command: ask_session}`: response with an welcome commannd and an valid session (can be delyed)
+* [deprecated] `{command: ask_session}`: response with an welcome commannd and an valid session (can be delyed)
+* `{command: fa_state, state: <on|...>}`: notify the server if the game has launched or closed
+* `{command: ladder_maps, maps: <map_ids>}`: select user specific maps for the ladder
+* `{command: quit_team}`: Leave a team
+* `{command: accept_team_proposal, leader: <leader_name>}`: Accept Team Invitation
+* `{command: hello, version: <...>, login: <...>, password: <...>, unique_id: <...>, (session: <...>)}`: Accept Team Invitation
 
-## Stream
+##  Stream
 
 It is possible to send data directly as Stream to the server. The response is normally a JSON Packet.
 
-* `PING`: response with a `PONG`
-* `PONG`: internal state changed to ponged
+* [deprecated] `PING`: response with a `PONG`
+* [deprecated] `PONG`: internal state changed to ponged
 * `FA_CLOSED`: internal cleanup of the player
 * `UPLOAD_MOD, login, session, zipmap, infos, size, fileDaatas`: Upload a mod
 * `UPLOAD_MAP, login, session, zipmap, infos, size, fileDatas`: Upload a map
--- a/README.md
+++ b/README.md
@@ -95,3 +95,14 @@ Can be combined !, e.g. `{command: social, teaminvite: <...>, friends: <..>}`
 ##### Misc
 
 * `{command: ask_session}`: response with an welcome commannd and an valid session (can be delyed)
+
+## Stream
+
+It is possible to send data directly as Stream to the server. The response is normally a JSON Packet.
+
+* `PING`: response with a `PONG`
+* `PONG`: internal state changed to ponged
+* `FA_CLOSED`: internal cleanup of the player
+* `UPLOAD_MOD, login, session, zipmap, infos, size, fileDaatas`: Upload a mod
+* `UPLOAD_MAP, login, session, zipmap, infos, size, fileDatas`: Upload a map
+* `CREATE_ACCOUNT, login, email, password_hash`: Send an email with a registration link
--- a/tests/unit_tests/test_lobbyconnection.py
+++ b/tests/unit_tests/test_lobbyconnection.py
@@ -297,7 +297,7 @@ def test_handle_action_invalidData(fa_server_thread):
 @mock.patch('src.lobbyconnection.QSqlQuery')
 @mock.patch('src.lobbyconnection.QFile')
 @mock.patch('src.lobbyconnection.zipfile')
-def test_handle_action_upload_map(mock_zipfile, mock_qfile, mock_query, mock_config, fa_server_thread):
+def test_handle_action_upload_mod(mock_zipfile, mock_qfile, mock_query, mock_config, fa_server_thread):
     fa_server_thread.sendJSON = mock.Mock()
 
     stream = mock.Mock()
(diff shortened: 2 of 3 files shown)


README.md
```diff
@@ -92,4 +92,15 @@ Can be combined !, e.g. `{command: social, teaminvite: <...>, friends: <..>}`
 
 ##### Misc
 
-* `{command: ask_session}`: response with an welcome commannd and an valid session (can be delyed)
\ No newline at end of file
+* `{command: ask_session}`: response with an welcome commannd and an valid session (can be delyed)
+
+## Stream
+
+It is possible to send data directly as Stream to the server. The response is normally a JSON Packet.
```
Sheeo: We should add that this section of the API is deprecated

tests/unit_tests/test_lobbyconnection.py
```diff
@@ -255,3 +256,230 @@ def test_avatar_select_no_avatar(mock_query, fa_server_thread):
...
+# handle action
+def test_handle_action_ping(fa_server_thread):
+    fa_server_thread.sendReply = mock.Mock()
+    fa_server_thread.handleAction('PING', mock.Mock())
+    fa_server_thread.sendReply.assert_called_once_with('PONG')
+
+def test_handle_action_pong(fa_server_thread):
+    assert fa_server_thread.ponged == False
+    fa_server_thread.handleAction('PONG', mock.Mock())
+    assert fa_server_thread.ponged == True
```
Sheeo: I think this is a bad test.

`lobbyconnection.ponged` should for one, not exist (We should use the `SO_KEEPALIVE` option) and if it should exist it should be a timestamp.

We should either have a TODO here or redo the implementation now (I don't think it's high priority)

IDragonfire:
I am only a tester, would should I do?

tests/unit_tests/test_lobbyconnection.py
```diff
@@ -255,3 +256,230 @@ def test_avatar_select_no_avatar(mock_query, fa_server_thread):
...
+    fa_server_thread.handleAction('FA_CLOSED', mock.Mock())
+    fa_server_thread.player.setAction.assert_called_once_with('NOTHING')
+    fa_server_thread.player.gameThread.abort.assert_called_once_with()
+
+def test_handle_action_unknownData(fa_server_thread):
+    fa_server_thread.receiveJSON = mock.Mock()
+    stream = mock.Mock()
+    fa_server_thread.handleAction('CrazyThing', stream)
+    fa_server_thread.receiveJSON.assert_called_once_with(\
+        'CrazyThing', stream)
```
Sheeo: What exactly does this test?

tests/unit_tests/test_lobbyconnection.py
```diff
@@ -255,3 +256,230 @@ def test_avatar_select_no_avatar(mock_query, fa_server_thread):
...
+                        'author': '', 'ui_only': '', 'version': '',
+                        'small': '', 'big': ''})
+    stream.readQString.side_effect = ['', '', zipMap, infos, 0, mock.Mock()]
+
+    # fake no db entry exists
+    mock_query.return_value.size.return_value = 0
+
+    fa_server_thread.handleAction('UPLOAD_MOD', stream)
+    fa_server_thread.sendJSON.assert_called_once_with( \
+        dict(command="notice", style="info", text="Mod correctly uploaded."))
```
Sheeo: This test seems to assert that you may correctly upload a mod that has an empty name, no uid, no description etc...

Is that right?

tests/unit_tests/test_lobbyconnection.py
```diff
@@ -255,3 +256,230 @@ def test_avatar_select_no_avatar(mock_query, fa_server_thread):
...
+                        'version': ''})
+    stream.readQString.side_effect = ['', '', zipMap, infos, 0, mock.Mock()]
+
+    # fake no db entry exists
+    mock_query.return_value.size.return_value = 0
+
+
+    fa_server_thread.handleAction('UPLOAD_MAP', stream)
+    fa_server_thread.sendJSON.assert_called_once_with( \
+        dict(command="notice", style="info", text="Map correctly uploaded."))
```
Sheeo: Ditto

tests/unit_tests/test_lobbyconnection.py
IDragonfire: Correct... 
We have no checks... 
Only existing...

tests/unit_tests/test_lobbyconnection.py
IDragonfire: What happen if we receive invalid data.

IDragonfire: That raise no exception... 
We store that...

tests/unit_tests/test_lobbyconnection.py
```diff
@@ -255,3 +256,230 @@ def test_avatar_select_no_avatar(mock_query, fa_server_thread):
...
+             'map_type': '', 'battle_type': '', 'map_size': {'0': '', '1': ''},
+             'version': ''}
+
+    error_messages = {'name': 'No map name provided.',
+                      'description': 'No map description provided.',
+                      'max_players': 'No max players provided.',
+                      'map_type': 'No map type provided.',
+                      'battle_type': 'No battle type provided.',
+                      'map_size': 'No map size provided.',
+                      'version': 'No version provided.'}
```
Sheeo: This only tests whether the message contains they keys, we should also verify each of them individually.

IDragonfire: From my view we should keep that and switch to Web in the future....

Sheeo: We will be doing that.

tests/unit_tests/test_lobbyconnection.py
IDragonfire: Maybe we have checks in the database?

Sheeo:
`I had encoding problems on my first mock test here:
base64.b64encode(str(uid))` 

What kind of problems?

IDragonfire:
Exception:
Str does not support the buffer interface....

Sheeo:
Ah, that's a python3 thing. We'd need to do `base64.b64encode(str(uid).encode())`, but that's silly.

This is going away with @ChrisKitching's patch anyway, so it's fine.

tests/unit_tests/test_lobbyconnection.py
IDragonfire: THIS IS IMPORTANT:
over this way we can execute commands over the Stream API:
https://github.com/FAForever/server/blob/develop/src/lobbyconnection.py#L2413

JeroenDeDauw:
This contains a number of serious http://c2.com/cgi/wiki?ArrangeActAssert violations. Uncountably this additional of tests is great, though don't forget one can have tech debt in tests themselves (and they can induce rigidity).

Sheeo:
Yes, I'm aware of that, some of our other tests violate the rule too.

I was never a fan of strict rules though, sometimes introducing an extra test just for an additional assert is overkill.

JeroenDeDauw:
I am also not a fan of strict rules. Reality is too complex for a single rule to make sense all of the time.

Violation of AAA in places where it actually is bad seems to be a common pattern here though, hence me pointing it out once again.

Sheeo:
Could you mention a specific bad test and suggest improvement?

JeroenDeDauw:
`test_fa_state`

IDragonfire:
What should I improve and how?

IDragonfire:
PS.: the problem from my view are not the tests...  More the units that are not decoupled will...

PR #68, opened 2015-04-03
Remove hostLocalPort from Game

JeroenDeDauw:
I noticed the field was not defined and then that while it was written to,
nothing used it. So I removed the setter and the calls to it. Also removed
the unused constructor parameter. As far as I can tell, nothing was calling
the constructor with the full arg list, so no further changes are needed.


Sheeo:
Yes I removed all handling of the clients local ip/port on sight. Guess I missed some, thanks!

PR #73, opened 2015-04-03
Use "foo not in" instead of "not foo in"

JeroenDeDauw:


src/games/gamesContainer.py
```diff
@@ -88,7 +88,7 @@ def findGameByUuid(self, uuid):
 
     def addGame(self, game):
         """Add a game to the list"""
-        if not game in self.games :
+        if game not in self.games :
```
Sheeo: For this, please remove the prefix space before ":" as well :)

JeroenDeDauw: At that rate we'll still have them by the time protons start decaying. What's the point of fixing it here if there are another 100 of these in the file? Better to fix them all in one go IMO.

Sheeo: Just fixing them as we see them will reduce them slowly but surely, if you want to do a full scale commit for it that is fine too.

JeroenDeDauw:
I've added a commit that fixes those things for this file. And some other stuff. Apparently PyCharm has no option to just fix the spacing stuff?

Sheeo:
Will do this as we come by them.

PR #76, opened 2015-04-04
Fix extraneous whitespace around keywords

JeroenDeDauw:
autopep8 . -r --in-place --select E22


Sheeo:
Will do this as we come by them

PR #96, opened 2015-10-22
Pseudocode of possible solution of lobby connections

Crotalus:
The current problem with the lobby is that ConnectToPeer on lobby side seems to block connections completely, even the NAT processing ones.

If we instead establish connections first before we start ordering everyone to connect to eachother we will fix the problem with ppl using proxy connections when they still are able to connect.

The code in this PR isn't complete, I don't have a python work-environment so can't test the actual changes either. 

Would be great if someone (@Sheeo ;)) could take a look at this, should be a quick change. What's needed is to be able to get the result from each EstablishConnection() to the later called ConnectToPeer. Either by putting that data on each peer object? Or maybe change the arguments of the function.

### DIFF
--- a/server/gameconnection.py
+++ b/server/gameconnection.py
@@ -218,10 +218,17 @@ class GameConnection(Subscribable, GpgNetServerProtocol):
                 yield from self.ConnectToHost(self.game.host.game_connection)
                 self._state = GameConnectionState.CONNECTED_TO_HOST
                 self.game.add_game_connection(self)
+                connections = []
                 for peer in self.game.connections:
                     if peer != self and peer.player != self.game.host:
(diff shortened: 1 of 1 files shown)


Crotalus:
Hmm, I realize `await` is a python 3.5 feature and server currently using 3.4? Maybe there's a similar way of doing it, I have too little experience with asyncio though

ChrisKitching:
> server currently using 3.4

Not for long.

Sheeo:
We're moving to python 3.5 and away from PySide. Should be done with that pretty soon.

Regarding the connectivity stuff: The current implementation you're hacking about with isn't in actual use on the server right now. It _does_ work as well as the current one (checkout the master branch), but we're going with (co)TURN instead.

PR #97, opened 2015-10-22
Install cmake

Crotalus:
cmake was missing when trying to create a fa/server image from a fresh Docker install (Ubuntu server)

### DIFF
--- a/Dockerfile
+++ b/Dockerfile
@@ -2,7 +2,7 @@ FROM python:3.4.3
 
 RUN apt-get update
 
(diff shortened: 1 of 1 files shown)


Sheeo:
cmake shouldn't be required -- what step of the process was needing it?

Crotalus:
Very weird, tested again now with exactly the same ubuntu image and cmake not required now. Can't remember exact step but it compiled something Py*. I think I forgot apt-get update after the first ubuntu install, did it on the second one though.

PR #99, opened 2015-10-24
Add support for searching by replay id in vault

Crotalus:
Hmm, another untested shot from the hip.

Would this be something useful?


Sheeo:
This is the repo you want: https://github.com/FAForever/legacy-secondaryServer

Sheeo:
But no, I think changing semantics of "player" to mean either "player name" (str) or "replay id" (int) is a bad idea.

PR #112, opened 2016-01-24
Process 'Rehost'

micheljung:
Processes the new `Rehost` command and sends a `game_launch` to the requesting client, as soon as the previous game connection has been closed.

### DIFF
--- a/server/lobbyconnection.py
+++ b/server/lobbyconnection.py
@@ -883,7 +883,7 @@ Thanks,\n\
         asyncio.async(self.send_coop_maps())
 
     @timed()
-    def command_game_host(self, message):
+    def command_game_host(self, message, is_rehost=False):
         if not self.able_to_launch_game:
             raise ClientError("You are already in a game or haven't run the connectivity test yet")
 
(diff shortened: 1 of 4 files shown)


micheljung:
Successfully tested with local server/client/game

server/lobbyconnection.py
```diff
@@ -937,10 +937,14 @@ def launch_game(self, game, port, is_host=False):
         self.player.game = game
         self.player.game_port = port
 
+        args = ["/numgames " + str(self.player.numGames)]
```
Sheeo: This should be client side

Sheeo:
This will invalidate quite a few assumptions in `Game` currently. We need to create a new `Game` object with a fully fresh state to be sure we're not messing up something.

The rehost procedure should do this like the `game_host` function currently does (There's probably some refactoring to be gained there)

server/lobbyconnection.py
micheljung: The reason I made it server-side is to that it's client-independent. But it's ok for me to have it client side....

micheljung:
Correct me if I'm wrong, but in my PR the procedure IS using game_host, so I don't see your point. It should be exactly like when the client sent a game_host message.

Sheeo:
Right, I missed the part that constructs a game_host message and passes it to lobbyconnection.

I'm against the way this is done -- I'd prefer for the client to intercept the rehost gpg command and issue `game_host` by itself, it could include `rehost: true` in the message so the server can let other clients know that this is a rehosted game. How does that sound?

I want to eventually get rid of the parent-pointer that is gameconnection.lobbyconnection, so this is a step in the wrong direction regarding that.

Sheeo:
We agreed in slack to do this clientside, so I'm closing this now :+1: :)

PR #119, opened 2016-04-09
Send ID instead of name for host in gameinfo message

yorick-ne:
This, together with some clientside stuff should fix the announcements when a friend is hosting / playing a game.

### DIFF
--- a/server/games/game.py
+++ b/server/games/game.py
@@ -659,7 +659,7 @@ class Game(BaseGame):
             "sim_mods": self.mods,
             "mapname": self.map_folder_name,
             "map_file_path": self.map_file_path,
-            "host": self.host.login if self.host else '',
+            "host": self.host.id if self.host else -1,
             "num_players": len(self.players),
             "game_type": self.gameType,
             "max_players": self.max_players,


Sheeo:
I'd like to test this out on the test server before we deploy @yorick-ne -- other things may well break.

Let's get in touch saturday?

Sheeo:
@micheljung mind checking this out on test server?

micheljung:
ack, @Sheeo

Sheeo:
Sparsely what needs verification is the following:

 - Replay metadata continues to contain player name in the previous place -- introducing player ID in replay metadata too would be nice. (This ensures compatibility for other tools that use them / tools that don't have the ability to lookup id -> name)
 - The various displays of `game_info` information should be unchanged

yorick-ne:
Apparently the replay metadata is basicaly a copy of the accumulated game_info messages, so it wouldnt contain the hosts name without further adaption.
Maybe sending both name and id like 
`"host": self.host.login if self.host else '',
"host_id": self.host.id if self.host else -1,`
is better here, atleast it should have alot less potential to break stuff.

Sheeo:
That does seem like the way to go. To be even more proper we could send a full 'player object', but this changes the type as well.

Issue #142, opened 2016-05-28
Server misreporting gameInfo to clients


HardlySoftly:
Server is reporting 100s of closed state games with 0 players to the FAF client on startup, causing client side performance issues.

See https://github.com/FAForever/client/issues/348

HardlySoftly:
The number of bad gameInfo items being sent has grown since the report yesterday (seems like a bad trend)

Crotalus:
An uneducated guess would be that games gets state ENDED but never marked dirty so they doesn't get cleaned up in report_dirties() function.

Could timeout-games be the culprit? 
https://github.com/FAForever/server/blob/develop/server/games/game.py#L142

No dirty marking there

HardlySoftly:
Can confirm some seriously old games are getting reported to the client (like 3+ weeks old)

aeoncleanse:
Specifically, it was that 97 out of 205 registered games were older than 24 hours.

Crotalus:
> old ones with an **open** state that are causing issues.

Aren't they all closed, or?

aeoncleanse:
https://github.com/FAForever/client/blob/develop/src/replays/_replayswidget.py#L353

At fault is that function. Specifically, lines 440 and 441, cutting those dramatically cuts execution time. State has to be 'Playing' to get there.

HardlySoftly:
Here's a pastebin of the games I got sent by the server: http://pastebin.com/0ETTYnRT

Crotalus:
Hmm, seems like all games are closed, visibility is 'None' which is a bit weird, AFAIK it should be "public" as default.

HardlySoftly:
Old games building up again and causing issues, if we don't have a fix yet it may be time for the classic server restart before people stop being able to log in.

Crotalus:
Client was monkey patched but server bugfix has been lingering since ~1 month

HardlySoftly:
Client hasn't been patched in ages, and right now I'm counting 111 games from the server that didn't start in the last hour, and 90 which didn't start in the last 24 hours.

The last I heard we knew what the problem was, is there a hold up with deploying a fix?

HardlySoftly:
People have come to me saying they're getting the Outbound Traffic Blocked errors.

I have told them its a known issue, and I'm sure it will be resolved soon.

Sheeo:
Yes I'll reboot the server with Crotalus' patch tonight when we've got a lower user count.

HardlySoftly:
[...]

PR #144, opened 2016-05-30
Call on_game_end when a game timeouts

Crotalus:
Games ending with timeout never seem to get marked dirty, thus never garbage collected?

Proposed fix for #142

### DIFF
--- a/server/games/game.py
+++ b/server/games/game.py
@@ -140,7 +140,7 @@ class Game(BaseGame):
 
     def timeout_game(self):
         if self.state == GameState.INITIALIZING:
-            self.state = GameState.ENDED
+            self.on_game_end()
 
     @property
     def armies(self):
--- a/server/games/game.py
+++ b/server/games/game.py
@@ -136,11 +136,12 @@ class Game(BaseGame):
 
         self.mods = {}
         self._logger.info("{} created".format(self))
-        asyncio.get_event_loop().call_later(20, self.timeout_game)
+        asyncio.get_event_loop().create_task(self.timeout_game())
 
-    def timeout_game(self):
+    async def timeout_game(self):
+        await asyncio.sleep(20)
         if self.state == GameState.INITIALIZING:
-            self.on_game_end()
+            await self.on_game_end()
 
     @property
     def armies(self):
--- a/tests/unit_tests/test_game.py
+++ b/tests/unit_tests/test_game.py
@@ -91,6 +91,12 @@ async def test_game_end_when_no_more_connections(game: Game, mock_game_connectio
 
     game.on_game_end.assert_any_call()
 
+async def test_game_marked_dirty_when_timed_out(game: Game):
+    game.state = GameState.INITIALIZING
+    await game.timeout_game()
+    assert game.state == GameState.ENDED
+    assert game in game.game_service.dirty_games
+
 async def test_clear_slot(game: Game, mock_game_connection: GameConnection):
     game.state = GameState.LOBBY
     players = [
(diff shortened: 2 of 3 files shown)


micheljung:
Looks good to me, thank you for the fix.

Please add a unit test, too.

Sheeo:
Unclear if this solves the problem fully. Games can wind up in the initialized state and time out as well; we need to make sure this kills the game too.

Sheeo:
I've set a due date for v0.3 at June 6, 2016. @Crotalus can you have unit tests for this ready by then? Otherwise we should assign someone else who'll get it done by then

Crotalus:
Yup, np, I'll try to get it done later tonight!

Crotalus:
Hmm, noticed that `on_game_end` is a courotine and `eventloop.call_later` didn't like that.

My test should now be somewhat correct? But how to transform call_later to something that works in the game init file?

server/games/game.py
```diff
@@ -136,11 +136,12 @@ def __init__(self, id, game_service, game_stats_service,
 
         self.mods = {}
         self._logger.info("{} created".format(self))
-        asyncio.get_event_loop().call_later(20, self.timeout_game)
+        asyncio.get_event_loop().create_task(self.timeout_game())
 
-    def timeout_game(self):
+    async def timeout_game(self):
+        asyncio.sleep(20)
```
Sheeo: Simply calling `asyncio.sleep(20)` doesn't actually cause a sleep, you need to `await` it :)

tests/unit_tests/test_game.py
```diff
@@ -91,6 +91,12 @@ def test_add_game_connection_throws_if_not_lobby_state(game: Game, players, mock
 
     game.on_game_end.assert_any_call()
 
+async def test_game_marked_dirty_when_timed_out(game: Game):
```
Sheeo: So, testing this behavior requires actually making it not sleep inside of `timeout_game`, otherwise our test suite is going to be unnecessarily delayed. You can do this by mocking out `asyncio.sleep()` to be a null coroutine that immediately returns.

See the [`@patch` decorator](https://docs.python.org/3/library/unittest.mock.html#unittest.mock.patch)

Crotalus: Yeah, I wondered why the testing routine didn't do the sleep, now I see why :) I'll take a look

micheljung:
btw @Crotalus please rename branches like `feature/144-call-on-game-end`, this way the issue automatically gets moved by waffle.io and it's a step towards `git flow` (which you could also start using :))

Crotalus:
Thx, will do that in the future, not using waffle myself. Would `feature/142-xxx` work or is `feature/#142-xxx` needed? Read somewhere that it only accept non-hashed ids if the branch begins with it aka `142-xxx`.

tests/unit_tests/test_game.py
```diff
@@ -91,6 +92,13 @@ def test_add_game_connection_throws_if_not_lobby_state(game: Game, players, mock
 
     game.on_game_end.assert_any_call()
 
+@patch('asyncio.sleep', return_value=None)
+async def test_game_marked_dirty_when_timed_out(game: Game):
+    game.state = GameState.INITIALIZING
+    await game.timeout_game()
+    assert game.state == GameState.ENDED
+    assert game in game.game_service.dirty_games
+
```
Sheeo: In the future, please seperate `arrange`, `act` and `assert` parts of the test for readability. We try and follow this pattern elsewhere in the code. Otherwise -- nice job on this 👍

micheljung:
Both, `feature/142-xxx` and `feature/#142-xxx` work.

PR #145, opened 2016-05-30
Don't send games in the "closed" state to the client

ChrisKitching:
The client discards games in the ENDED or INITIALIZING states.
A recentish server change seems to have greatly increased the
number of such games (or perhaps simply changed how the server
is reporting them so the client is now being told about them).

Due to a stupidity in the client, this causes it to freeze for a
rather long time when it gets a huge pile of junk games dumped
on it on startup. We can save ourself some resources by not doing
so anyway, so let's do that.

We should also make the client not be stupid.

Fixes client/348 from the server direction.

### DIFF
--- a/server/game_service.py
+++ b/server/game_service.py
@@ -183,8 +183,20 @@ class GameService:
 
     @property
     def open_games(self):
+        """
+        Return all games that meet the client's definition of "not closed".
+        Server game states are mapped to client game states as follows:
+
+            GameState.LOBBY: 'open',
+            GameState.LIVE: 'playing',
+            GameState.ENDED: 'closed',
+            GameState.INITIALIZING: 'closed',
+
+        The client ignores everything "closed". This property fetches all such not-closed games.
+        :return:
+        """
         return [game for game in self.games.values()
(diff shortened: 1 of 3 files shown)


server/game_service.py
```diff
@@ -183,8 +183,20 @@ def live_games(self):
...
+            GameState.LIVE: 'playing',
+            GameState.ENDED: 'closed',
+            GameState.INITIALIZING: 'closed',
+
+        The client ignores everything "closed". This property fetches all such not-closed games.
+        :return:
+        """
         return [game for game in self.games.values()
-                if game.state == GameState.LOBBY]
+                if game.state == GameState.LOBBY or game.state == GameState.LIVE]
```
ChrisKitching: Note: open_games was not used prior to this patch.

ChrisKitching:
Aaactuallu don't do this. The client relies on receiving these messages to delete ended games. A more interesting solution is required (only broadcast closed games once).

ChrisKitching:
Actually actually, _do_ do this,

Sheeo:
Good work, thanks :)

PR #147, opened 2016-05-31
Some added tests

Crotalus:


Crotalus:
Yey, managed to break travis I think, pushed this branch to FAForever by mistake, deleted it and pushed to my fork branch instead (which this PR is based on). Now travis tries to grab a non-existing branch it seems.

tests/unit_tests/test_game.py
```diff
@@ -31,6 +30,27 @@ def test_instance_logging(game_stats_service):
...
+        ['Victory', Victory.SANDBOX, Victory.DEMORALIZATION, ValidityState.WRONG_VICTORY_CONDITION],
+        ['FogOfWar', 'none', 'explored', ValidityState.NO_FOG_OF_WAR],
+        ['CheatsEnabled', 'true', 'false', ValidityState.CHEATS_ENABLED],
+        ['PrebuiltUnits', 'On', 'Off', ValidityState.PREBUILT_ENABLED],
+        ['NoRushOption', 20, 'Off', ValidityState.NORUSH_ENABLED],
+        ['RestrictedCategories', 1, 0, ValidityState.BAD_UNIT_RESTRICTIONS]
+    ]
+
+    for data in settings:
+        game.gameOptions[data[0]] = data[1]
```
Sheeo: You should use a list of tuples rather than a list of lists, would allow you to write this instead:

```python
for data in settings:
    key, value, default, expected = data
    game.gameOptions[key] = value
    await game.validate_game_settings()
    assert game.validity is expected
...
```

tests/unit_tests/test_game.py
```diff
@@ -315,3 +377,19 @@ def test_hashing(game):
...
 
     assert game.players == frozenset(players)
+
+def test_victory_conditions():
+    conds = [("demoralization", Victory.DEMORALIZATION),
+             ("domination", Victory.DOMINATION),
+             ("eradication", Victory.ERADICATION),
+             ("sandbox", Victory.SANDBOX)]
+
+    assert (Victory.from_gpgnet_string(t[1]) == t[2] for t in conds)
```
Sheeo: This assertion will always be true, since you're asserting truthyness of a generator object.

Instead unfold the loop (And again; use a list of tuples):
```python
for string_value, enum_value in conds:
    assert Victory.from_gpgnet_string(string_value) == enum_value
```

tests/unit_tests/test_game.py
```diff
@@ -315,3 +378,18 @@ def test_hashing(game):
...
+
+def test_visibility_states():
+    states = [("public", VisibilityState.PUBLIC),
+              ("friends", VisibilityState.FRIENDS)]
+
+    assert (VisibilityState.from_string(t[1]) == t[2] and
```
Sheeo: Ditto

tests/unit_tests/test_game.py
Crotalus: Ah, that was a nice syntax, changed it!

Sheeo:
Nice job, thanks :)

tests/unit_tests/test_game.py
Sheeo: Yep, python tuples are neat. lua does support the roughly same with the use of `unpack()`

PR #149, opened 2016-06-03
Prune unused gameType field from Game objects

ChrisKitching:
Not to be confused with the gameType field in the DB, which is
actually the victory condition. This particular field is always
zero.

### DIFF
--- a/server/game_service.py
+++ b/server/game_service.py
@@ -169,7 +169,6 @@ class GameService:
             game = Game(**args)
         self.games[id] = game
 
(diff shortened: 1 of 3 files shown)


Sheeo:
Did we verify that this is actually unused in the client?

aeoncleanse:
Seems not to be, there's nothing in a grep for it.

ChrisKitching:
It's unused in the client, and since it's always zero if it was used we'd've had a bug anyway.

PR #164, opened 2016-06-23
Send complete player information for the connecting user

ChrisKitching:
This removes a state from the client's state machine: it
previously had to handle having a half-initialised player, then
getting the real information later.

The client can now initialise itself properly with "me" on
welcome. It also knows that it will get this before it gets any
player_info objects about anyone else, which comes in handy.


micheljung:
I like the idea, I do not like the syntax error and the missing unit tests :-)

PR #165, opened 2016-06-23
Updated GeoIP.dat from Debian Package: http://packages.debian.org/sid…

imangerah:
…/geoip-database

Certain IP Addresses from South Africa are displayed as US in the lobby. It appears to be due to an outdated GeoIP.dat file that is in use.

Attached is the current GeoIP.dat, a proposed GeoIP.dat that I have extracted from the debain geoip-database package and a python script to demonstrate the problem and the fix.

[geoip_fix_test.zip](https://github.com/FAForever/server/files/329670/geoip_fix_test.zip)


micheljung:
Thank you very much!

PR #170, opened 2016-06-27
Wait until host is ready before sending GpgNet stuff

Crotalus:
Make sure the host is ready before giving order to the other ladder player to connect

### DIFF
--- a/server/gameconnection.py
+++ b/server/gameconnection.py
@@ -165,7 +165,11 @@ class GameConnection(GpgNetServerProtocol):
         if not result:
             self.abort("Failed connecting to host {}".format(peer))
 
-        await asyncio.wait_for(self.game.joinable, None)
+        try:
+            await asyncio.wait_for(self.game.joinable, 60.0)
+        except asyncio.TimeoutError:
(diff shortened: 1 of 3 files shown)


server/gameconnection.py
```diff
@@ -163,10 +164,18 @@ def send_message(self, message):
         result = await self.EstablishConnection(peer)
         if not result:
             self.abort("Failed connecting to host {}".format(peer))
+
+        try:
+            await asyncio.wait_for(self.game.joinable, 60.0)
```
Sheeo: This is too late, you need this wait prior to establishing the connection with the host, or this will not have the desired effect.

Sheeo:
Want to do my amendment on this? :)

server/gameconnection.py
Crotalus: Yeah, I was thinking about that. Isn't EstablishConnection all about the connection setup which is separate from needing the actual game to be listening, it's all handled by the other players lobby-client?

When the connection handshake is done we need to wait for Zlo's slow ForgedAlliance.exe to actually be prepared for the JoinGame command.

PR #172, opened 2016-06-30
Send the channel list in the welcome message

ChrisKitching:
Makes client-side initialisation simpler.


Brutus5000:
@ChrisKitching: Are you working on the client side as well?
Looks like we just need to move the client-code in the client from handle_social to handle_welcome. But as you wrote "Makes client-side initialisation simpler." there is obviously something I did not notice yet.

PR #183, opened 2016-07-11
Improve the way the server reports bans and login failures

ChrisKitching:
In order to avoid localisation issues, we should avoid sending
strings to show in the UI over the protocol.

Here, the global exception handler around all message handlers
that listens for login failures is moved into the login function.
Its behaviour is changed to not send a string to the client.
Bans now report deltas, not naive timezone-ignorant expiry dates.
Unit tests for bans and invalid passwords have been added.

### DIFF
--- a/server/lobbyconnection.py
+++ b/server/lobbyconnection.py
@@ -66,9 +66,17 @@ class ClientError(Exception):
 
 
 class AuthenticationError(Exception):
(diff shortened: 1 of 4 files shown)


Sheeo:
Should redo without introducing an additional new client message and just reuse `authentication_failed`

Issue #200, opened 2016-09-06
Ingame swords still display in aeolus


Kalvirox:
The crossed swords still display between adjuton and I even though the game ended days ago. I have played other games since then and refafed several times but this has not solved the issue.

micheljung:
This should be fixed by #201

shalkya:
http://imgur.com/o3AKfHe
still happening

Geosearchef:
Still happening, on both clients, everytime I join, I get a message from the server containing this game: (which ended about a week ago)
faflive://lobby.faforever.com/55538999/73910.SCFAreplay?map=equilibrium.v0007&mod=phantomx

![alreadyingame](https://cloud.githubusercontent.com/assets/5748922/21547432/506f193a-cde6-11e6-8bb1-fc795c583959.PNG)

hugst:
Any update on this?

shalkya:
![image](https://user-images.githubusercontent.com/11618427/28332579-a0e36cc4-6bf5-11e7-92c9-a42ef6279b4c.png)
still happening. (just a reminder)

Askaholic:
Pretty sure this is fixed by #576

PR #201, opened 2016-09-06
Make sure game / players are dirtied when a game connection ends

Crotalus:
Make sure the cleanup of a game connection gets done regardless of eventual write exceptions to a peer.

### DIFF
--- a/server/gameconnection.py
+++ b/server/gameconnection.py
@@ -433,7 +433,10 @@ class GameConnection(GpgNetServerProtocol):
             if self.game.state == GameState.LOBBY:
                 for peer in self.game.connections:
                     if peer != self:
-                        peer.send_DisconnectFromPeer(self.player.id)
+                        try:
+                            peer.send_DisconnectFromPeer(self.player.id)
+                        except:
+                            pass
             self._state = GameConnectionState.ENDED
             self.loop.create_task(self.game.remove_game_connection(self))
             self._mark_dirty()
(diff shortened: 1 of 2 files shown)


micheljung:
Fixes #200

Crotalus:
Hopefully yes, only a theory from my side though while parsing the code. If someone with access to logs could check if the server sometime spits out exception in send_DisconnectToPeer (I guess lobbyconnection.send() at the lower level) that would be nice.

server/gameconnection.py
```diff
@@ -433,7 +433,10 @@ def abort(self, logspam=''):
             if self.game.state == GameState.LOBBY:
                 for peer in self.game.connections:
                     if peer != self:
-                        peer.send_DisconnectFromPeer(self.player.id)
+                        try:
+                            peer.send_DisconnectFromPeer(self.player.id)
+                        except:
```
Sheeo: We definitely don't want to just pass this and not at least log it :)

server/gameconnection.py
```diff
@@ -433,7 +433,10 @@ def abort(self, logspam=''):
             if self.game.state == GameState.LOBBY:
                 for peer in self.game.connections:
                     if peer != self:
-                        peer.send_DisconnectFromPeer(self.player.id)
+                        try:
+                            peer.send_DisconnectFromPeer(self.player.id)
+                        except Exception as ex: # pragma no cover
+                            self.log.exception("peer_sendDisconnectFromPeer failed for player %i", self.player.id)
```
Sheeo: Can we log the actual exception value, too?

Crotalus: I thought self.log.exception did this automatically? I may be wrong though

Crotalus: I guess it just calls  sys.exc_info() internally to get the exception?

micheljung: hm ok, maybe you're right then x)

PR #206, opened 2016-09-25
Unrank FFA, uneven teams and unreported games

### LINKED ISSUE #202: Games with uneven and FFA player count are rated (2v6)
iirc only games that had same number of players in the teams were rated. Now they are players that are abusing this to boost their rating a lot.
Can we get this fixed?

Crotalus:
Fixes #202 (Unranked when FFA or uneven teams)

### DIFF
--- a/server/games/game.py
+++ b/server/games/game.py
@@ -2,6 +2,7 @@ from enum import IntEnum, unique
 import logging
 import time
 import functools
+from collections import defaultdict
 import asyncio
 from typing import Union
 import trueskill
@@ -76,6 +77,7 @@ class ValidityState(IntEnum):
     COOP_NOT_RANKED = 11
     MUTUAL_DRAW = 12
     SINGLE_PLAYER = 13
+    FFA_NOT_RANKED = 14
 
 
 class GameError(Exception):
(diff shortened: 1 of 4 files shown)


micheljung:
Has this been tested, can we merge it for the next server update?

Crotalus:
AFAIK it works as intended, it passes both the old tests and the new specific ones aimed at FFA / uneven teams.

Sheeo:
I'd like to test this+upcoming fixes for the rating calculations on the test server.

Since my day-week is rather booked, can we schedule some time for this on saturday, with sunday as a last resort?

While I don't like postponing things to the day before push, I'm a bit out of options on time.

server/games/game.py
```diff
@@ -186,6 +189,45 @@ def teams(self):
...
                           for player in self.players})
 
+    def team_count(self):
+        teams = defaultdict(int)
+        for player in self.players:
+            teams[self.get_player_option(player.id, 'Team')] += 1
+
+        return teams
+
+    def even_teams(self):
```
Sheeo: For a predicate-like function I'd prefer to make it clear -- consider naming it just `is_even`? This would be consistent with your predicate-like `is_ffa` as well.

Perhaps even cleaner for the callsites would be to make these getter functions, so you can merely write `game.is_even` or `game.is_ffa`.

tests/unit_tests/test_game.py
```diff
@@ -52,6 +52,56 @@ def test_instance_logging(game_stats_service):
     await game.validate_game_settings()
     assert game.validity is ValidityState.VALID
 
+async def test_ffa_not_rated(game_service, game_stats_service):
+    game = Game(6, game_service, game_stats_service)
+    game.state = GameState.LOBBY
+    players = [
```
Sheeo: I'm aware I'm responsible for this duplication, but we should keep in mind that test code should be maintainable as well. A refactor into a reusable fixture of sorts would be much apreciated :)

Sheeo:
Well, aside from the failing tests--LGTM.

Crotalus:
Also added check if no game results were reported at all, in those cases do not persist the result.

Fixes #213 

(why the games don't get the game results is another question...)

Sheeo:
Nice work @Crotalus

PR #217, opened 2016-10-06
Feature/#211 adapt mod tables

micheljung:


tests/conftest.py
```diff
@@ -121,32 +121,12 @@ def opt(val):
                                                db=db))
     pool = loop.run_until_complete(pool_fut)
 
-    @asyncio.coroutine
```
Sheeo: Without this, now tests will not run in isolation with the default database state, so they cannot rely on fixtures defined. (Unless we mandate all tests never break them). While the unit tests still appear to run, it will be an annoying failure later on, so we should probably wrap the mysql session in a transaction for each test case and revert post-test run instead. I'll make an issue for this and defer so we don't block the merge.

PR #223, opened 2016-10-11
Bugfix/rating not persisted

Crotalus:
* Mutual draw was never triggered due to checking wrong index
* Add safety net around processing of stats if something goes wrong

Should be seen as general hardening of the code, not as a definite solution to the current rating problems (haven't seen any server logs of the latest problem)


Crotalus:
With the hotfix part of this PR is deprecated, will submit a new one

PR #232, opened 2016-10-23
Add server side support to update game title from lobby

Crotalus:


micheljung:
Can you add `Fixes #232` to the commit message so it automatically closes the related issue once we merge it? (for this and future issues)

PR #234, opened 2016-11-05
match the example private key the uid

muellni:
This key matches the example key in the uid repository and should let you set-up the server and connect to it with a self-compiles uid. See https://github.com/FAForever/uid#example-key-for-the-server


Sheeo:
Superceded with policy server

PR #236, opened 2016-11-06
Feature/#233 ice support

micheljung:
Support for ICE adapter. Breaks current client, but updated implementation is ready and has been tested.


Sheeo:
Looks good on the face of it. Can we get the updated client code pushed so we can test in unison prior to merging though?

muellni:
I created another PR for ICE: https://github.com/FAForever/server/pull/341

PR #243, opened 2016-11-17
Make mod listing more robust

### LINKED ISSUE #241: Server does not handle missing mod icon
https://github.com/FAForever/client/issues/490

duk3luk3:
When encountering unparseable mod in db, skip and log an error

Dunno if this is the best kind of change, but if we assume that the mod table is kind of open to shenanigans from people submitting mods, I think it's good to have it.

Fixes #241, but I still believe we should also get rid of NULL icons.

I looked at the tests, but how would I mock up a db row containing an invalid mod entry?


micheljung:
Empty string in a `NULL` column isn't invalid, therefore you can just insert a mod into the database in your test

PR #254, opened 2016-12-01
re-enable VM-check for new UID

muellni:
This is a totally untested change to re-enable the VM check for the new uid implementation.


server/lobbyconnection.py
```diff
@@ -594,10 +594,10 @@ def decodeUniqueId(self, serialized_uniqueid):
                 SerialNumber = str(machine.get('SerialNumber', 0)).encode() # BIOS serial number
                 VolumeSerialNumber = str(machine.get('VolumeSerialNumber', 0)).encode() # https://www.raymond.cc/blog/changing-or-spoofing-hard-disk-hardware-serial-number-and-volume-id/
```
micheljung: While we're at it, could we not give everybody looking at this code a hint on how to circumvent it? :P

muellni: :) I think that's a reminder that combining all features into a single hash is not a good idea. Feel free to remove this comment.

Sheeo:
Superceded by policy server

Issue #255, opened 2016-12-05
ladder game not reported right


GrotheFAF:
ladder game #5497983 is shown and counted like both players lost, but I actually won.
(the server stored replay look right, but shows no score at the end.)
[forever.zip](https://github.com/FAForever/server/files/630436/forever.zip)

micheljung:
```
mysql> select * from game_player_stats where gameid = 5497983;
+----------+---------+----------+----+---------+-------+------+-------+---------+-----------+------------+-----------------+-------+---------------------+
| id       | gameId  | playerId | AI | faction | color | team | place | mean    | deviation | after_mean | after_deviation | score | scoreTime           |
+----------+---------+----------+----+---------+-------+------+-------+---------+-----------+------------+-----------------+-------+---------------------+
| 10242620 | 5497983 |   112924 |  0 |       1 |     1 |    1 |     1 | 786.982 |   66.0333 |    789.799 |          65.582 |     0 | 2016-12-05 07:10:50 |
| 10242621 | 5497983 |    86430 |  0 |       1 |     2 |    1 |     2 | 865.812 |   66.6863 |     862.94 |         66.1922 |     0 | 2016-12-05 07:10:50 |
+----------+---------+----------+----+---------+-------+------+-------+---------+-----------+------------+-----------------+-------+---------------------+
```

micheljung:
Grothe the log says you got disconnected from the server twice:
```
2016-12-05 08:10:42,696 INFO     client._clientwindow           Outgoing JSON Message: {"args": [], "command": "EnforceRating", "target": "game"}
2016-12-05 08:10:42,742 INFO     client._clientwindow           Timeout/network error: The remote host closed the connection
2016-12-05 08:10:42,742 WARNING  client._clientwindow           Disconnected from lobby server.
2016-12-05 08:10:42,742 INFO     client._clientwindow           Reconnecting immediately
2016-12-05 08:10:42,742 WARNING  client._clientwindow           Disconnected from lobby server.
2016-12-05 08:10:42,742 INFO     client._clientwindow           Reconnecting immediately
2016-12-05 08:10:42,821 INFO     client._clientwindow           Outgoing JSON Message: {"version": "0.11.61-pre2-24-gcb14d90\n", "command": "ask_session", "user_agent": "faf-client"}
2016-12-05 08:10:42,835 INFO     fa.game_connection.GPGNetConnection GC >> : GameResult : [1, u'score 1']
[...]

PR #257, opened 2016-12-17
Bugfix/#255 stats bug

Crotalus:
The stats reporting testing mocked its own invalid results data so that's why this bug slipped through. I changed it so it setups with the real thing instead.

Also changed get_army_result to return the most reported result.

Fixes the stats reporting issue in #255.

### DIFF
--- a/server/games/game.py
+++ b/server/games/game.py
@@ -661,7 +661,16 @@ class Game(BaseGame):
         return score
 
     def get_army_result(self, player):
-        return self._results.get(self.get_player_option(player.id, 'Army'))
+        results = self._results.get(self.get_player_option(player.id, 'Army'))
+        if not results:
+            return None
+
+        counts = defaultdict(int)
+        for result in results:
+            if result[1] in ['victory', 'defeat', 'draw', 'mutual_draw']:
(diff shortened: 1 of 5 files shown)


Crotalus:
Travis seems unhappy, don't have time to investigate atm

```
Waiting for mysql container...
docker: Error response from daemon: driver failed programming external connectivity on endpoint faf-db (97e8e5015814a1ce88cbf6a88d3d8c13887e37dc1bb5be0f515ccddcd543de03): Error starting userland proxy: listen tcp 0.0.0.0:3306: bind: address already in use.
This is nc from the netcat-openbsd package. An alternative nc is available
in the netcat-traditional package.
usage: nc [-46bCDdhjklnrStUuvZz] [-I length] [-i interval] [-O length]
	  [-P proxy_username] [-p source_port] [-q seconds] [-s source]
	  [-T toskeyword] [-V rtable] [-w timeout] [-X proxy_protocol]
	  [-x proxy_address[:port]] [destination] [port]
<repeated...>
```

micheljung:
@Crotalus please see #262. You have write access to the server repo so you don't need to work on your fork :)

PR #260, opened 2016-12-23
Track user agent version

duk3luk3:
* Add user agent version to statd
* Add version to the user agent column in the login table
* Refactor statsd identifiers into constants

### DIFF
--- a/server/lobbyconnection.py
+++ b/server/lobbyconnection.py
@@ -663,10 +663,13 @@ Thanks,\n\
         else:
             return None
 
-    def statsd_gauge(self, delta):
-        vgauge = self.STATSD_VERSION.format(self.user_agent, self.version.replace('.','_'))
-        server.stats.gauge(vgauge, delta, delta=True)
-        server.stats.gauge(self.STATSD_USERAGENT.format(self.user_agent), delta, delta=True)
+    def update_stats(self, delta):
+        agauge = self.STATSD_USERAGENT.format(self.user_agent)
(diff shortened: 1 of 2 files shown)


server/lobbyconnection.py
```diff
@@ -653,18 +657,32 @@ def decodeUniqueId(self, serialized_uniqueid):
...
 
+    def clean_version(self):
+        if self.version is not None:
+            return self.version.split('-')[0].split('+')[0]
+        else:
+            return None
+
+    def statsd_gauge(self, delta):
+        vgauge = self.STATSD_VERSION.format(self.user_agent, self.version.replace('.','_'))
+        self._logger.debug('updating version gauge string: {} ({})'.format(vgauge, delta))
```
Sheeo: Let's avoid logging this now we know it works? :)

server/lobbyconnection.py
```diff
@@ -653,18 +657,32 @@ def decodeUniqueId(self, serialized_uniqueid):
...
 
+
         self.user_agent = message.get('user_agent')
-        version = message.get('version')
-        server.stats.gauge('user.agents.None', -1, delta=True)
-        server.stats.gauge('user.agents.{}'.format(self.user_agent), 1, delta=True)
+        self.version = message.get('version')
+        clean_version = self.clean_version()
+        # revert the "None" user agent set in servercontext
+        self.statsd_gauge(1)
```
Sheeo: Since we set `self.user_agent` above, this won't actually clean the `None` gauge? I presume the comment is wrong, taking the later changes into account.

PR #262, opened 2016-12-28
Bugfix/#255 stats bug

micheljung:
Replaces #257


Rackover:
Now here :
https://github.com/FAForever/server/pull/375

Issue #197, opened 2016-08-30
People can't join my Coop game


speed2CZ:
When hosting a coop game, no one can join me. It says for them that "Host has left the lobby"
My log https://gist.github.com/speed2CZ/ae0c750eea7ee3ce7d239c9cf408cf3e
theeggroll's log https://gist.github.com/theeggroll/1dc337d6a5c49436cb014df1cedce7f0

No problem when hosting a normal game

Sheeo:
eggroll's log only shows him successfully joining and launching a normal game. Your log shows some people trying to join but not succeeding, but the reason isn't clear. Please try and gather more logs.

speed2CZ:
ok, another try

My log https://gist.github.com/speed2CZ/9738669eb38746cbc552a2eb54c7dd54
theeggroll's log https://gist.github.com/theeggroll/0eec10b496b5ef33f7f9f5e7acd22f64
Typo's log https://gist.github.com/speed2CZ/d358fa5f2fafdd46917463510aed2cf5
WittySeal's log https://gist.github.com/speed2CZ/5fa4f90f99aea136a584739fdb1a4e7a

theeggroll:
People can't join my games period, ill be posting as many logs as I can get on this issue
FTXCommando trying to join my game: https://gist.github.com/anonymous/f48cf4139c2bf9a48f4837cc68bb83e1

HardlySoftly:
This happened to me, and there is nothing interesting in the log.

I suggest what we really need is server logs.

PR #269, opened 2017-02-04
Increase timeout for launching coop games

duk3luk3:
Coop games just take longer to launch, 20s timeout is not enough.

There is a substantial number of people who can never host coop because anyone trying to join gets "Host has left the game" error. After debugging this with speed2 it turned out his games timed out.

> 20:50:11 < duk3luk3> why would coop take longer to host?
> 20:50:32 < icedreamer> It does
> 20:50:33 < speed2> dunno, updating more files? game loading more data?
> 20:50:47 < icedreamer> When I launch coop locally it takes nearly 5 times longer than launching normal
> 20:50:58 < icedreamer> Both launching from files on my SSD
> 20:51:07 < icedreamer> Normal FA takes <1s
> 20:51:21 < icedreamer> My coop shortcut takes at least 4 or 5s to launch up

Fixes: #197

### DIFF
--- a/server/games/game.py
+++ b/server/games/game.py
@@ -146,7 +146,9 @@ class Game(BaseGame):
         return await asyncio.sleep(n)
 
     async def timeout_game(self):
-        await self.sleep(20)
+        # coop takes longer to set up
+        tm = 30 if self.game_mode != 'coop' else 60
+        await self.sleep(tm)
         if self.state == GameState.INITIALIZING:
             await self.on_game_end()
 


speed2CZ:
Fixes #197

duk3luk3:
Thanks, added that to the commit msg.

duk3luk3:
After jagged's comment in https://github.com/FAForever/client/issues/540 I've decided to also increase the default timeout to 30 seconds.

PR #281, opened 2017-03-18
Decrease cron data update interval

### LINKED ISSUE #280: Synchronise critical data more often
The server fetches some information only hourly:

https://github.com/FAForever/server/blob/develop/server/player_service.py#L26
https://github.com/FAForever/server/blob/61364ef820869265a3ea8416a3db1316df548efd/server/game_service.py#L51

This doesn't seem reasonable to me. It puts pretty much no load on the db. There's no difference between fetching these once per hour or once per minute.

And it's bad to push a bad update that can't be reverted until an hour later.

duk3luk3:
The server periodically fetches information like the list of featured
mods and the minimum client version from the database.

A sixty-minutes interval is not reasonable. This commit changes it to 10
minutes.

Fixes #280

### DIFF
--- a/server/game_service.py
+++ b/server/game_service.py
@@ -48,7 +48,7 @@ class GameService:
         loop = asyncio.get_event_loop()
         loop.run_until_complete(asyncio.async(self.initialise_game_counter()))
         loop.run_until_complete(loop.create_task(self.update_data()))
-        self._update_cron = aiocron.crontab('0 * * * *', func=self.update_data)
+        self._update_cron = aiocron.crontab('*/10 * * * *', func=self.update_data)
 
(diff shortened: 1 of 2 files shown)


duk3luk3:
@micheljung can this go into 0.8?

duk3luk3:
I am using a test account to test something, which I also need to make uniqueid_exempt so I can use it.

If I do that, I have to wait on average half an hour until the server updates the uid exempt users dict and I can actually use it.

This is extremely annoying to me.

micheljung:
Tried to merge from mobile; Server error. Please try again

duk3luk3:
Rebased, maybe it'll work better now :-)

PR #284, opened 2017-04-12
276 potentially fix zombie games

Wesmania:
It appears that in a scenario where a player leaves while we wait for the STUN connection, EstablishConnection ignores that and keeps trying to connect the player. If it succeeds, we add a GameConnection for a player that already left. We try to fix this behaviour here.


Wesmania:
I greatly simplified the fix - I believe the ConnectToHost method does not need any cleanup, since it just produces a pair of addresses.

micheljung:
is this ready for v0.8?

Wesmania:
I'm pretty sure it fixes the zombie game bug, and it's a prerequisite for #289, so sure.

PR #290, opened 2017-05-01
Add log formatting

duk3luk3:
Will add log level and logger name to log output.


duk3luk3:
@micheljung can we add this to v0.8?

PR #293, opened 2017-05-01
Track and remember game (sim) end

### LINKED ISSUE #289: Explicitly set endTime
Depends on FAForever/db#84.

`game_stats.endTime` needs to be set explicitly.
It's not easy for the server to know when a game actually finishes; right now it assumes a game to have finished when the last player disconnected. This causes #263. From my analysis, the best way to determine if a game finished is if either:

* All connected players reported army results for all players
* There's no more player connected

Even better may be to have the game send a `GameEnded` message, because the game end may actually not be coupled to the army result report (in fact, I think the current game code ends the game 5s after the last score report). But for now the above will suffice.

Wesmania:
Treat GameResult message in GameConnection as a signal that a player has
ended his sim. Track these ending sims in Game; once all remaining
GameConnections are marked as ended and the game is in a Live state,
mark the Game as (sim-wise) ended and set its end time.

Fixes #289.

Signed-off-by: Igor Kotrasinski <ik345186@students.mimuw.edu.pl>

### DIFF
--- a/server/gameconnection.py
+++ b/server/gameconnection.py
@@ -61,6 +61,8 @@ class GameConnection(GpgNetServerProtocol):
 
         self.connectivity = self.lobby_connection.connectivity  # type: Connectivity
 
+        self.finished_sim = False
+
     @property
     def state(self):
         """
@@ -334,6 +336,14 @@ class GameConnection(GpgNetServerProtocol):
                                    ['score', 'defeat', 'victory', 'draw'])):
                         raise ValueError()  # pragma: no cover
                     result = result.split(' ')
+
+                    # This is the most common way for the player's sim to end
+                    # We should add a reliable message to lua in the future
+                    if result[0] in ['victory', 'draw'] and not self.finished_sim:
+                        self.finished_sim = True
+                        await self.game.check_sim_end()
+
+
                     await self.game.add_result(self.player, army, result[0], int(result[1]))
                 except (KeyError, ValueError):  # pragma: no cover
                     self.log.warn("Invalid result for %s reported: %s", army, result)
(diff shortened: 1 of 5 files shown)


Wesmania:
We will probably need to update the way we set scoreTime in game_player_stats, too.

Wesmania:
What does the scoreTime column signify exactly, anyway? Is it the time the player sent his score for the game (i.e. ended his sim)?

micheljung:
`scoreTime` isn't relevant for fixing #289, is it? Let's think about this separately. Can you fix the build?

duk3luk3:
The game doesn't end when a `GameResult` message is sent. `GameResult` messages are sent every time an ACU dies.

You need to check at least if it's a 'Victory' or 'Draw' message.

Wesmania:
Roger, I'll check for that then.

Wesmania:
From what I see test failures come from mocks not having proper attributes / functions, and I'm having some trouble understanding the mock framework that's set up.

Wesmania:
The PR is pretty much ready except for the tests, which require mocking the DB connection. I can add it, though I'd like some input on that. I can either:

1. mock `db.db_pool` in the test (fastest, doesn't refactor anything, dirty),
2. add a `db_pool` argument to `Game` with `db.db_pool` as default and give mock DB pool in `Game` fixture (fast, low risk of errors, I don't know if it's good practice),
3. add a `db_pool` argument to `Game` without default, fix up all uses (slow, potential for errors or missing fixes, most correct).

micheljung:
I don't have time to look at the code and think myself into this, but;
1. Sounds OK for me, because we need it ASAP and can refactor it later
2. Sounds bad, because a `Game` should not know about a database pool (but I know that the server design is bad as well, so in that context it's probably not as bad)
3. See 2.

PR #299, opened 2017-05-15
Use `clan_tags` view of own database

IDragonfire:
Use `clan_tags` view of own database
`fafclans` database is deprecated

Based on https://github.com/FAForever/db/pull/117

### DIFF
--- a/server/player_service.py
+++ b/server/player_service.py
@@ -67,7 +67,7 @@ class PlayerService:
             try:
                 yield from cur.execute(
                     "SELECT `clan_tag` "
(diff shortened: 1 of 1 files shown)


IDragonfire:
Drop views and use direct queries on the server to have more control over the query,
potential performacne issue if the view use temptable algorithm:
https://dev.mysql.com/doc/refman/5.7/en/view-restrictions.html

Also because of https://github.com/FAForever/server/blob/a5adc6f441ea49b3e5167eed0e1414bb322af3fa/server/player_service.py#L71

PR #301, opened 2017-06-05
Support multiple UID Private Keys

duk3luk3:
To enable seamless UID Key rollover, multiple private keys can be
stored and used.


duk3luk3:
Tested on the test server, ready to go!

PR #303, opened 2017-06-06
Explicitly handle malformed encoded UIDs

### LINKED ISSUE #302: Unique id handling code does not handle malformed UIDs very well
Right now the function decodeUniqueId that validates the encoded UID logs an exception and implicitly returns None when something fails. This causes the validate_unique_id function to try and assign None to a pair of values, which causes and exception.

We should explicitly return None after an exception in decodeUniqueId and detect and handle it in validate_unique_id.

Wesmania:
Fixes #302.

### DIFF
--- a/server/lobbyconnection.py
+++ b/server/lobbyconnection.py
@@ -604,14 +604,20 @@ Thanks,\n\
 
             return m.hexdigest(), (UUID, mem_SerialNumber, DeviceID, Manufacturer, Name, ProcessorId, SMBIOSBIOSVersion, SerialNumber, VolumeSerialNumber)
         except Exception as ex:
-            self._logger.exception(ex)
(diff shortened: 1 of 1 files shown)


server/lobbyconnection.py
```diff
@@ -604,14 +604,20 @@ def decodeUniqueId(self, serialized_uniqueid):
 
             return m.hexdigest(), (UUID, mem_SerialNumber, DeviceID, Manufacturer, Name, ProcessorId, SMBIOSBIOSVersion, SerialNumber, VolumeSerialNumber)
         except Exception as ex:
-            self._logger.exception(ex)
```
duk3luk3: Please leave in logging for any catch-all exception handlers.

Wesmania:
Obsolete.

PR #307, opened 2017-07-03
[WIP] Reimplementation of the division system

Brutus5000:
- [X] Build a generic DivisionService that calculates scores and promotions
- [X] Add database persistence
- [ ] Create a DivisionService at server start and invoke it upon game results


server/games/ladder_game.py
```diff
@@ -19,7 +19,22 @@ def __init__(self, id, *args, **kwargs):
...
+            army = self.get_player_option(player.id, 'Army')
+            try:
+                if self.get_army_score(army) == 1:
+                    winner = player
+                if self.get_army_score(army) == 0:
+                    loser = player
+            except KeyError:
+                return
+
+            self.game_service.game_stats_service.process_ladder_scores(winner, loser, 1)
```
GrotheFAF: that is inside the player loop, is that intentional?
# ladder game 2 players (scores: 1,0):
loop 1. player: -> process_ladder_scores(winner, None(?), 1) 
loop 2. player: -> process_ladder_scores(winner, loser, 1)  <- same player as winner again

Brutus5000: No it was not intended. Of course that needs to be outside the for-loop.
But you are right about the draw. I should only proccess scores `if winner is not None`

GrotheFAF:
I got a new one: 2 winners  6472739  (replay shows one player get announced defeated, for no visable reason)

Brutus5000:
That's a weird server bug (affected 1500 games so far). I will not consider it here, because it needs to be fixed elsewhere.

GrotheFAF:
Not that weird, my guess is: the players lost connection to each other, maybe one lost internet but regained it before closing the game. So from both players faf view the other left and them won, and both report as such. I don't know if the server can detect such cases.

server/stats/division_service.py
```diff
@@ -0,0 +1,179 @@
...
+
+        winner = players[player_one] if winning_slot == 1 else players[player_two]
+        loser = players[player_two] if winning_slot == 1 else players[player_one]
+
+        if winner.is_in_inferior_league(loser):
+            gain = 1.5
+            loss = 1.0
+        elif winner.is_in_superior_league(loser):
+            gain = 0.5
+            loss = 0.5
```
GrotheFAF: That is so not in the wiki. (and as the difference gain-loss in the other case 0.5, this should be 0.0)

server/games/ladder_game.py
```diff
@@ -19,7 +19,23 @@ def __init__(self, id, *args, **kwargs):
...
+            try:
+                if self.get_army_score(army) == 1:
+                    winner = player
+                if self.get_army_score(army) == 0:
+                    loser = player
+            except KeyError:
+                return
+
+        if winner is not None:
+            self.game_service.game_stats_service.process_ladder_scores(winner, loser, 1)
```
GrotheFAF: your tests seem not to cover the case of: loser = none  (happens if both player have score = 1)
until this is 'fixed elsewhere', shouldn't it be covered?
currently - if it doesn't error - the random 2nd player gets the win.

server/stats/division_service.py
Brutus5000: I see 2 arguments against loss = 0.0:
Losing no points increases the chance that the player simply disconnects or Ctrl+Ks because he doesn't want to play against a more difficult enemy.
If you want to advance into a higher league, you need to be able to compete with them anyway. What's the point of getting into a league of players where you have no chance to defeat them?

server/games/ladder_game.py
Brutus5000: So what do we do best in this case? Don't process the scores at all?

GrotheFAF: a simple L37 ' and loser is not None' would fix that. (no points for both - acceptable specially if it's an exploit)

server/stats/division_service.py
GrotheFAF: The player will still lose rating.
And it's a simple systematic (against lower - equal - higher):
win: 0.5 - 1.0 - 1.5
lose: 1.0 - 0.5 - **0.0**

Brutus5000:
I do not plan to finish this.

PR #311, opened 2017-08-23
Send social info before games info

duk3luk3:
The information dump sent to the client after it logs in includes
friends and foes (`social` message) and currently hosted / running
games.

Sending the `social` message first means friend/foe information will be
available for the client so it can use it for coloring the hosted game
display.

Making this trivial change on the server avoids adding complex logic to
the client to update the display of games after social information is
received or holding back the display of games completely until that
point.

Fixes https://github.com/FAForever/client/issues/845


micheljung:
Looks good to me

Issue #320, opened 2017-11-01
Server is sending empty strings as countries for some players


Wesmania:
This is happening since last server migration.

GrotheFAF:
maybe https://github.com/FAForever/server/blob/develop/GeoIP.dat needs an update

MathieuB8:
Yeah I've looked in python file and you are most likely right. 
I've done a PR about that (didn't test it because I can't for the time being)

duk3luk3:
If we can install this file on the server, we should just map it as a volume instead of carrying it in the git repo.

MathieuB8:
Ugh..? You are thinking about docker I guess ? (no clue how to do what you said, never used docker before..)

PR #321, opened 2017-11-03
Missing country flag for users

MathieuB8:
By reading what GrotheFAF and when I looked into the lobbyconnection file, it seems like the country is found by IP using a database, the GeoIP file was ~1 year ago so I updated it. 
Unfortunately I couldn't test this change on my server so someone else will have to verify. But I don't see any reason why it shouldn't work.

Issue #320


micheljung:
Thanks, will go online with next server update on 18th Dec.

PR #325, opened 2017-11-18
Reject new accounts from logging in

duk3luk3:
**Needs Testing**

This change rejects server logins (similarly to when a user is banned) if:

* User is not linked to steam **and**
* Account is not at least 24h old

This is intended as a first-line measure to defend against griefers mass-creating accounts to avoid bans, because these accounts can be found and banned before the 24h period is over.


duk3luk3:
Tests that login with new accounts fail as expected, needs some adjustments to make accounts older (or fudge a steam id)

duk3luk3:
Tested on my test server, works as expected.

PR #341, opened 2018-03-22
ICE

muellni:
I did some cleanup of my previous aggregated ICE branch https://github.com/muellni/server/tree/feature/233-271-237-ICE


server/lobbyconnection.py
```diff
@@ -683,6 +683,28 @@ def check_version(self, message):
...
+            return
+
+        self._logger.debug("Restoring game session of player %s to game %s", self.player, game)
+        self.game_connection = GameConnection(self.loop, self, self.player_service, self.game_service, self.player,
+                                              game, GameConnectionState.CONNECTED_TO_HOST)
+        self.player.game_connection = self.game_connection
+        game.add_game_connection(self.game_connection)
+        self.player.state = PlayerState.PLAYING
+        if not hasattr(self.player, "game"):
+            self.player.game = game
```
Geosearchef: The player is now connected to this game, shouldn't we set it regardless of it having already a game attribute?

server/lobbyconnection.py
```diff
@@ -941,6 +930,22 @@ def command_modvault(self, message):
             else:
                 raise ValueError('invalid type argument')
 
+    @asyncio.coroutine
+    async def command_ice_servers(self, message):
+        if self.player:
+
+            ttl = 24*3600
+            ice_servers = self.coturn_generator.fetch_token(self.player.id, ttl)
```
Geosearchef: Do we wanna calculate the longterm-credential token anyways even if no static auth secret is defined and twilio is going to be used?

Rackover:
Superseded by https://github.com/FAForever/server/pull/411

PR #347, opened 2018-07-19
Add adjustment of matchmaking for new players

Exotic-Retard:
new players dont come in with an average of 1500 rating, unlike assumed
by trueskill. (we can see this in the rating distribution of the ladder
playerbase and notice that it doesnt match the distribution that
trueskill assumes)

so what we do to solve this, is to adjust their mean rating for
matchmaking purposes only (this is important) and let them play against
lower rated players. over the course of their first few games, the
adjustment decreases in strength and they begin to be matched according
to their actual rating.

ftx approved so i expect this to be merged within 12 seconds!

also its important to really carefully check this because i am incompetent.

### DIFF
--- a/server/matchmaker/search.py
+++ b/server/matchmaker/search.py
@@ -14,6 +14,9 @@ class Search:
     """
     Represents the state of a users search for a match.
     """
+    noobFixBaseRating = 500 # The base rating for a noob with 0 games
+    noobFixMaxGames = 10 # The number of games until rating should be applied fully
+
     def __init__(self, player, start_time=None, rating_prop='ladder_rating'):
         """
         Default ctor for a search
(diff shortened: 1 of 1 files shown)


server/matchmaker/search.py
```diff
@@ -40,9 +43,29 @@ def __init__(self, player, start_time=None, rating_prop='ladder_rating'):
             0: 0.8
         }
 
+        self.adjusted_rating = self.adjust_rating()
```
Sheeo: Rather than assigning it here you'll probably want to make `adjusted_rating` a computed proper like `rating`:

```python
@property
def adjusted_rating(self): ...
```

And inline the definition of `adjust_rating`.

server/matchmaker/search.py
```diff
@@ -14,6 +14,9 @@ class Search:
     """
     Represents the state of a users search for a match.
     """
+    noobFixBaseRating = 500 # The base rating for a noob with 0 games
```
Sheeo: Let's stick with `snake_case` and perhaps use a less offensive name? E.g. `newbie_base_mean` and `newbie_game_threshold`.

server/matchmaker/search.py
```diff
@@ -40,9 +43,29 @@ def __init__(self, player, start_time=None, rating_prop='ladder_rating'):
...
+        self.adjusted_rating = self.adjust_rating()
+
+    def adjust_rating(self):
+        if self.rating_prop=='ladder_rating':
+            numgames = self.player.numGames
+            if numgames <= self.noobFixMaxGames:
+                mean, dev = self.player.ladder_rating
+                # A simple linear interpolation
+                adjusted_mean = ((self.noobFixMaxGames - numgames) * self.noobFixBaseRating + numgames * mean) / self.noobFixMaxGames
+                self._logger.info('Adjusted mean rating for {player} with {numgames} games from {mean} to {adjusted_mean}'.format(
```
Sheeo: This logging statement will be very noisy -- it would be preferable to put it elsewhere, e.g. where the match is actually made here: https://github.com/FAForever/server/blob/develop/server/matchmaker/search.py#L125

Sheeo:
Otherwise looks good, thanks for the contribution :)

Wesmania:
IIRC this also needs a client-side fix, at least in the python client. Possibly in the java client too.

Rackover:
https://github.com/FAForever/server/pull/361

PR #348, opened 2018-07-21
Added more unrank conditions

Rackover:
… game

I couldn't run the server on my side to test it, but : 
- It is valid python code (it runs)
- The modification is pretty small, so it shouldn't be any trouble

If you have a python server working on your side, feel free to test the PR !

### DIFF
--- a/server/games/game.py
+++ b/server/games/game.py
@@ -600,9 +600,13 @@ class Game(BaseGame):
 
                 if not ranked:
                     await self.mark_invalid(ValidityState.BAD_MAP)
+            
+            else:
+                await self.mark_invalid(ValidityState.BAD_MAP)
 
             modId = self.game_service.featured_mods[self.game_mode].id
-
+            
+            
             # Write out the game_stats record.
             # In some cases, games can be invalidated while running: we check for those cases when
             # the game ends and update this record as appropriate.
--- a/server/games/game.py
+++ b/server/games/game.py
@@ -80,7 +80,8 @@ class ValidityState(IntEnum):
     FFA_NOT_RANKED = 14
     UNEVEN_TEAMS_NOT_RANKED = 15
     UNKNOWN_RESULT = 16
-
+    MULTI_TEAM = 17
+    
 
 class GameError(Exception):
     pass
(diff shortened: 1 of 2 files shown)


micheljung:
@Rackover will you write a unit test?

Rackover:
I don't know how to write a unit test, but the first cool thing would be to simply be able to test it on a local python server. Which for now, is not possible. 😢

server/games/game.py
```diff
@@ -80,7 +80,8 @@ class ValidityState(IntEnum):
     FFA_NOT_RANKED = 14
     UNEVEN_TEAMS_NOT_RANKED = 15
     UNKNOWN_RESULT = 16
-
+    MULTI_TEAM = 17
```
Sheeo: The database entries for these should be kept in sync

server/games/game.py
```diff
@@ -199,6 +200,15 @@ def teams(self):
         return frozenset({self.get_player_option(player.id, 'Team')
                           for player in self.players})
 
+    @property
+    def is_multi_team(self):
+        teams = set()
```
Sheeo: Couldn't this simply be `return len(self.teams) > 2`?

Sheeo:
We'll get you setup to test it locally; for now it's been out of sync with what's been going on in the db repo for a while, so it's hard to setup a local environment off the bat.

server/games/game.py
Rackover: I didn't know this var existed. Well if we can use it for that, then sure.

server/games/game.py
Rackover: you mean I must make a PR to the faf-db aswell ?

server/games/game.py
Sheeo: It's defined immediately above :)

server/games/game.py
Sheeo: Yes, as discussed in private.

Rackover:
https://github.com/FAForever/db/pull/162

server/games/game.py
```diff
@@ -199,6 +200,10 @@ def teams(self):
         return frozenset({self.get_player_option(player.id, 'Team')
                           for player in self.players})
 
+    @property
+    def is_multi_team(self):
+        return len(teams) > 2
```
Sheeo: Should be `len(self.teams) > 2`

Rackover:
https://github.com/FAForever/server/pull/357

PR #351, opened 2018-08-26
Forced steam linking for users registered after a certain date

Rackover:
Code runs, so it's valid python, but I'm still unable to set up Python server locally.

### DIFF
--- a/server/config.py
+++ b/server/config.py
@@ -53,3 +53,4 @@ API_TOKEN_URI = os.getenv("API_TOKEN_URI", "https://api.dev.faforever.com/jwt/au
 API_BASE_URL = os.getenv("API_BASE_URL", "https://api.dev.faforever.com/jwt")
 
 FAF_POLICY_SERVER_BASE_URL = os.getenv("FAF_POLICY_SERVER_BASE_URL", "http://faf-policy-server")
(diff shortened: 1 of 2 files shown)


Rackover:
Tested on a local copy on the python server, works as intended. Non-null STEAM_ID are allowed to connect anytime. NULL STEAM_ID are only able to connect if their create_time.timestamp() is above the config.NEW_ACCOUNTS_DATE_LIMIT limit.

micheljung:
The way you coded it, null steamIDs are only allowed to connect if their create time is **before** 1. sept.

micheljung:
I can merge this if you add a feature toggle. That is, an environment variable that allows us to enable/disable the check.

Can also be done by checking whether or not `NEW_ACCOUNTS_DATE_LIMIT` is set (which, btw., doesn't describe itself very well so I'd rename to `FORCE_STEAM_LINK_AFTER_DATE` or something better)

Rackover:
You mean a variable in config.py, or just an env variable check inside the lobbyconnection code ?
Doesn't that toggle already exist ? (by putting the forced steam link to an absurd date)

micheljung:
Right, setting it far into the future would also work (but that'd be a hack and I usually don't include hacks into my thinking ;) )

PR #353, opened 2018-09-03
Improve error messages of policy violations

micheljung:


server/lobbyconnection.py
```diff
@@ -507,22 +507,28 @@ def check_version(self, message):
         if response.get('result', '') == 'vm':
             self._logger.debug("Using VM: %d: %s", player_id, uid_hash)
             self.sendJSON(dict(command="notice", style="error",
-                               text="You need to link your account to Steam in order to use FAF in a Virtual Machine. "
-                                    "You can contact an admin on the forums."))
+                               text="You need to link your account to Steam in order to use FAF in a virtual machine. "
```
Brutus5000: I don't always add linebreaks into my popup messages, but when I do, I add `+` to concat the strings ;-)

server/lobbyconnection.py
```diff
@@ -507,22 +507,28 @@ def check_version(self, message):
...
-                                               "admin on the forums", fatal=True)
+                                               "admin or moderator on the forums", fatal=True)
             return False
 
         if response.get('result', '') == 'fraudulent':
             self._logger.info("Banning player %s for fraudulent looking login.", player_id)
             self.send_warning("Fraudulent login attempt detected. As a precautionary measure, your account has been "
-                              "banned permanently. Please contact a moderator if you feel this is a false positive.",
+                              "banned permanently. Please contact an admin or moderator on the forums if you feel this is "
+                              "a false positive.",
```
Brutus5000: same here

server/lobbyconnection.py
```diff
@@ -507,22 +507,28 @@ def check_version(self, message):
...
             self.sendJSON(dict(command="notice", style="error",
-                               text="You need to link your account to Steam in order to use FAF in a Virtual Machine. "
-                                    "You can contact an admin on the forums."))
+                               text="You need to link your account to Steam in order to use FAF in a virtual machine. "
+                                    "Please contact an admin or moderator on the forums if you feel this is a false positive."))
+            self.send_warning("Your computer seems to be a virtual machine.<br><br>In order to "
+                              "log in from a VM, you have to link your account to Steam: <a href='" +
+                              config.WWW_URL + "/account/link'>" +
+                              config.WWW_URL + "/account/link</a>.<br>If you need an exception, please contact an "
+                                               "admin or moderator on the forums", fatal=True)
```
Brutus5000: same here

server/lobbyconnection.py
```diff
@@ -507,22 +507,28 @@ def check_version(self, message):
...
 
         if response.get('result', '') == 'already_associated':
             self._logger.warning("UID hit: %d: %s", player_id, uid_hash)
             self.send_warning("Your computer is already associated with another FAF account.<br><br>In order to "
-                              "log in with a new account, you have to link it to Steam: <a href='" +
+                              "log in with an additional account, you have to link it to Steam: <a href='" +
                               config.WWW_URL + "/account/link'>" +
                               config.WWW_URL + "/account/link</a>.<br>If you need an exception, please contact an "
-                                               "admin on the forums", fatal=True)
+                                               "admin or moderator on the forums", fatal=True)
```
Brutus5000: same here

server/lobbyconnection.py
micheljung: Except in Python where this is optional and thus redundant :-)

micheljung: when you have a string following a string, the `+` is redundant. But if you have a string followed by a variable, the `+` is mandatory.

So let's say you only have constants multiline strings, w/o `+`. You have 20 of them in the same file. Now you add one variable to one of them - now you have to change the 19 others as well, because consistency? Meh.

This is how it has been done throughout the server code base. I suggest that, if we want to clean it up (hint; I don't ;-)), we do this separately and use string-format instead.

I'll merge this for now since the situation seems painful.

PR #356, opened 2018-09-22
StatsD is now optional

Rackover:
You can now disable statsD via an env/or the configuration file so that server set up is easier for testing.
It's replaced by a mock object when disabled.

### DIFF
--- a/server/__init__.py
+++ b/server/__init__.py
@@ -25,6 +25,7 @@ from .player_service import PlayerService
 from .game_service import GameService
 from .ladder_service import LadderService
 from .control import init as run_control_server
+from unittest import mock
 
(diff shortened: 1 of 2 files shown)


micheljung:
Nice in principle, however, production code should not depend on `unittest`. This dependency should be for tests only.
I leave it up to someone else whether to merge this or not but if it was my project I wouldn't (but there you wouldn't even face this problem :P)

micheljung:
According to rackover, sheeo approved this

Sheeo:
I most certainly did not approve of this.

Brutus5000:
Preferably I wouldn't need to quote the chat, but as things seem to escalate I bring up the whole conversation.
grafik

PR #357, opened 2018-09-23
More unrank conditions + Few improvement

Rackover:
This PR is linked to https://github.com/FAForever/db/pull/173
( Please delete my previous PR https://github.com/FAForever/server/pull/348 )

Added 3 new unrank possibilities
- Map not in the vault triggers BAD_MAP
- HAS_AI for games with at least one AI
- MULTI_TEAM for 2v2v2 or 3v3v3 or this kind of pseudo-ffaesque configuration

I wrote unit tests for the two new validation marks aswell.

### DIFF
--- a/server.py
+++ b/server.py
@@ -44,7 +44,8 @@ if __name__ == '__main__':
         from docopt import docopt
         args = docopt(__doc__, version='FAF Server')
 
-        logger.info("Using StatsD server: ".format(config.STATSD_SERVER))
+        if config.ENABLE_STATSD:
+            logger.info("Using StatsD server: ".format(config.STATSD_SERVER))
 
         # Make sure we can shutdown gracefully
         signal.signal(signal.SIGTERM, signal_handler)
--- a/server/games/game.py
+++ b/server/games/game.py
@@ -8,6 +8,7 @@ from typing import Union
 import trueskill
 from trueskill import Rating
 import server.db as db
+import aiomysql
 from server.abc.base_game import GameConnectionState, BaseGame, InitMode
 from server.players import Player, PlayerState
 
@@ -80,8 +81,9 @@ class ValidityState(IntEnum):
     FFA_NOT_RANKED = 14
     UNEVEN_TEAMS_NOT_RANKED = 15
     UNKNOWN_RESULT = 16
-    MULTI_TEAM = 17
-    HAS_AI_PLAYERS = 18
+    UNLOCKED_TEAMS = 17
+    MULTI_TEAM = 18
+    HAS_AI_PLAYERS = 19
 
 
 class GameError(Exception):
@@ -139,6 +141,7 @@ class Game(BaseGame):
                             'CheatsEnabled': 'false',
                             'PrebuiltUnits': 'Off',
                             'NoRushOption': 'Off',
+                            'TeamLock': 'locked',
                             'RestrictedCategories': 0}
 
         self.mods = {}
@@ -566,6 +569,9 @@ class Game(BaseGame):
         elif self.gameOptions["RestrictedCategories"] != 0:
             await self.mark_invalid(ValidityState.BAD_UNIT_RESTRICTIONS)
 
+        elif self.gameOptions["TeamLock"] != "locked":
+            await self.mark_invalid(ValidityState.UNLOCKED_TEAMS)
+
     async def launch(self):
         """
         Mark the game as live.
@@ -594,20 +600,15 @@ class Game(BaseGame):
         care about recording stats for, after all).
         """
         async with db.db_pool.get() as conn:
-            cursor = await conn.cursor()
+            cursor = await conn.cursor(aiomysql.DictCursor)
 
             # Determine if the map is blacklisted, and invalidate the game for ranking purposes if
             # so, and grab the map id at the same time.
             await cursor.execute("SELECT id, ranked FROM map_version "
                                  "WHERE lower(filename) = lower(%s)", (self.map_file_path,))
             result = await cursor.fetchone()
-            if result:
-                (self.map_id, ranked) = result
 
-                if not ranked:
-                    await self.mark_invalid(ValidityState.BAD_MAP)
-
-            else:
+            if not result or not result['ranked']:
                 await self.mark_invalid(ValidityState.BAD_MAP)
 
             modId = self.game_service.featured_mods[self.game_mode].id
--- a/server/games/game.py
+++ b/server/games/game.py
@@ -142,6 +142,7 @@ class Game(BaseGame):
                             'PrebuiltUnits': 'Off',
                             'NoRushOption': 'Off',
                             'TeamLock': 'locked',
+                            'AIReplacement': 'Off',
                             'RestrictedCategories': 0}
 
         self.mods = {}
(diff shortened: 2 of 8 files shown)


Geosearchef:
just close your previous PR, you cannot delete a PR or issue

server.py
```diff
@@ -30,7 +30,7 @@
     stderr_handler = logging.StreamHandler()
     stderr_handler.setFormatter(logging.Formatter('%(levelname)-8s %(name)-30s %(message)s'))
     logger.addHandler(stderr_handler)
-    logger.setLevel(logging.DEBUG)
+    logger.setLevel(logging.INFO)
```
Geosearchef: sure you wanna push that to production?

server.py
```diff
@@ -44,7 +44,8 @@ def signal_handler(signal, frame):
         from docopt import docopt
         args = docopt(__doc__, version='FAF Server')
 
-        logger.info("Using StatsD server: ".format(config.STATSD_SERVER))
+        if config.ENABLE_STATSD:
+            logger.info("Using StatsD server: ".format(config.STATSD_SERVER))
```
Geosearchef: I may be blind/not have enough knowledge of python, but isn't the parameter never inserted anywhere?

server/games/game.py
```diff
@@ -80,6 +80,8 @@ class ValidityState(IntEnum):
     FFA_NOT_RANKED = 14
     UNEVEN_TEAMS_NOT_RANKED = 15
     UNKNOWN_RESULT = 16
+    MULTI_TEAM = 17
+    HAS_AI = 18
```
Geosearchef: Both are not self explanatory, please extend the name (like UNEVEN_TEAM_NOT_RANKED) or at least add a comment

server/games/game.py
```diff
@@ -593,6 +607,9 @@ def clear_slot(self, slot_index):
                 if not ranked:
                     await self.mark_invalid(ValidityState.BAD_MAP)
 
+            else:
+                await self.mark_invalid(ValidityState.BAD_MAP)
```
Geosearchef: that's a lot of whitespace above

Geosearchef: you set invalid if result and not ranked but also when no result
--> boolean logic can be shortened if    `not result present OR result not ranked` while evaluating just the not result present if it isn't present (I don't know python logic operators) so you don't try to access a null object.

server/lobbyconnection.py
```diff
@@ -485,6 +485,7 @@ def check_version(self, message):
             update_msg['command'] = 'welcome'
             # For compatibility with 0.10.x updating mechanism
             self.sendJSON(update_msg)
+            self._logger.debug("Rejected connection (unknown version) from "+self.user_agent+"("+version+")")
```
Geosearchef: change to warning, unknown version is sth that is never to be actually expected

server/lobbyconnection.py
```diff
@@ -495,10 +496,12 @@ def check_version(self, message):
                 if "+" in version:
                     version = version.split('+')[0]
                 if semver.compare(versionDB, version) > 0:
+                    self._logger.debug("Rejected connection (outdated) from "+self.user_agent+"("+version+")")
                     self.sendJSON(update_msg)
                     return False
             except ValueError:
                 self.sendJSON(update_msg)
+                self._logger.debug("Rejected connection (value error) from "+self.user_agent+"("+version+")")
```
Geosearchef: also no debug, error is not sth ordinary

tests/unit_tests/test_game.py
```diff
@@ -72,6 +72,29 @@ def test_instance_logging(game_stats_service):
     await game.on_game_end()
     assert game.validity == ValidityState.FFA_NOT_RANKED
 
+async def test_multi_team(game):
```
Geosearchef: this test is about invalidity, add that to the method name for easy understandability, see below `uneven_games_not_ranked`, includes what you expect the test to result

tests/unit_tests/test_game.py
```diff
@@ -72,6 +72,29 @@ def test_instance_logging(game_stats_service):
...
+    add_players(game, 2, team=1)
+    add_players(game, 2, team=2)
+    add_players(game, 2, team=3)
+    await game.launch()
+    await game.add_result(0, 1, 'VICTORY', 5)
+    game.launched_at = time.time() - 60*20 # seconds
+    await game.on_game_end()
+    assert game.validity == ValidityState.MULTI_TEAM
+
+async def test_has_ai(game):
```
Geosearchef: same issue

tests/unit_tests/test_lobbyconnection.py
```diff
@@ -246,7 +246,7 @@ def test_command_admin_closeFA(mocker, lobbyconnection):
 
     protocol.send_message.assert_any_call({
         "command": "avatar",
-        "avatarlist": [{'url': 'http://content.faforever.com/faf/avatars/qai2.png', 'tooltip': 'QAI'}]
+        "avatarlist": [{'url': 'http://content.faforever.com/faf/avatars/qai2.png', 'tooltip': 'QAI'}, {'url': 'http://content.faforever.com/faf/avatars/UEF.png', 'tooltip': 'UEF'}]
```
Geosearchef: this has nothing to do with your PR or commit name, add one feature each time, is this causing a problem elsewhere?

server/games/game.py
```diff
@@ -533,7 +543,11 @@ def clear_slot(self, slot_index):
...
 
-        if self.is_ffa:
+        if self.has_ai:
+            await self.mark_invalid(ValidityState.HAS_AI)
+        elif self.is_multi_team:
+            await self.mark_invalid(ValidityState.MULTI_TEAM)
+        elif self.is_ffa:
             await self.mark_invalid(ValidityState.FFA_NOT_RANKED)
         elif not self.is_even:
             await self.mark_invalid(ValidityState.UNEVEN_TEAMS_NOT_RANKED)
```
Geosearchef: calling self.mark_invalid looks a bit duplicated to me/verbose, maybe consider (I'm not quite sure though) replacing that.

Is there some way to create a mapping between the variables and Validity states and then just get the validity state into a local variable that you then:

```
if(invalidState != unkownNullThingy) {
   await self.mark_inv(invalidState)
}
```

server.py
Rackover: Imported config.py as config !

server/games/game.py
Rackover: MULTI_TEAM is already referenced in the DB so i'm going to comment this one.
I might be able to change HAS_AI though

tests/unit_tests/test_lobbyconnection.py
Rackover: This is preventing a build entirely

Geosearchef: Then we should find out why that was missing and confirm that adding it is the way to go before just smuggeling into this commit.

server/games/game.py
Rackover: But ranked is not declared on the same level, so it won't be accessable out of the "if result" loop. How am I supposed to solve this ?

server/games/game.py
Rackover: Those are already documented in the DB with a fancy comment and referenced under the exact same name - they're also in the category "ValidityState", making their function quite obvious

tests/unit_tests/test_lobbyconnection.py
Rackover: It is, it's because the test data contains 2 avatars and the server is waiting for only 1 to exist. As I've said on slack the best way to solve this is to compare the test result to what's inside the database (instead of this static string test), but for the moment I consider that to be enough. What broke this is that the test-data has been modified at one point which made this server impossible to build, something I've fixed a few days ago by editing the test data again. This test has simply become outdated.

server/games/game.py
Geosearchef: yes, you somehow need to get the ranked out of result in the same expression, I don't know python, if result is present meaning the first expression evaluated to false meaning the second one will be evaluated, can't you just get that out of the tuple (via index out of an array)?

PR #361, opened 2018-10-25
Implementing exotic newb-friendly matchmaking

Rackover:
@Exotic-Retard 
Implemented a system where new players (<10 games, or any threshold specified by ENV vars) get a rating adjustement for matchmaking so that matchmaking becomes more fair

### DIFF
--- a/server/config.py
+++ b/server/config.py
@@ -56,3 +56,6 @@ API_BASE_URL = os.getenv("API_BASE_URL", "https://api.dev.faforever.com/jwt")
 FAF_POLICY_SERVER_BASE_URL = os.getenv("FAF_POLICY_SERVER_BASE_URL", "http://faf-policy-server")
 FORCE_STEAM_LINK_AFTER_DATE = int(os.getenv('FORCE_STEAM_LINK_AFTER_DATE', 1536105599)) # 5 september 2018 by default
 FORCE_STEAM_LINK = os.getenv('FORCE_STEAM_LINK', 'false').lower() == 'true'
+
+NEWBIE_BASE_MEAN = int(os.getenv('NEWBIE_BASE_MEAN', 500))
+NEWBIE_MIN_GAMES = int(os.getenv('NEWBIE_MIN_GAMES', 10))
--- a/server/matchmaker/search.py
+++ b/server/matchmaker/search.py
@@ -2,6 +2,7 @@ import asyncio
 import time
 
 import math
+import server.config as config
 
 from server.decorators import with_logger
 from trueskill import quality_1vs1, Rating
@@ -14,6 +15,7 @@ class Search:
     """
     Represents the state of a users search for a match.
     """
+
     def __init__(self, player, start_time=None, rating_prop='ladder_rating'):
         """
         Default ctor for a search
@@ -41,7 +43,26 @@ class Search:
         }
 
     @property
+    def adjusted_rating(self):
+        """
+        Returns an adjusted mean with a simple linear interpolation between current mean and a specified base mean
+        """
+        mean, dev = self.player.ladder_rating
+        adjusted_mean = ((config.NEWBIE_MIN_GAMES - self.player.numGames) * config.NEWBIE_BASE_MEAN
+                         + self.player.numGames * mean) / config.NEWBIE_MIN_GAMES
+        return (adjusted_mean, dev)
+
+    @property
     def rating(self):
+        num_games = self.player.numGames
+        # New players (less than config.NEWBIE_MIN_GAMES games) match against less skilled opponents
+        if num_games <= config.NEWBIE_MIN_GAMES and self.rating_prop == 'ladder_rating':
+            return self.adjusted_rating
+        else:
+            return self.raw_rating
+
+    @property
+    def raw_rating(self):
         return getattr(self.player, self.rating_prop)
 
     @property
@@ -86,13 +107,13 @@ class Search:
             if deviation >= d:
                 return max(q - self.search_expansion, 0)
 
-    def quality_with(self, opponent):
-        if not isinstance(opponent, Player):
-            raise TypeError("{} is not a valid player to match with".format(opponent))
-        if not getattr(opponent, self.rating_prop):
+    def quality_with(self, opponent_search: 'Search'):
+        if not isinstance(opponent_search.player, Player):
+            raise TypeError("{} is not a valid player to match with".format(opponent_search.player))
+        if opponent_search.raw_rating is None:
             return 0
         return quality_1vs1(Rating(*self.rating),
-                            Rating(*getattr(opponent, self.rating_prop)))
+                            Rating(*opponent_search.rating))
 
     @property
     def is_matched(self):
@@ -111,8 +132,8 @@ class Search:
         """
         if not isinstance(other, Search):
             return False
-        elif self.quality_with(other.player) >= self.match_threshold and \
-            other.quality_with(self.player) >= other.match_threshold:
+        elif self.quality_with(other) >= self.match_threshold and \
+            other.quality_with(self) >= other.match_threshold:
             return True
         return False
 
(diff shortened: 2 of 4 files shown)


server/config.py
```diff
@@ -56,3 +56,6 @@
 FAF_POLICY_SERVER_BASE_URL = os.getenv("FAF_POLICY_SERVER_BASE_URL", "http://faf-policy-server")
 FORCE_STEAM_LINK_AFTER_DATE = int(os.getenv('FORCE_STEAM_LINK_AFTER_DATE', 1536105599)) # 5 september 2018 by default
 FORCE_STEAM_LINK = os.getenv('FORCE_STEAM_LINK', 'false').lower() == 'true'
+
+NEWBIE_BASE_MEAN = int(os.getenv('NEWBIE_BASE_MEAN', 500))
+NEWBIE_MIN_GAMES = int(os.getenv('NEWBIE_MIN_GAMES', 10))
```
norraxx: New line at the end of file

server/matchmaker/search.py
```diff
@@ -40,8 +42,29 @@ def __init__(self, player, start_time=None, rating_prop='ladder_rating'):
             0: 0.8
         }
 
+    @property
```
norraxx: Why do we need property, not function, historic reasons?

Properties by default are evil

server/matchmaker/search.py
```diff
@@ -40,8 +42,29 @@ def __init__(self, player, start_time=None, rating_prop='ladder_rating'):
...
+        Returns an adjusted mean with a simple linear interpolation between current mean and a specified base mean
+        """
+        if self.rating_prop=='ladder_rating':
+            numgames = self.player.numGames
+            # New players (less than config.NEWBIE_MIN_GAMES games) match against less skilled opponents
+            if numgames <= config.NEWBIE_MIN_GAMES:
+                mean, dev = self.player.ladder_rating
+                adjusted_mean = ((config.NEWBIE_MIN_GAMES - numgames) * config.NEWBIE_BASE_MEAN + numgames * mean) / config.NEWBIE_MIN_GAMES
+                return (adjusted_mean, dev)
+
```
norraxx: No default return, what will happen of game is not "ladder_rating"?

server/matchmaker/search.py
```diff
@@ -40,8 +42,29 @@ def __init__(self, player, start_time=None, rating_prop='ladder_rating'):
...
 
+    @property
+    def adjusted_rating(self):
+        """
+        Returns an adjusted mean with a simple linear interpolation between current mean and a specified base mean
+        """
+        if self.rating_prop=='ladder_rating':
+            numgames = self.player.numGames
+            # New players (less than config.NEWBIE_MIN_GAMES games) match against less skilled opponents
+            if numgames <= config.NEWBIE_MIN_GAMES:
```
norraxx: Double condition check? At "rating" call..

server/matchmaker/search.py
```diff
@@ -40,8 +42,29 @@ def __init__(self, player, start_time=None, rating_prop='ladder_rating'):
...
+                mean, dev = self.player.ladder_rating
+                adjusted_mean = ((config.NEWBIE_MIN_GAMES - numgames) * config.NEWBIE_BASE_MEAN + numgames * mean) / config.NEWBIE_MIN_GAMES
+                return (adjusted_mean, dev)
+
     @property
     def rating(self):
+        numGames = self.player.numGames
+        if numGames <= config.NEWBIE_MIN_GAMES:
+            return self.adjusted_rating
```
norraxx: just use simple functions

server/matchmaker/search.py
```diff
@@ -40,8 +42,29 @@ def __init__(self, player, start_time=None, rating_prop='ladder_rating'):
             0: 0.8
         }
 
+    @property
+    def adjusted_rating(self):
+        """
+        Returns an adjusted mean with a simple linear interpolation between current mean and a specified base mean
+        """
+        if self.rating_prop=='ladder_rating':
```
norraxx: ` == `

server/matchmaker/search.py
```diff
@@ -123,6 +146,17 @@ def match(self, other: 'Search'):
         :return:
         """
         self._logger.info("Matched %s with %s", self.player, other.player)
+        
+        numgames = self.player.numGames
+        if numgames <= config.NEWBIE_MIN_GAMES:
+            mean, dev = self.unadjusted_rating
+            adjusted_mean = self.adjusted_rating
```
norraxx: That property returns 2 values or None, but result is in one

server/matchmaker/search.py
Rackover: what do you mean ?

server/matchmaker/search.py
Rackover: What do you mean ?

server/matchmaker/search.py
Rackover: What is the issue ? Can you please elaborate a little bit ?

server/matchmaker/search.py
Rackover: For now I will leave the @property, because they are used throughout the whole server code and shall we decide to remove them, we would be better doing a huge "clean up" commit that would remove all @property that don't belong.

Geosearchef:
unadjusted rating doesn't seem like a good name to me, sounds like it requires adjustment, maybe just call it rating? Otherwise it looks fine to me except for the @Property norrax mentioned, I have no idea what that does. I also didn't check the rating calc formula but I bet enough people have thought about that.

server/matchmaker/search.py
```diff
@@ -123,6 +147,17 @@ def match(self, other: 'Search'):
         :return:
         """
         self._logger.info("Matched %s with %s", self.player, other.player)
+        
+        numgames = self.player.numGames
+        if numgames <= config.NEWBIE_MIN_GAMES:
```
Geosearchef:
```suggestion
        if numgames == config.NEWBIE_MIN_GAMES:
```
why <= instead of < ?
If a newbie has a the min amount of games doesn't that imply he isn't one anymore?

tests/unit_tests/test_matchmaker_queue.py
```diff
@@ -13,19 +14,25 @@ def matchmaker_queue(player_service, game_service):
...
-    return Player('Dostya',   id=1, ladder_rating=(1500, 50)), \
-           Player('Brackman', id=2, ladder_rating=(1500, 50)), \
-           Player('Zoidberg', id=3, ladder_rating=(1500, 50)), \
-           Player('QAI',      id=4, ladder_rating=(1500, 50)), \
-           Player('Rhiza',    id=5, ladder_rating=(1500, 50))
+    return Player('Dostya',   id=1, ladder_rating=(1500, 50), numGames=(config.NEWBIE_MIN_GAMES+1)), \
+           Player('Brackman', id=2, ladder_rating=(1500, 50), numGames=(config.NEWBIE_MIN_GAMES+1)), \
+           Player('Zoidberg', id=3, ladder_rating=(1500, 50), numGames=(config.NEWBIE_MIN_GAMES+1)), \
+           Player('QAI',      id=4, ladder_rating=(1500, 50), numGames=(config.NEWBIE_MIN_GAMES+1)), \
+           Player('Rhiza',    id=5, ladder_rating=(1500, 50), numGames=(config.NEWBIE_MIN_GAMES+1))
```
Geosearchef: I would remove the +1

Rackover:
Rating calc formula has been provided by exotic who I trust over the rating calculations.
Regarding the name, you're right - I can't call it rating, because the `rating` is supposed to return the rating to use for that specific Search. However unadjusted_rating sure is bad, I'm going to change that name.

server/matchmaker/search.py
```diff
@@ -40,8 +42,30 @@ def __init__(self, player, start_time=None, rating_prop='ladder_rating'):
...
+            mean, dev = self.player.ladder_rating
+            adjusted_mean = ((config.NEWBIE_MIN_GAMES - self.player.numGames) * config.NEWBIE_BASE_MEAN
+                             + self.player.numGames * mean) / config.NEWBIE_MIN_GAMES
+            return (adjusted_mean, dev)
+
+        return self.unadjusted_rating
+
     @property
     def rating(self):
+        num_games = self.player.numGames
```
Geosearchef: why define extra variable?

server/matchmaker/search.py
Rackover: This was in the original code, so I assume it is intended behavior from Exotic_retard. The newbie must have "more than X games".

server/matchmaker/search.py
norraxx: CPython compiles code into pyc, it doesn't has JIT compiler.

PR #362, opened 2018-10-28
Filtering non-latin names

Rackover:
Prevents database errors on game renaming.
Client displaying of game names is not affected.

### DIFF
--- a/server/games/game.py
+++ b/server/games/game.py
@@ -2,6 +2,7 @@ from enum import IntEnum, unique
 import logging
 import time
 import functools
+import re
 from collections import defaultdict
 import asyncio
 from typing import Union
@@ -122,7 +123,7 @@ class Game(BaseGame):
         self.visibility = VisibilityState.PUBLIC
         self.max_players = 12
         self.host = host
-        self.name = name
+        self.name = self.sanitize(name)
         self.map_id = None
         self.map_file_path = 'maps/%s.zip' % map
         self.map_scenario_path = None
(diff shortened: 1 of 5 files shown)


server/games/game.py
```diff
@@ -664,8 +665,9 @@ def clear_slot(self, slot_index):
     def getGamemodVersion(self):
         return self.game_service.game_mode_versions[self.game_mode]
 
+    # Allowing latin names only, and truncating them to be less than 64 characters long
     def sanitize(self, name):
-        return re.sub('[^\x20-\xFF]+', '_', name)
+        return re.sub('[^\x20-\xFF]+', '_', name)[0:64]
```
Brutus5000: The database length is 128. Why choose 64?

Rackover: 128 is hella long for a game name, in best cases I think 64 would be enough. But I can make it [0:127] if you prefer.

tests/unit_tests/test_gameconnection.py
```diff
@@ -124,7 +124,7 @@ def test_abort(game_connection, game, players):
 
 async def test_handle_action_GameOption_change_name(game, game_connection):
     await game_connection.handle_action('GameOption', ['Title', 'All welcome'])
-    assert game.name == 'All welcome'
+    assert game.name == game.sanitize('All welcome')
```
Brutus5000: Calling functions for asserts is bad practice. That adds unnecessary dependencies into the test.

Rackover: I'm encountering a problem with that one where "game.name" becomes a mock() object, because when instantiating a game, the name is automatically sanitized (but here the game is mocked). So I must either run the test like this (so it gives out "mock == mock"), or remove it.
(or else i have no idea)

Brutus5000: Good lord. That's a horrible test setup, but I guess we got to live with it for now.

server/games/game.py
Brutus5000: I thought you do this to prevent games not being saved. So yes, the limit should be the database limit.

PR #368, opened 2018-11-13
[WIP] Army stats fix

Rackover:
There seems to be shady stuff going on with that army_result
When put to 0 a lot of tests about achievements start failing, but somehow it doesn't feel so related. I wonder if those tests ran in the past or if they were simply skipped because of the way game_stats_service returns without exceptions when encountering weird results.


Rackover:
(yes obviously this is WIP)

PR #370, opened 2018-11-27
Java api

Rackover:
Now uses java api instead of Python API

Added a config option to disable the java api since the java api is currently unable to handle achievements and events. This will allow us to update the python server early and to enable the feature later via an ENV variable.


server/api/api_accessor.py
```diff
@@ -3,39 +3,47 @@
...
-        loop = asyncio.get_event_loop()
-        result = await loop.run_in_executor(
-            None,
-            partial(self.http(player_id).request, API_BASE_URL + path, "POST", headers=headers, body=json.dumps(data))
-        )
+        code, text = await self.api_patch("events/update", events_data)
+        return code, text
 
+    async def api_get(self, path):
+        result = self._session.get(API_BASE_URL + path)
```
norraxx: Why, `api_get` returns full result, but `api_patch`, `update_events` and `update_achievements` tuple?

server/api/api_accessor.py
```diff
@@ -3,39 +3,47 @@
...
+        self._session = oauth_session(client=client)
+        self._token = self._session.fetch_token(token_url=API_TOKEN_URI, client_id=API_CLIENT_ID, client_secret=API_CLIENT_SECRET)
+
+    async def update_achievements(self, achievements_data, player_id):
+
+        # Converting the achievements to a format the jAPI can understand
+        for achievement in achievements_data:
+            achievement['playerId'] = player_id
+            achievement['achievementId'] = achievement.pop('achievement_id')
+            achievement['operation'] = achievement.pop('update_type')
```
norraxx: Is there is reason why you mutate `achievements_data`?

`.get(` is much better

server/api/api_accessor.py
```diff
@@ -3,39 +3,47 @@
...
-    async def api_get(self, path, player_id):
-        loop = asyncio.get_event_loop()
-        result = await loop.run_in_executor(None, partial(self.http(player_id).request, API_BASE_URL + path, "GET"))
-        return result
+    def __init__(self, backend_app_client=BackendApplicationClient, oauth_session=OAuth2Session):
+        client = backend_app_client(client_id=API_CLIENT_ID)
+        self._session = oauth_session(client=client)
+        self._token = self._session.fetch_token(token_url=API_TOKEN_URI, client_id=API_CLIENT_ID, client_secret=API_CLIENT_SECRET)
+
+    async def update_achievements(self, achievements_data, player_id):
```
norraxx: Why `ApiAccessor` has logic (`update_achievements`, `update_events`)?

server/api/api_accessor.py
```diff
@@ -3,39 +3,47 @@
...
-        self._service_account_credentials.token_uri = API_TOKEN_URI
-
-    async def api_get(self, path, player_id):
-        loop = asyncio.get_event_loop()
-        result = await loop.run_in_executor(None, partial(self.http(player_id).request, API_BASE_URL + path, "GET"))
-        return result
+    def __init__(self, backend_app_client=BackendApplicationClient, oauth_session=OAuth2Session):
+        client = backend_app_client(client_id=API_CLIENT_ID)
+        self._session = oauth_session(client=client)
+        self._token = self._session.fetch_token(token_url=API_TOKEN_URI, client_id=API_CLIENT_ID, client_secret=API_CLIENT_SECRET)
```
norraxx: where token is used? and what does it for?

server/config.py
```diff
@@ -48,10 +48,11 @@
 CHALLONGE_KEY = "challonge_key"
 CHALLONGE_USER = "challonge_user"
 
-API_CLIENT_ID = os.getenv("API_CLIENT_ID", "6ccaf75b-a1f3-48be-bac3-4e9ffba81eb7")
+API_CLIENT_ID = os.getenv("API_CLIENT_ID", "client_id")
```
norraxx: why?

server/stats/achievement_service.py
```diff
@@ -91,12 +91,30 @@ def __init__(self, api_accessor: ApiAccessor):
...
+                    current_state=achievement['attributes']['state'],
+                    newly_unlocked=achievement['attributes']['newlyUnlocked']
+                )
+                if 'currentSteps' in achievement['attributes']:
+                    converted_achievement['current_steps'] = achievement['attributes']['currentSteps']
+
+                achievements_data.append(converted_achievement)
+
+            return achievements_data
+        return None
```
norraxx: `return achievements_data`
Skip just returning `None`

See lower

tests/unit_tests/test_achievement_service.py
```diff
@@ -8,11 +8,10 @@
 from server.stats.achievement_service import AchievementService
```
norraxx: `import asyncio` is unused
`MagicMock` is unused

tests/unit_tests/test_achievement_service.py
```diff
@@ -8,11 +8,10 @@
...
 from tests import CoroMock
 
-
```
norraxx: \n

tests/unit_tests/test_achievement_service.py
```diff
@@ -21,13 +20,15 @@ def service(api_accessor: ApiAccessor):
     return AchievementService(api_accessor)
```
norraxx: \n

tests/unit_tests/test_event_service.py
```diff
@@ -37,6 +39,22 @@ def service(api_accessor: ApiAccessor):
...
+        converted_event = dict(
+            event_id=event['attributes']['eventId'],
+            count=event['attributes']['count']
+        )
+        events_data.append(converted_event)
+
+
+    assert result == events_data
+
+    service.api_accessor.update_events.assert_called_once_with(queue, 42)
```
norraxx: `(queue, 42)` - space

tests/unit_tests/test_event_service.py
```diff
@@ -37,6 +39,22 @@ def service(api_accessor: ApiAccessor):
...
+
+    # API is not working
+    service.api_accessor.update_events.coro.return_value = (500, None)
+    result = await service.execute_batch_update(42, queue)
+    assert result is None
+
```
norraxx: This test you can split in 3 smaller

server/api/api_accessor.py
```diff
@@ -3,39 +3,47 @@
...
 
 import pkg_resources
-from httplib2 import Http
-from oauth2client.service_account import ServiceAccountCredentials
-from server.config import API_TOKEN_URI, API_BASE_URL
+from server.config import API_TOKEN_URI, API_BASE_URL, API_CLIENT_SECRET, API_CLIENT_ID
+from oauthlib.oauth2 import BackendApplicationClient
+from requests_oauthlib import OAuth2Session
+import os
+os.environ['OAUTHLIB_INSECURE_TRANSPORT'] = '1'
```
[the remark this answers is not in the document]
Rackover: you are right, it's only here because afaik test api does not support HTTPS. I will comment it, but let it here.

server/stats/event_service.py
```diff
@@ -57,13 +57,25 @@ def __init__(self, api_accessor: ApiAccessor):
...
+            events_data = []
+            for event in json.loads(content)['data']:
+                converted_event = dict(
+                    event_id=event['attributes']['eventId'],
+                    count=event['attributes']['count']
+                )
+                events_data.append(converted_event)
 
+            return events_data
+        return None
```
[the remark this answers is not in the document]
Rackover: It's better to not return a empty events_data with the wrong structure because in case of event service it will be fine (the returned value is not used anyway), but in case of achievement service it could cause trouble in the server OR send wrong data to the client which we don't want. In case there's an issue with the API, i think we are better returning None and wait for the storm to pass.

server/api/api_accessor.py
```diff
@@ -3,39 +3,47 @@
 import asyncio
```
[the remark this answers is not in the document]
Rackover: isn't Asyncio needed for async def functions ?

tests/unit_tests/test_api_accessor.py
```diff
@@ -0,0 +1,49 @@
+import asyncio
+import json
+from unittest.mock import Mock
+
+import pytest
+
+from server.config import API_BASE_URL
+from server.api.api_accessor import ApiAccessor
+
```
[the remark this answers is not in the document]
Rackover: I don't understand

tests/unit_tests/test_api_accessor.py
```diff
@@ -0,0 +1,49 @@
...
+class Fake_session:
+    def __init__(self, client):
+        self.request = Mock(return_value=Fake_request_response())
+        self.get = Mock(return_value='test')
+
+    def fetch_token(self, token_url, client_id, client_secret):
+        pass
+
```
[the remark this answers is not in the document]
Rackover: Where do you want me to put the line break ?

tests/unit_tests/test_api_accessor.py
```diff
@@ -0,0 +1,49 @@
...
+class Fake_request_response:
+    def __init__(self):
+        self.status_code = 200
+        self.text = "test"
+
```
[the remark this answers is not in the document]
Rackover: Where do you want me to put the line break ?

server/api/api_accessor.py
```diff
@@ -3,39 +3,47 @@
...
-from httplib2 import Http
-from oauth2client.service_account import ServiceAccountCredentials
-from server.config import API_TOKEN_URI, API_BASE_URL
+from server.config import API_TOKEN_URI, API_BASE_URL, API_CLIENT_SECRET, API_CLIENT_ID
+from oauthlib.oauth2 import BackendApplicationClient
+from requests_oauthlib import OAuth2Session
+import os
+os.environ['OAUTHLIB_INSECURE_TRANSPORT'] = '1'
 
 CACERTS_FILE = pkg_resources.resource_filename('static', 'cacerts.txt')
```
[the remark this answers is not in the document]
Rackover: Nowhere, removing it

server/api/api_accessor.py
Rackover: I don't know what get( does. What do you mean ?

server/api/api_accessor.py
Rackover: Cause it needs conversion that is related to the API we're communicating with. If the api had been different, the transcription would have been different. If we change the API one day, we wanna have to modify ApiAccessor, not EventService and AchievementService again. I'm trying to make the API-related functions in that class.

server/api/api_accessor.py
Rackover: I believe it's used internally by request_oauthlib (the fetching of it, not the storing in particular)

server/config.py
Rackover: Client_id has changed, so the ID that was there is no longer relevant anyway

tests/unit_tests/test_event_service.py
Rackover: what ? there should not be space ? or there should be space ?

tests/unit_tests/test_event_service.py
Rackover: Yes but I would have to repeat the first few steps and assert they work before doing the others, which means if the first test was failing the three would too, and if the second test was failing the remaining would aswell. So I don't know if splitting that in multiple tests is actually better.

server/stats/achievement_service.py
```diff
@@ -91,12 +91,30 @@ def __init__(self, api_accessor: ApiAccessor):
...
             }]
+
+        Else, it returns None
         """
         self._logger.debug("Updating %d achievements", len(queue))
-        data = dict(updates=queue)
-        response, content = await self.api_accessor.api_post("/achievements/updateMultiple", player_id, data=data)
-
-        return json.loads(content.decode('utf-8'))['updated_achievements']
+        response, content = await self.api_accessor.update_achievements(queue, player_id)
```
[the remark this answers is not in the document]
Rackover: Okay, it is now

tests/unit_tests/test_api_accessor.py
```diff
@@ -0,0 +1,49 @@
...
+@pytest.fixture()
+def api_accessor():
+    return ApiAccessor(Mock, Fake_session)
+
```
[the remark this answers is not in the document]
Rackover: Where do you want me to put the line break ?

server/api/api_accessor.py
Rackover: Fixed

server/stats/achievement_service.py
Rackover: (I answered that concern in your other comment)

tests/unit_tests/test_achievement_service.py
Rackover: Where ?

server/stats/event_service.py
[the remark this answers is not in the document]
norraxx: Make code simple as much as it possible. Let it return just list. If return value is not used - don't return it.

server/api/api_accessor.py
```diff
@@ -3,39 +3,47 @@
...
 
-    def http(self, sub=None):
-        credentials = self._service_account_credentials.create_delegated(sub)
-        # FIXME ca_certs=CACERTS_FILE should be used, but it didn't work for some reason.
-        # Since we'll access the API locally over HTTP in future anyway, I decided to just skip validation for now
-        return credentials.authorize(Http(disable_ssl_certificate_validation=True))
+    async def api_patch(self, path, json_data):
+        headers = {'Content-type': 'application/json'}
+        result = self._session.request("PATCH", API_BASE_URL + path, headers=headers, json=json_data)
+        return result.status_code, result.text
```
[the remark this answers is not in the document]
norraxx: That mean put new line

tests/unit_tests/test_achievement_service.py
norraxx: There is https://www.python.org/dev/peps/pep-0008/.
Please check it, there are some rules, how code should be formatted.

tests/unit_tests/test_event_service.py
norraxx: space betwen comma and number

tests/unit_tests/test_event_service.py
norraxx: Splitting into multiple tests is better, even if you have to make some configuration. Configuration can be made via fixtures.

norraxx: If test will fail, you'll know what's going on.

server/stats/event_service.py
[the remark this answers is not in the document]
Rackover: Why should I return data with wrong structure on failure and data with right structure on success ? This is not as simple as simply returning None

server/api/api_accessor.py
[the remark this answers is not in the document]
Rackover: Yes I know \n means new line, my question is Where

tests/unit_tests/test_event_service.py
Rackover: It's already there 
image

server/api/api_accessor.py
Rackover: No, the pop is intentional, i'm formatting the dict() to send it as JSON later

server/api/api_accessor.py
```diff
@@ -1,41 +1,45 @@
...
+from oauthlib.oauth2 import BackendApplicationClient
+from requests_oauthlib import OAuth2Session
 
-import pkg_resources
-from httplib2 import Http
-from oauth2client.service_account import ServiceAccountCredentials
-from server.config import API_TOKEN_URI, API_BASE_URL
+"""
+Uncomment the following line if your API uses HTTP instead of HTTPS
+"""
```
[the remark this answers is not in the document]
Rackover: I won't ! Documentation is meant to be written in docstring whatever the length

tests/unit_tests/test_event_service.py
```diff
@@ -21,11 +20,14 @@ def service(api_accessor: ApiAccessor):
...
-    content = '''{"updated_events": [
-        { "event_id": "1-2-3", "count": 1},
-        { "event_id": "2-3-4", "count": 4}
-    ]}'''.encode('utf-8')
-    service.api_accessor.api_post = CoroMock(return_value=(None, content))
+
+    content = '{"data": [\
+        {"attributes": {"eventId": "1-2-3", "count": 1}},\
+        {"attributes": {"eventId": "2-3-4", "count": 4}}\
+    ]}'
```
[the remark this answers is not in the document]
Rackover: done

PR #371, opened 2018-11-28
Fix stats treating result list as a single result

Wesmania:

### DIFF
--- a/server/games/game.py
+++ b/server/games/game.py
@@ -710,7 +710,7 @@ class Game(BaseGame):
             score = max(score, result[2])
         return score
 
-    def get_army_result(self, player):
+    def get_army_results(self, player):
         return self._results.get(self.get_player_option(player.id, 'Army'))
 
     def compute_rating(self, rating='global'):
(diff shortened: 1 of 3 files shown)


Rackover:
https://github.com/FAForever/server/pull/262
https://github.com/FAForever/server/pull/257

This problem has been addressed differently in the past.
As much as your method works, those work aswell. I am not sure of which is the best.
I would say that making `get_army_result()` returning the most reported result makes more sense than renaming the functions `get_army_results` and make it return multiple things. The former does fit more into the actual system.
We could do a compromise by making a function `pick_most_reported` that would do that operation, and use `get_army_resulst` to other extents (although for the moment, *I see no use to such function returning a list of probably similar results*).

What do you think ?

Rackover:
After having seen DL/Crotalus's fix, I think i will go with that idea instead for the moment.
However I'd love to have you review it once it's done 🙂  Thanks for the help on this issue otherwise
Closing this PR to open another.

Issue #367, opened 2018-11-10
Army statistics can't be processed


micheljung:
```
faf-python-server                | INFO     CustomGame.8816629             Player(neVdun, 213834, (1145.36, 98.9353), (1500.0, 500.0)) reported result for army 4: defeat -10
faf-python-server                | DEBUG    server.stats.game_stats_service.GameStatsService Processing game stats for player: Dirtierdragon
faf-python-server                | ERROR    CustomGame.8816629             Army stats could not be processed from player Player(Dirtierdragon, 82365, (1103.032901286831, 102.59100077092083), (627.416, 71.2472)) in game Game(8816629,LittleKiller,maps/dual_gap_fix.v0004.v0004.zip,12)
faf-python-server                | Traceback (most recent call last):
faf-python-server                |   File "/code/server/games/game.py", line 297, in _process_army_stats_for_player
faf-python-server                |     await self._game_stats_service.process_game_stats(player, self, self._army_stats)
faf-python-server                |   File "/code/server/stats/game_stats_service.py", line 56, in process_game_stats
faf-python-server                |     survived = army_result[1] == 'victory'
faf-python-server                | IndexError: list index out of range

```

This was supposed to be fixed but it's still happening, see #252

PR #375, opened 2018-12-12
get_army_result now takes the most reported result

Rackover:
New version of https://github.com/FAForever/server/pull/262 that replaces #257 
Credit to @micheljung  or @Crotalus , I'm not sure. 
Closing theirs.

### DIFF
--- a/server/games/game.py
+++ b/server/games/game.py
@@ -12,7 +12,7 @@ import server.db as db
 import aiomysql
 from server.abc.base_game import GameConnectionState, BaseGame, InitMode
 from server.players import Player, PlayerState
-
+from collections import Counter
 
 @unique
 class GameState(IntEnum):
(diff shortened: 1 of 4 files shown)


Rackover:
Also, fixes https://github.com/FAForever/server/issues/367

server/games/game.py
```diff
@@ -711,7 +711,14 @@ def get_army_score(self, army):
         return score
 
     def get_army_result(self, player):
-        return self._results.get(self.get_player_option(player.id, 'Army'))
+        results = self._results.get(self.get_player_option(player.id, 'Army'))
+        if not results:
+            return None
+        counts = defaultdict(int)
```
Wesmania: Plain dict with setdefault would work just as well. Or I guess if we're using collections anyway, we could just use a Counter for the whole thing.

server/games/game.py
```diff
@@ -711,7 +711,14 @@ def get_army_score(self, army):
         return score
 
     def get_army_result(self, player):
-        return self._results.get(self.get_player_option(player.id, 'Army'))
+        results = self._results.get(self.get_player_option(player.id, 'Army'))
+        if not results:
```
Wesmania: Check if it's None instead plox, it's the usual idiom

server/games/game.py
Rackover: what do you mean by "plain dict with setdefault" ?

Wesmania: This is a throwaway dict, so I don't see what kind of data you'd need to write into it "later on". You're using a class from collections anyway (defaultdict), so you might as well use the one that's explicitly made for the thing you want to do. If anything, it's less code.

PR #377, opened 2018-12-22
Refactored search to remove redundant parameter 'player'

Askaholic:
Make the client responsible for creating the `Search` object using it's constructor instead of proxying arguments to in from `search()`.


norraxx:
TypeError: wait_for() missing 1 required positional argument: 'timeout'

Askaholic:
@norraxx What file/line number? Pretty sure I fixed all of those already.

Issue #372, opened 2018-12-08
Coop replays missing mapname in the replay file


speed2CZ:
When opening the replay files I found out that the coop replays are missing the mapname. The client is then not showing them properly in the online vault, and downlord's client can't even lauch the replays.

I dont know if it would completelly fix the issue with coop replays, but it would surely be a good start.

Some examples of the replays files:
Coop Game:
```
{"num_players": 1, "featured_mod_versions": {"1": 55, "2": 55, "3": 49, "4": 49, "5": 49, "6": 49, "7": 51, "8": 49, "9": 50, "10": 50, "11": 50, "12": 50, "13": 50, "14": 50, "15": 49, "16": 49, "17": 50, "18": 51, "19": 50, "20": 50, "21": 51, "22": 49}, "game_end": 1544305811.890091, "uid": 8945286, "title": "Wastefield's game", "game_type": 3, "teams": {}, "mapname": "", "featured_mod": "coop", "complete": true}
```
`"mapname": ""`

FAF game
```
{"num_players": 2, "featured_mod_versions": {}, "game_end": 1544306269.963174, "uid": 8945278, "title": "FaReNoy Vs bisq01", "game_type": 0, "teams": {"1": ["FaReNoy"]}, "mapname": "x1mp_017", "host": "FaReNoy", "featured_mod": "ladder1v1", "complete": true}
```
`"mapname": "x1mp_017"`

Brutus5000:
How does the python client know which map to load if the mapname is not specified?

Wesmania:
I believe there's a mapname in the lua replay header. Some data is duplicated between it and the json header, probably to make it easier for the client do parse replays in bulk.

Wesmania:
This is the query that the legacy replayserver does to pull out info including the mapname:
```
"SELECT game_featuredMods.gamemod, gameType, filename, gameName, host, login, playerId, AI, team FROM `game_stats` LEFT JOIN game_player_stats ON `game_player_stats`.`gameId` = game_stats.id LEFT JOIN table_map ON `game_stats`.`mapId` = table_map.id LEFT JOIN login ON login.id = `game_player_stats`.`playerId`  LEFT JOIN  game_featuredMods ON `game_stats`.`gameMod` = game_featuredMods.id WHERE game_stats.id = %i" % self.uid
```

Askaholic:
[...]

Issue #195, opened 2016-08-21
Coop Leaderboards


theeggroll:
Coop leaderboards have been down for a long time. Whenever you play a game in coop it just doesn't appear in the leaderboards. Also some missions have the best times recorded involve people using in game bugs to beat the mission significantly faster than should be possible. More details if needed

speed2CZ:
The game should send the data to the server here https://github.com/FAForever/fa/blob/develop/schook/lua/UserSync.lua#L135

and the server should process it here https://github.com/FAForever/server/blob/develop/server/gameconnection.py#L339

One thing that changed since the Zep time was that the coop now support changing difficulty as well, so that is one thing that should get added to the leaderboards as well. That will also require changing the data sent by the game to include it
Difficulty is represented by a number 1-3, 1 for easy, 2 medium, 3 hard

Rackover:
5ee9b868f1d8a4ed790410ed309ab4e8afa4087f

PR #380, opened 2018-12-23
Issue/#195 Fix coop scores not getting posted to leaderboard

Askaholic:
The LIKE query was not matching `filename` column correctly. `filename` has the format 'maps/name.version.zip' which is the same format as self.game.map_file_path. So the old query was checking for something like: '%/maps/name.version.zip.%' which clearly doesn't match the entries in the db.

The test will pass if it the db contains an entry for the prothyon coop map, however, this won't be available to Travis until a new DB release is made and faf-stack is updated to use it. But feel free to test it locally with the latest test-data.sql :D

### DIFF
--- a/tests/unit_tests/test_gameconnection.py
+++ b/tests/unit_tests/test_gameconnection.py
@@ -169,8 +169,9 @@ async def test_handle_action_GameResult_draw_ends_sim(game, game_connection):
     assert game.check_sim_end.called
 
 
-@pytest.mark.skip(reason="requires test data from faf-db:v63 which doesn't exist yet")
 async def test_handle_action_OperationComplete(game, game_connection):
+    pytest.skip("requires test data from faf-db:v63 which doesn't exist yet")
+
     game.map_file_path = "maps/prothyon16.v0005.zip"
     secondary = 1
     time_taken = '09:08:07.654321'
--- a/tests/unit_tests/test_gameconnection.py
+++ b/tests/unit_tests/test_gameconnection.py
@@ -170,6 +170,12 @@ async def test_handle_action_GameResult_draw_ends_sim(game, game_connection):
 
 
 async def test_handle_action_OperationComplete(game, game_connection):
+    """
+        Sends an OperationComplete action to handle action and verifies that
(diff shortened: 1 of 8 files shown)


Askaholic:
#195

Rackover:
May be related to https://github.com/FAForever/server/issues/372

Sheeo:
Good catch  👍

speed2CZ:
If this actually works, can someone wipe out the current times?

Rackover:
The current times ?

speed2CZ:
All the replays that are currently in the leaderboards, they are acient, broken, outdated

tests/unit_tests/test_gameconnection.py
```diff
@@ -147,16 +152,35 @@ def test_abort(game_connection, game, players):
...
+
 async def test_handle_action_GameResult_draw_ends_sim(game, game_connection):
     game_connection.ConnectToHost = CoroMock()
     await game_connection.handle_action('GameResult', [0, 'draw'])
 
     assert game_connection.finished_sim
     assert game.check_sim_end.called
+
+
+async def test_handle_action_OperationComplete(game, game_connection):
```
norraxx: IMHO we should write what do we test. Yes, we can talk about right name of the test, but we won't explain to everybody what test really does and why.

tests/unit_tests/test_gameconnection.py
```diff
@@ -147,16 +152,35 @@ def test_abort(game_connection, game, players):
...
+    pytest.skip("requires test data from faf-db:v63 which doesn't exist yet")
+
+    game.map_file_path = "maps/prothyon16.v0005.zip"
+    secondary = 1
+    time_taken = '09:08:07.654321'
+    await game_connection.handle_action('OperationComplete', ['1', secondary, time_taken])
+
+    import server.db as db
```
norraxx: don't we have fixture for db connection? If not, i wish that we would have it.

Rackover: We have a fixture for the db_pool in tests/conftest.py

Issue #292, opened 2017-05-01
aiohttp breakage


duk3luk3:
aiohttp removed `aiohttp.post` method in version 1.1.4, probably was long deprecated.

Need to fix this properly, meanwhile #291 is a band-aid fix.

PR #384, opened 2019-01-01
Cleanup dependencies and imports

Askaholic:
Removed `-f http://content.faforever.com/wheel/` from `requirements.txt` because this page is a 403 error.

Upgraded `aiohttp` as #292 is no longer relevant.

### DIFF
--- a/Dockerfile
+++ b/Dockerfile
@@ -6,7 +6,7 @@ RUN rm -rf /var/lib/apt/lists/* /tmp/* /var/tmp/*
 
 COPY requirements.txt /tmp/requirements.txt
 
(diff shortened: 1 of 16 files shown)


Askaholic:
It's used in `control.py` to run some sort of a server

Rackover:
I don't know if the control server even works, I've never managed to connect to it. Did you ?

norraxx:
It would be nice to have fully working pipenv file.

Askaholic:
@norraxx That would be easy to set up. Is there really a big advantage though?

norraxx:
Is there any reason, why do we use python3.5 not upper version python3.6?

norraxx:
Yes, there is big advantage for developer.

PR #385, opened 2019-01-02
Fix the game title sometimes being None throwing an exception

Rackover:
Provided by @dragonite


server/lobbyconnection.py
```diff
@@ -819,7 +819,7 @@ def command_game_host(self, message):
 
         assert isinstance(self.player, Player)
 
-        title = cgi.escape(message.get('title', ''))
+        title = cgi.escape(message.get('title'), '')
```
Askaholic: You still need to have the default of `''` in there if message doesn't have a `title` field. And if message does have a `title` field, but the value is `None`, you need to replace it with empty string. Also `cgi.escape` only takes one string parameter [https://docs.python.org/3/library/cgi.html#cgi.escape](https://docs.python.org/3/library/cgi.html#cgi.escape)

So it should look like this:

```suggestion
        title = cgi.escape(message.get('title', '') or '')
```

Which is pretty awful to read so maybe this is better:
```python
title = message.get('title', '') or ''
title = cgi.escape(title)
```

Or better yet, make a function for it.

server/lobbyconnection.py
```diff
@@ -771,7 +772,8 @@ def command_game_host(self, message):
 
         assert isinstance(self.player, Player)
 
-        title = cgi.escape(message.get('title', ''))
+        default_title = "%s' game".format(self.player.login)
+        title = html.escape(message.get('title', True) or default_title)
```
micheljung: Shouldn't this be
```suggestion
        title = html.escape(message.get('title', False) or default_title)
```
or even

```suggestion
        title = html.escape(message.get('title') or default_title)
```

PR #386, opened 2019-01-02
Control server now displays some stats, removed POST route to kick

Rackover:
Removed the POST routes to kick players as they're no longer needed, turned the control server into something fancier. 
One day maybe we can make something really interesting out of it - for the moment I think this is enough, and we may find some use for it as there's no real way ATM to look at what the python server is doing from the outside.

### DIFF
--- a/server/control.py
+++ b/server/control.py
@@ -5,11 +5,13 @@ Tiny local-only http server for getting stats and performing various tasks
 import asyncio
 
 import msgpack
+import socket
 
 from aiohttp import web
 import logging
-from server import PlayerService, GameService, LobbyConnection
+from server import PlayerService, GameService, LobbyConnection, config
 from server.games import Game
+from json import dumps
 
 logger = logging.getLogger(__name__)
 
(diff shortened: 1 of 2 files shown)


server/control.py
```diff
@@ -20,18 +22,13 @@ def __init__(self, game_service: GameService, player_service: PlayerService):
         self.player_service = player_service
 
     def games(self, request):
-        body = repr(list(map(Game.to_dict, self.game_service.all_games))).encode()
-        return web.Response(body=msgpack.packb(body), content_type='application/x-msgpack')
+        body = dumps(list(map(lambda p: p.to_dict(), self.game_service.all_games))).encode()
+        #body = dumps(list(self.game_service.all_games).__dict__).encode()
```
Askaholic: Delete commented code

server/control.py
```diff
@@ -20,18 +22,13 @@ def __init__(self, game_service: GameService, player_service: PlayerService):
...
     def players(self, request):
-        body = list(map(lambda p: p.to_dict(), self.player_service.players.values()))
-        return web.Response(body=msgpack.packb(body), content_type='application/x-msgpack')
-
-    async def kick_player(self, request):
-        player = self.player_service.players[int(request.match_info.get('player_id'))]
-        assert isinstance(player.lobby_connection, LobbyConnection)
-        player.lobby_connection.kick()
-        return web.Response()
+        body = dumps(list(map(lambda p: p.to_dict(), self.player_service.players.values()))).encode()
```
Askaholic: Maybe you can break out this part:
```python
list(map(lambda p: p.to_dict(), self.player_service.players.values()))
```

Into a reusable function. Something like:
```python
def to_dict_list(list_):
    return list(map(lambda p: p.to_dict(), list_))
```
Then this line becomes:
```suggestion
        body = dumps(to_dict_list(self.player_service.players.values()))
```
And you can also do it for the `/games` route as well.

And I would probably do the encode step separately for readability
```python
return web.Response(body=body.encode(), content_type='application/json')
```

PR #388, opened 2019-01-03
Issue/#387 validate coop results before adding to leaderboard

### LINKED ISSUE #387: Coop leaderboards doesn't validate game settings
Report from the forums: https://forums.faforever.com/viewtopic.php?f=3&t=17070

The server should check if the game was played without any sim mods and with defaults settings. Else the replay should not be added into the leaderboards.

Askaholic:
Fixes #387 

* [x] add test

Would be nice if there was a catch all 'Invalid Game Setting' validity state.

Requires:
https://github.com/FAForever/db/pull/187

### DIFF
--- a/server/games/game.py
+++ b/server/games/game.py
@@ -237,27 +237,38 @@ class Game(BaseGame):
         return len(self.AIs) > 0
 
     @property
-    def is_even(self):
+    def is_even(self) -> bool:
         teams = self.team_count()
-        if 1 in teams: # someone is in ffa team, all teams need to have 1 player
-            c = 1
-            teams.pop(1)
-        else:
-            n = len(teams)
-            if n <= 1: # 0 teams are considered even, single team not
-                return n == 0
+        if len(teams) < 2:
+            return False
+        team_member_counts = list(teams.values())
 
-            # all teams needs to have same count as the first
-            c = list(teams.values())[0]
-
-        for _, v in teams.items():
-            if v != c:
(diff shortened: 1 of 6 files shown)


server/games/game.py
```diff
@@ -540,39 +542,76 @@ def clear_slot(self, slot_index):
...
+                    return False
+        return True
 
-        elif self.gameOptions["NoRushOption"] != "Off":
-            await self.mark_invalid(ValidityState.NORUSH_ENABLED)
+    async def _validate_coop_game_settings(self):
+        """
+        Checks which only apply to the coop mode
+        """
+        # Note, the UNEVEN_TEAMS_NOT_RANKED state is used as a placeholder for reasons
```
Rackover: This sounds really dirty to me. I would prefer you to add all of those  conditions in the DB via a migration scripts, and to add them to server code later.

server/games/game.py
```diff
@@ -87,6 +87,10 @@ class ValidityState(IntEnum):
     UNLOCKED_TEAMS = 17
     MULTI_TEAM = 18
     HAS_AI_PLAYERS = 19
+    CIVILIANS_REVEALED = 20
+    TOO_EASY = 21
```
Rackover: WRONG_DIFFICULTY please 😛

Askaholic: Ok. Was trying to be consistent, but wrong difficulty is probably better.

Rackover:
Tested, working, merged.

PR #390, opened 2019-01-04
Update python to 3.6

Askaholic:
I've included a Dockerfile for using python:3.6-slim (which is debian based) as well in case there is some really compelling reason not to use alpine. But as long as there are no objections to alpine, I'll delete the slim one.


Rackover:
It passes build and I've been testing the server using only python 3.6 for months - there is nothing that indicates it doesn't work. I'm merging that in.

PR #392, opened 2019-01-06
Added tests for the none title being set automatically

Rackover:


tests/unit_tests/test_lobbyconnection.py
```diff
@@ -206,6 +220,14 @@ def test_command_game_host_calls_host_game_invalid_title(lobbyconnection,
         dict(command="notice", style="error", text="Non-ascii characters in game name detected."))
 
 
+def test_command_game_host_none_title(lobbyconnection,
+                                      mock_games,
+                                      test_game_info_none_title):
+
+    lobbyconnection.command_game_host(test_game_info_none_title)
+    assert(lobbyconnection.game_connection.game.name == "%s' game".format(lobbyconnection.player.login))
```
Askaholic: Don't use `%s` for formatting. This is the old syntax and doesn't work with `.format`:
```python
>>> "%s's game".format("askaholic")
"%s's game"
```

Use `"{}'s game".format(playername)` instead:
```python
>>> "{}'s game".format("askaholic")
"askaholic's game"

`

Comment (Askaholic) [reply tests/unit_tests/test_lobbyconnection.py]: As far as testing is concerned it would actually be better to assert that the game name is equal to some hard coded string. Otherwise if you have a bug in your formatting code, the test won't even catch it. 

```python
assert lobbyconnection.game_connection.game.name == "Dummy's game"
```

PR #393, opened 2019-01-06
Check game validity only if it's currently already valid.

Rackover:
I see frequently in the server console games being unranked twice. Once at start because of BAD_MOD, then once at the end because of MUTUAL_DRAW or whatever.
If the game is already unranked, no need to overwrite the unrank statement with another unrank statement. It does not provide additional info and it's superfluous.

With this commit, the validity state only gets eventually written to invalid if it was valid at first.


server/games/game.py
```diff
@@ -358,26 +358,28 @@ def add_game_connection(self, game_connection):
...
-                    return
-
-                if len(self._results) == 0:
-                    await self.mark_invalid(ValidityState.UNKNOWN_RESULT)
-                    return
-
-                await self.persist_results()
-                await self.rate_game()
-                await self._process_pending_army_stats()
+                if self.validity is ValidityState.VALID:
```
Askaholic: You should invert your logic which makes it easier to read.
```suggestion
                if self.validity is not ValidityState.VALID:
                    return
```

PR #400, opened 2019-02-11
[Refactor] Remove unnecessary dependencies between services

Askaholic:
Trying to decouple the services as much as possible. There are a lot of cyclic dependencies that should be removed. Some are handled as weak references, but most are owning such as `LobbyConnection` <-> `GameConnection`

### DIFF
--- a/server/lobbyconnection.py
+++ b/server/lobbyconnection.py
@@ -53,7 +53,6 @@ class LobbyConnection():
     @timed()
     def __init__(
         self,
-        context,
         games: GameService,
         players: PlayerService,
         geoip: GeoIpService,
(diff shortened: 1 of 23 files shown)


server/lobbyconnection.py
```diff
@@ -786,38 +790,45 @@ def command_game_host(self, message):
...
-        # FIXME: Setting up a ridiculous amount of cyclic pointers here
+        # TODO: Fix setting up a ridiculous amount of cyclic pointers here
         if self.game_connection:
             self.game_connection.abort("Player launched a new game")
-        self.game_connection = GameConnection(self.loop,
-                                              self,
-                                              self.player_service,
-                                              self.game_service)
-        self.game_connection.player = self.player
-        self.player.game_connection = self.game_connection
```
Askaholic: Need to add this back in

Askaholic: Probably best to do it in the constructor of GameConnection since that constructor now takes the player as an argument anyways.

PR #404, opened 2019-02-26
Fixup api session exception handling

Wesmania:


server/stats/achievement_service.py
```diff
@@ -95,7 +95,11 @@ def __init__(self, api_accessor: ApiAccessor):
         Else, it returns None
         """
         self._logger.debug("Updating %d achievements", len(queue))
-        response, content = await self.api_accessor.update_achievements(queue, player_id)
+        try:
```
Askaholic: Wouldn't it be better to do the error handling inside of `APIAccessor` instead of achievements service? That way there the error handling won't need to be duplicated everywhere where api_accessor is used

Wesmania: We return a response / content pair and don't have a meaningful return value if we get an exception. If we return None, we have to check it everywhere anyway. We can hack in some error-like response/contant pair, but that's ugly and *still* needs manual checking.

server/api/api_accessor.py
```diff
@@ -28,47 +28,45 @@ def __init__(self):
...
                 config.API_BASE_URL,
                 config.API_TOKEN_URI
             )
-
-        elif exc_type == SSLError:
+            raise ConnectionError
+        except SSLError:
             self.logger.error("The certificate verification failed while connecting the API")
+            raise ConnectionError
+        except Exception as e:   # TODO do we need this?
```
Askaholic: I think the answer to this question is 'no'

server/stats/achievement_service.py
Askaholic: Ah, yea that's true.

server/api/api_accessor.py
Wesmania: @Rackover disagreed on that one.

Rackover: ATM I think it's better if we log it as a general exception and eventually if we see a specific exception occurs frequently as part of normal functioning we can add the except case

Askaholic: Well then remove the todo comment

Rackover:
Fixes https://github.com/FAForever/server/pull/404/commits/709e51852f45596fd99ab8df8ed3e18228b2f1db

server/api/api_accessor.py
Askaholic: Sounds to me like you have a good reason for what you’re doing so there’s no need to add a comment bringing it into question

server/api/api_accessor.py
```diff
@@ -28,47 +28,45 @@ def __init__(self):
...
-                    client_id=API_CLIENT_ID,
-                    client_secret=API_CLIENT_SECRET
-                )
-            except:
-                self.__exit__(*sys.exc_info())
-
-        return self.session
-
-    def __exit__(self, exc_type, exc_val, exc_tb):
-        if exc_type == MissingTokenError:
```
norraxx: What's wrong with context manager?

server/stats/event_service.py
```diff
@@ -60,7 +60,11 @@ def __init__(self, api_accessor: ApiAccessor):
         Else, returns None
         """
         self._logger.debug("Recording %d events", len(queue))
-        response, content = await self.api_accessor.update_events(queue, player_id)
+        try:
+            response, content = await self.api_accessor.update_events(queue, player_id)
+        except ConnectionError:
+            self._logger.error("Failed to update events: connection error")
+            return None
```
norraxx: Let's catch exception with some info and raise exception with no info.

server/stats/achievement_service.py
norraxx: :cactus:

server/api/api_accessor.py
Wesmania: It is of no use here. Context managers are good when you allocate a resource and want to make sure you clean it up when you're done, but here we're creating a resource then keeping it around forever. Grabbing it in a context manager then making special checks to ensure it doesn't do what it was designed to do is pointless, especially since a try-catch block does _exactly_ what we want to do.

server/stats/event_service.py
Wesmania: Error type was logged further down the stack. Besides this function already returned None in some cases where things went wrong - do you want to refactor all its callers, too?

Issue #406, opened 2019-03-02
The server freezes waiting for the API to answer


Rackover:
When updating events or achievements, should the API take 10 minutes to answer, the python server will become unresponsive during that 10 minutes. The code handling the ach/evnt updates should be async

Askaholic:
Maybe we can try calling the request library functions with `run_in_executor`.

https://stackoverflow.com/questions/22190403/how-could-i-use-requests-in-asyncio

PR #409, opened 2019-03-08
Request is no longer blocking

Rackover:
Closes https://github.com/FAForever/server/issues/406

### DIFF
--- a/server/api/api_accessor.py
+++ b/server/api/api_accessor.py
@@ -4,6 +4,7 @@ from oauthlib.oauth2 import BackendApplicationClient
 from oauthlib.oauth2.rfc6749.errors import MissingTokenError, InsecureTransportError
 from requests.exceptions import SSLError
 from requests_oauthlib import OAuth2Session
+import asyncio
 import sys
 
 import server.config as config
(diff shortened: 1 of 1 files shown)


server/api/api_accessor.py
```diff
@@ -98,11 +99,23 @@ def __init__(self):
 
     async def api_get(self, path):
         with self.api_session as api:
-            result = api.get(API_BASE_URL + path)
+            loop = asyncio.get_event_loop()
```
Askaholic: Could refactor to util function, something like:

```python
async def run_in_executor(fn):
    loop = asyncio.get_event_loop()
    future = loop.run_in_executor(None, fn)
    return await future
```

Rackover:
Impossible, the token fetching is almost instantaneous, what is blocking here is the API achievements/events call

Rackover:
Plus I'm not sure I can `with await self.api_session as api:`

Askaholic:
You would have to use the async syntax: `async with self.api_session`

Sheeo:
Might I suggest using [aiohttp](https://aiohttp.readthedocs.io/en/stable/) instead?

It's already in `requirements.txt`.

Askaholic:
@Sheeo Sure, if someone has the motivation to write the oauth2 implementation.

Rackover:
Superseded by https://github.com/FAForever/server/pull/433

PR #410, opened 2019-03-09
Ladder map repetition fix

Askaholic:
When translating my query to expression language I managed to forget the order by clause, meaning it will just filter by an arbitrary choice of maps that you played within the last day. I also decided to use the game start time instead of the player score time, because it should be more reliable.

Can't seem to get transactions in tests working. Pytest always hangs whenever I try.


tests/unit_tests/test_ladder.py
```diff
@@ -85,3 +85,24 @@ def test_inform_player(ladder_service: LadderService):
...
+                )
+            await conn.execute(
+                f"""insert into game_player_stats
+                    (gameId, playerId, AI, faction, color, team, place, mean, deviation, scoreTime)
+                    values ({game_id_start+i}, 1, 0, 0, 0, 2, 0, 1500, 500, NOW() + interval {i+1} minute)
+                """)
+
+    history = await ladder_service.get_ladder_history(players.hosting)
+
+    assert history == list(reversed(range(num_maps-3, num_maps)))
```
Rackover: Shouldn't that arbitrary `3` become a config.py parameter for both tests and code ? We could want to put it to 5 or 7 in the future - or back to zero.

tests/unit_tests/test_ladder.py
```diff
@@ -85,3 +85,24 @@ def test_inform_player(ladder_service: LadderService):
...
+            await conn.execute(
+                f"""insert into game_stats
+                    (id, startTime, gameType, gameMod, host, mapId, gameName, validity)
+                    values ({game_id_start+i}, NOW() + interval {i+1} minute, '0', 6, 1, {i}, 'MapRepitition', 0)"""
+                )
+            await conn.execute(
+                f"""insert into game_player_stats
+                    (gameId, playerId, AI, faction, color, team, place, mean, deviation, scoreTime)
+                    values ({game_id_start+i}, 1, 0, 0, 0, 2, 0, 1500, 500, NOW() + interval {i+1} minute)
+                """)
```
Rackover: If you could add comments on these ones it would be great - they're kind of hard to read and i'm not sure of what they do

PR #411, opened 2019-03-10
ICE adapter support

Rackover:
Replacement of


server/gameconnection.py
```diff
@@ -535,6 +482,9 @@ def abort(self, logspam=''):
     def address_and_port(self):
         return "{}:{}".format(self.player.ip, self.player.game_port)
 
+    def address_and_port(self):
```
Askaholic: Duplicate

server/gameconnection.py
```diff
@@ -417,6 +349,21 @@ def send_message(self, message):
                 (teamkiller_id, victim_id, self.game.id, gametime)
             )
 
+    async def handle_ice_message(self, receiver_id, ice_msg):
+        peer = self.player_service.get_player(receiver_id)
```
Askaholic: Make sure `receiver_id` is an int

```suggestion
        reveiver_id = int(receiver_id)
        peer = self.player_service.get_player(receiver_id)
```

server/gameconnection.py
```diff
@@ -417,6 +349,21 @@ def send_message(self, message):
...
+        if not peer:
+           self._logger.info("Ignoring ICE message for unknown player: %s", receiver_id)
+           return
+
+        game_connection = peer.game_connection
+        if not game_connection:
+            self._logger.info("Ignoring ICE message for player without game connection: %s", receiver_id)
+            return
+
+        game_connection.send_message(dict(command="IceMessage", args=[int(self.player.id), ice_msg]))
```
Askaholic: Use `{}` syntax. Why are we calling `int` on `player.id` isn't that an int already?

server/gameconnection.py
```diff
@@ -166,91 +164,24 @@ def send_message(self, message):
...
-            # Opposite of the above
-            remote_addr = remote_addr.result()
-            own_addr = await self.connectivity.ProbePeerNAT(peer, use_address=remote_addr)
-        for p in pending:
-            if not p.done():
-                p.cancel()
-        return own_addr, remote_addr
+        self.send_ConnectToPeer(peer.player.login,
+                                peer.player.id,
+                                True)
```
Askaholic: What does this boolean mean? Use kwarg

server/gameconnection.py
```diff
@@ -166,91 +164,24 @@ def send_message(self, message):
...
+        peer.send_ConnectToPeer(self.player.login,
+                                self.player.id,
+                                False)
```
Askaholic: What does this boolean mean? Use kwarg

server/gameconnection.py
```diff
@@ -166,91 +164,24 @@ def send_message(self, message):
...
-                           peer.player.login,
+        self.send_JoinGame(peer.player.login,
                            peer.player.id)
-        peer.send_ConnectToPeer(own_addr,
-                                self.player.login,
-                                self.player.id)
+
+        peer.send_ConnectToPeer(self.player.login,
+                                self.player.id,
+                                True)
```
Askaholic: What does this boolean mean? Use kwarg

server/__init__.py
```diff
@@ -47,17 +47,18 @@
 stats = None
 
 if not config.ENABLE_STATSD:
-    from . import fake_statsd
-    stats = fake_statsd.DummyConnection()
+    from unittest import mock
+    stats = mock.MagicMock()
```
Askaholic: Somehow this feels wrong to me. Using unittest mock in actual code?

server/gameconnection.py
```diff
@@ -563,5 +505,6 @@ def __str__(self):
     "Rehost":               GameConnection.handle_rehost,
     "Bottleneck":           GameConnection.handle_bottleneck,
     "BottleneckCleared":    GameConnection.handle_bottleneck_cleared,
-    "Disconnected":         GameConnection.handle_disconnected
+    "Disconnected":         GameConnection.handle_disconnected,
+    "IceMsg":           GameConnection.handle_ice_message
```
Askaholic: Line up the indents please

server/ice_servers/coturn.py
```diff
@@ -0,0 +1,45 @@
...
+
+class CoturnHMAC:
+    def __init__(self, coturn_hosts=None, coturn_keys=None):
+        self.coturn_hosts = coturn_hosts or COTURN_HOSTS
+        self.coturn_keys = coturn_keys or COTURN_KEYS
+
+    def fetch_token(self, username='faf-user', ttl=None):
+        servers = []
+
+        ttl = ttl or TWILIO_TTL
```
Askaholic: Can't `TWILIO_TTL` just become the default for the kwarg?

```python
def fetch_token(self, username='faf-user', ttl=TWILIO_TTL):
```

server/ice_servers/coturn.py
```diff
@@ -0,0 +1,45 @@
...
+        if ttl is not None:
+            ttl = int(ttl)
+        else:
+            ttl = 3600*24
```
Askaholic: What? Why is the default hardcoded in here? Especially after a block of code which ensures that this branch will NEVER happen!

server/ice_servers/coturn.py
```diff
@@ -0,0 +1,45 @@
...
+
+        # See https://github.com/coturn/coturn/wiki/turnserver#turn-rest-api
+        # create hmac of coturn_key + timestamp:username
```
Askaholic: This comment tells me that the following code should be extracted to a function for readability. You could probably make it a static method

server/ice_servers/nts.py
```diff
@@ -0,0 +1,43 @@
...
+        self.twilio_token = token or TWILIO_TOKEN
+        self.client = TwilioRestClient(self.twilio_account_sid, self.twilio_token)
+
+    async def fetch_token(self, ttl=None):
+        """
+        Fetches token from Twilio
+
+        :param ttl - ttl in seconds or None for default of 24h
+        """
+        ttl = ttl or TWILIO_TTL
```
Askaholic: Again why is the default being created like this?

server/lobbyconnection.py
```diff
@@ -51,23 +51,25 @@ def __init__(self, message, *args, **kwargs):
...
 class LobbyConnection():
     @timed()
-    def __init__(
-        self,
-        games: GameService,
-        players: PlayerService,
-        geoip: GeoIpService,
-        matchmaker_queue: MatchmakerQueue
-    ):
+    def __init__(self,
```
Askaholic: Please fix indents to how the were before

server/lobbyconnection.py
```diff
@@ -51,23 +51,25 @@ def __init__(self, message, *args, **kwargs):
...
+                 games: GameService,
+                 players: PlayerService,
+                 nts_client: TwilioNTS,
+                 geoip: GeoIpService,
+                 matchmaker_queue: MatchmakerQueue):
+        super(LobbyConnection, self).__init__()
```
Askaholic: LobbyConnection has no superclass

server/lobbyconnection.py
```diff
@@ -51,23 +51,25 @@ def __init__(self, message, *args, **kwargs):
...
+
         self.geoip_service = geoip
         self.game_service = games
-        self.player_service = players
+        self.player_service = players  # type: PlayerService
```
Askaholic: Type annotations are already handled in kwargs, no need to comment here.

server/lobbyconnection.py
```diff
@@ -542,15 +544,18 @@ def check_version(self, message):
         self.player = Player(login=str(login),
```
Askaholic: Can we fix indents here

server/lobbyconnection.py
```diff
@@ -628,6 +633,28 @@ def check_version(self, message):
...
+            self.send_warning("The game you were connected to does no longer exist")
+            return
+
+        game = self.game_service[game_id]  # type: Game
+        if game.state != GameState.LOBBY and game.state != GameState.LIVE:
+            self.send_warning("The game you were connected to is no longer available")
+            return
+
+        self._logger.debug("Restoring game session of player %s to game %s", self.player, game)
+        self.game_connection = GameConnection(self.loop, self, self.player_service, self.game_service, self.player,
```
Askaholic: Let's use kwargs for GameConnection constructor. Some of these aren't needed anymore. (Also fix indents)

server/lobbyconnection.py
```diff
@@ -628,6 +633,28 @@ def check_version(self, message):
...
+                                              game, GameConnectionState.CONNECTED_TO_HOST)
+        self.player.game_connection = self.game_connection
```
Askaholic: GameConnection constructor handles this now.

server/lobbyconnection.py
```diff
@@ -888,6 +893,21 @@ def launch_game(self, game, port, is_host=False, use_map=None):
             else:
                 raise ValueError('invalid type argument')
 
+    @asyncio.coroutine
+    async def command_ice_servers(self, message):
+        if self.player:
+
+            ttl = 24 * 3600
```
Askaholic: Here's that mysterious hardcoded constant again

server/ice_servers/coturn.py
Geosearchef: it doesn't ensure this will never happen, TWILIO_TTL can be none.

server/ice_servers/coturn.py
Geosearchef: it is already extracted to a function, everything this function does is generate the long term credential

server/lobbyconnection.py
Geosearchef: I don't see an issue with that, it's probably a specified constant in the turn long term cred mechanism rfc, I wouldn't introduce a constant/variable for that as it will never change and its meaning is obvious.

Askaholic: I think your comment just demonstrated exactly why it is an issue. No one besides the original author knows what the constant means. It should be extracted to a named constant for readability. Even if it's something like TWILLIO_DEFAULT_TTL. If it is in fact from an RFC, then you could even put a comment with a reference to the RFC at the place where that constant is defined. Also the whole point of constants is that they are... constant, they don't change.

https://www.itjungle.com/2004/08/18/fhg081804-story01/

server/lobbyconnection.py
Askaholic: Lets put the first parameter on a new line and line up the remaining parameters to that (see the deletions in this diff). IMO this is nicer because it creates consistent indentations (even if the function names have different lengths).

server/ice_servers/coturn.py
Askaholic: I see. Well it actually can't be `None` but it could be empty string, since it's defined as:
```python
TWILIO_TTL = os.getenv("TWILIO_TTL", 86400)
```
(again note the mysterious default constant, only it's written differently this time)

I don't think that we need to do error handling for the environment variables being set up wrong. And even if we do, that error handling should be done in `config.py` not in `coturn`. This file should assume that TWILIO_TTL has been set to a default correctly.

server/ice_servers/coturn.py
Askaholic: So, this comment is purely a "Clean code" nit pick. Although reading this code more closely I think the function name is misleading.

This function does about 5 things:
1. A bunch of weird (IMO unnecessary) logic for determining what the ttl should be
2. Iterates through all of the coturn hosts:
    1. Computes the hmac for each host
    2. Computes the host urls
3. Aggregates all of the computed keys and hosts to a single list and returns that list

Ideally the contents of the loop would go into their own function which actually calculates ONE token. This function is called `fetch_token` which implies that it will return ONE token. But it doesn't, it returns a list of dictionaries which contain a bunch of information, among other things, several tokens.

Geosearchef: yeah, we could just extract the generation of a single credential, and I agree, that function name is chosen very badly

server/config.py
```diff
@@ -59,8 +59,8 @@
 TWILIO_TOKEN = os.getenv("TWILIO_TOKEN", "")
 TWILIO_TTL = os.getenv("TWILIO_TTL", 86400)
 
-COTURN_HOSTS = os.getenv('COTURN_HOSTS', "").split(',')
-COTURN_KEYS = os.getenv('COTURN_KEYS', "").split(',')
+COTURN_HOSTS = os.getenv('COTURN_HOSTS', "geosearchef.de").split(',')
+COTURN_KEYS = os.getenv('COTURN_KEYS', "wop0CC7EUxrrcavH7aKCvaUjmpT3ZwW7N5ozYI7Jps").split(',')
```
Geosearchef: yes, I've seen what I did there, this shouldn't have been published, the file needs to be reverted

server/lobbyconnection.py
```diff
@@ -55,27 +58,30 @@ def __init__(
...
         players: PlayerService,
+        nts_client: Optional[TwilioNTS],
         geoip: GeoIpService,
         matchmaker_queue: MatchmakerQueue
     ):
         self.geoip_service = geoip
         self.game_service = games
         self.player_service = players
+        self.nts_client = nts_client
+        self.coturn_generator = CoturnHMAC()
```
Askaholic: Why do we create a new `CoturnHMAC` for every connection, but only have a single `TwilioNTS`? Couldn't `CoturnHMAC` just be a singleton as well?

tests/unit_tests/test_lobbyconnection.py
```diff
@@ -440,3 +444,65 @@ def test_command_admin_closeFA(mocker, lobbyconnection):
 
     player.lobby_connection.send_warning.assert_called_with("This is a test message")
     tuna.lobby_connection.send_warning.assert_called_with("This is a test message")
+
+async def test_game_connection_not_restored_if_no_such_game_exists(lobbyconnection: LobbyConnection, mocker, mock_player):
+    protocol = mocker.patch.object(lobbyconnection, 'protocol')
+    lobbyconnection.player = mock_player
+    lobbyconnection.player.game_connection = None
+    lobbyconnection.player.state = PlayerState.IDLE
+    lobbyconnection.command_restore_game_session({'game_id': 123})
```
Askaholic: We should be calling `lobbyconnection.on_message_received` here instead

tests/unit_tests/test_lobbyconnection.py
```diff
@@ -440,3 +444,65 @@ def test_command_admin_closeFA(mocker, lobbyconnection):
...
+    lobbyconnection.player.state = PlayerState.IDLE
+    lobbyconnection.game_service = game_service
+    game = mock.create_autospec(Game(42, game_service, game_stats_service))
+    game.state = game_state
+    game.password = None
+    game.game_mode = 'faf'
+    game.id = 42
+    game_service.games[42] = game
+
+    lobbyconnection.command_restore_game_session({'game_id': 42})
```
Askaholic: Same as above

tests/unit_tests/test_lobbyconnection.py
```diff
@@ -440,3 +444,65 @@ def test_command_admin_closeFA(mocker, lobbyconnection):
...
```
Askaholic: Same as above

Rackover:
Everything OK, has been tested successfully.
Merging in.

Askaholic:
@Rackover Seriously?! You guys didn't address any of the issues that I mentioned like a month ago. This PR also doens't add tests for ANYTHING ICE RELATED! It actually mostly removes tests. This is not a finished PR, it needs thorough regression tests to ensure that this stuff doesn't break in the future

PR #413, opened 2019-03-12
Integration tests update

Askaholic:
Make the integration tests actually check something. This will be an OK way of testing things which are otherwise hard to do with unit tests because of the mocks. For example checking that a game's default title is set correctly if the host doesn't specify a title.

Added integration tests for:
  * Queueing a ladder match
  * Players banned form ladder receive the 'you are banned' message
  * Hosting a game with missing or empty fields

I also added a test-data.sql file to this repo so that we don't need to insert data in the test themselves, and don't need to submit PR's to Faforever/db just to add some test data.

### DIFF
--- a/server/lobbyconnection.py
+++ b/server/lobbyconnection.py
@@ -759,33 +759,32 @@ class LobbyConnection():
 
         assert isinstance(self.player, Player)
 
-        default_title = "%s' game".format(self.player.login)
-        title = html.escape(message.get('title') or default_title)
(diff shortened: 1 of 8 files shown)


Rackover:
Does that mean https://github.com/FAForever/server/pull/392 is obsolete ?

Askaholic:
Yea, this covers the game title stuff.

Rackover:
Tested, workign, merging

PR #415, opened 2019-03-21
Issue/#414 Fix friends/foes not being saved in database

### LINKED ISSUE #414: Newly Added Friends and Foes are not saved
Any newly added friends and foes are not saved once the client is closed. You can add someone as a friend, restart the client, and they will not be on your friends list anymore. Same for foes. I tested in both python and java client.

Askaholic:
Fixes #414 

The fix was to remove the `timed()` decorator since it isn't designed to work on async functions. I took the opportunity to add a couple of tests and to rewrite the sql statements with sqlalchemy.

### DIFF
--- a/server/lobbyconnection.py
+++ b/server/lobbyconnection.py
@@ -236,9 +236,9 @@ class LobbyConnection():
 
     async def command_social_remove(self, message):
         if "friend" in message:
-            subject_id = message['friend']
+            subject_id = message["friend"]
         elif "foe" in message:
-            subject_id = message['foe']
+            subject_id = message["foe"]
         else:
             self.abort("No-op social_remove.")
             return
--- a/server/lobbyconnection.py
+++ b/server/lobbyconnection.py
@@ -13,6 +13,7 @@ import pymysql
 import semver
 import server
 import server.db as db
+from sqlalchemy import and_
 
 from . import config
 from .config import FAF_POLICY_SERVER_BASE_URL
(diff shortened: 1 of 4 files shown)


Sheeo:
Nice work @Askaholic.

@norraxx regarding the timed decorator -- from my cursory testing it does work, but the dispatcher code that calls the given functions here invokes `asyncio.iscoroutinefunction` to test whether to `await` or not, and so `timed()` functions are not `await`'ed.

The decorator can be patched to use the same check and return a coroutine in the case of wrapping a coroutine instead.

I think it's best to remove the use of the decorator entirely instead, however. Instead the dispatcher code should have an option to globally test for slow-running commands, ideally configured by the environment.

The timed decorator is cute, but having to add it everywhere is a bad idea and we don't always want it on.

Askaholic:
Ah, yea makes sense for the dispatcher to do the timing. The only down side is that it wouldn't allow you to add a custom time limit to each handler function. Then again maybe that wasn't a good choice in the first place; Either a function is slow or it's not. No reason to have differen't standards for slowness.

I think another solution is to make all handlers async and always await them. I don't think there's a downside.

tests/unit_tests/test_lobbyconnection.py
```diff
@@ -423,6 +427,51 @@ def test_command_admin_closeFA(mocker, lobbyconnection):
...
+        )
+
+        return [row['subject_id'] async for row in result]
+
+
+async def test_command_social_add_friend(lobbyconnection, mock_player, db_engine):
+    lobbyconnection.player = mock_player
+    lobbyconnection.player.id = 1
+    lobbyconnection._authenticated = True
+
```
Askaholic: Get friends here and assert that it's empty

server/lobbyconnection.py
```diff
@@ -234,33 +236,35 @@ def send_game_list(self):
...
+            await conn.execute(friends_and_foes.delete().where(and_(
+                friends_and_foes.c.user_id == self.player.id,
+                friends_and_foes.c.subject_id == subject_id
+            )))
 
-    @timed()
     async def command_social_add(self, message):
         if "friend" in message:
             status = "FRIEND"
-            target_id = message['friend']
```
Askaholic: Change quotes back

PR #419, opened 2019-03-22
Pass game mode as string again

Askaholic:
I changed this to an enum in a different PR, but then realized it was probably a bad idea since game_mode can be any featured mod from the database. This changes the game_mode back to a plain string but keeps the refactoring/cleanup that I also did in the other PR.

Also adds a default gamemode 'faf' if the client doesn't send one when hosting a game.


Askaholic:
Ran a labwars game with this on test server. Seems to work again

PR #421, opened 2019-03-25
Use pipenv instead of requirements.txt

Askaholic:
A big advantage is that this can separate prod and dev dependencies.

- Update to latest version of pytest
- Fix many deprecation warnings
- Remove unused dependency `rsa`


Pipfile
```diff
@@ -0,0 +1,32 @@
...
+"geoip2" = "*"
+msgpack-python = "*"
+trueskill = "*"
+aiocron = "*"
+marisa-trie = "*"
+oauthlib = "*"
+requests-oauthlib = "*"
+sqlalchemy = "*"
+
+[dev-packages]
```
norraxx: Isn't it packages-dev?

server/ladder_service.py
```diff
@@ -24,6 +24,10 @@ def __init__(self, games_service: "GameService"):
         self._informed_players = []
         self.game_service = games_service
 
+    async def sleep(self, secs: float):
+        "So we can mock sleeping only for this class"
```
norraxx: Use monkey patch in tests?

tests/unit_tests/test_achievement_service.py
```diff
@@ -21,7 +21,6 @@ def service(api_accessor: ApiAccessor):
     return AchievementService(api_accessor)
 
 
-@pytest.fixture()
```
norraxx: R u sure?
If yes, then that func isn't needed anymore

tests/unit_tests/test_event_service.py
```diff
@@ -19,7 +19,6 @@ def service(api_accessor: ApiAccessor):
     return EventService(api_accessor)
 
 
-@pytest.fixture()
```
norraxx: -//-

Pipfile
```diff
@@ -0,0 +1,32 @@
...
+pycrypto = "*"
+semver = "*"
+"geoip2" = "*"
+msgpack-python = "*"
+trueskill = "*"
+aiocron = "*"
+marisa-trie = "*"
+oauthlib = "*"
+requests-oauthlib = "*"
+sqlalchemy = "*"
```
norraxx: I would pin the package versions. Reason: stable environment for all.

Pipfile
Askaholic: Nope. This is pipenv generated file

Pipfile
Askaholic: That's what the lock file is for :)

norraxx: But if would like to make some stuff and add another library, my pipenv + environment will load another versions of libraries on `pipenv update/lock`.

Sheeo: @norraxx only if you specifically do a general `pipenv update`, otherwise the lock-file is respected as-is. You don't want to do that unless you want to  do that.

tests/integration_tests/conftest.py
```diff
@@ -28,7 +28,7 @@ def lobby_server(request, loop, player_service, game_service, geoip_service):
     server.NatPacketServer.instance = NatPacketServer(addresses=server.config.LOBBY_NAT_ADDRESSES, loop=loop)
     loop.run_until_complete(server.NatPacketServer.instance.listen())
     ctx = run_lobby_server(
-        address=('127.0.0.1', None),
+        address=('127.0.0.1', 8002),
```
Sheeo: Any particular reason for this?

Askaholic: Only because the logging statement that tried to print the port number was failing. I can probably undo this, since i don’t think that was the reason why the test was failing.

Sheeo: All right.

If it works without I think it's best to keep the port randomly assigned: It makes sure to use an unbound port this way, and we avoid faux errors where someone has another service running on it.

Sheeo:
Looks good to me 👍

Issue #19, opened 2015-03-20
Reintroduce matchmaker


Sheeo:
FAForever/client#181

PR #423, opened 2019-03-29
Party system, team matchmaking

### LINKED ISSUE #613: TMM: Get player faction from party
For queuing with a party it's going to be easiest if we move the faction selection to the server. When the party leader joins a queue, the server needs to read all of the players faction preferences, choose an appropriate faction and set the `Player.faction` attribute for every player. 

This should be added to the party system PR.

### LINKED ISSUE #632: Server allows queuing for ladder while already setting up a previous ladder match
During the time where players have found a match and are waiting for the match to be set up, the server will allow them to add themselves back into the queue a second time, potentially causing a situation where the server tries to make a single player join 2 games at the same time. 

We probably need to track setup time of automatch games as its own player state, or use an existing state to make sure players can't double join a queue in this way.

Related tasks:
- Add an integration test that simulates this scenario

### LINKED ISSUE #679: TMM: Choose factions at game launch time
Right now the factions are chosen when you enter into the queue. However we want to allow players to change their faction selection while waiting in queue, so factions should be chosen after a match is found.

Somehow the ladder service and party service need to communicate the faction choices.

Geosearchef:
Add party system.

TODO:
- [x] Remove `invalid_state` messages? These states are internal things and violations should probably generate ClientErrors instead (i.e. if a player tries to join a queue while they are still in a game, the client should be able to catch this and prevent it)
- [x] Rename all `queue` or `queue_name` fields to the same thing (pick one for consistency)
- [x] Clean up player state enum

Closes #613
Closes #632
Closes #679


Sheeo:
Please use the new draft pull request feature for WIP PR's

server/game_service.py
```diff
@@ -117,6 +118,9 @@ def __init__(self, player_service, game_stats_service):
             # meh meh
             self.ladder_service = LadderService(self)
 
+            # meh meh meh, TODO: pls explain why I am doing this here
+            self.team_matchmaking_service = TeamMatchmakingService()
```
Sheeo: There's no reason to do it here; you can just do it from the ctor.

server/team_matchmaker/player_party.py
```diff
@@ -0,0 +1,43 @@
...
+        self.team_matchmaking_service.player_parties.pop(player)
+
+        self.broadcastParty()
+        self.sendParty(player)
+
+    def broadcastParty(self):
+        for player in self._members:
+            self.sendParty(player)
+
+    def sendParty(self, player: Player):
```
Sheeo: Please consistently use `snake_case` and do not start intermixing `camelCase`

server/players.py
```diff
@@ -12,6 +12,7 @@ class PlayerState(IntEnum):
     HOSTING = 3,
     JOINING = 4,
     SEARCHING_LADDER = 5,
+    IN_PARTY = 6,
```
Sheeo: I question the usefulness of this new disjoint state -- can a player not be both `PLAYING` and `IN_PARTY`?

I realize the same is true for `SEARCHING_LADDER`, but this smells like a clue of state-bloat. The primary reason this is here is to quickly determine valid protocol elements; e.g. no GameConnection messages are allowed if the state isn't either `PLAYING` or `HOSTING`. It seems like a bit-field/set of states is more appropriate though?

server/team_matchmaker/player_party.py
```diff
@@ -0,0 +1,43 @@
...
+        player.lobby_connection.send({
+            "command": "update_party",
+            "owner": self.owner,
+            "members": {m.id for m in self._members}
+        })
+
+    def isDisbanded(self) -> bool:
+        return len(self._members) <= 1 or self.owner not in self._members
```
Sheeo: So a party with just the owner is disbanded?

server/team_matchmaker/player_party.py
```diff
@@ -0,0 +1,43 @@
...
+
+    def disband(self):
+        for member in self.members:
+            self.remove_player(member)
```
Sheeo: Due to the semantics of `remove_player`, this broadcasts a party update for each removal of a member. It would be better to disband the party fully and do a single broadcast

server/team_matchmaking_service.py
```diff
@@ -0,0 +1,117 @@
...
+@with_logger
+class TeamMatchmakingService:
+    """
+    Service responsible for managing the global team matchmaking. Does grouping, matchmaking, updates statistics, and
+    launches the games.
+    """
+
+    def __init__(self, games_service: GameService):
+        self.game_service = games_service
+        self.player_parties = dict() # player -> current party
```
Sheeo: I would annotate this on the class as follows:

`player_parties: Dict[Player, PlayerParty]`

server/team_matchmaking_service.py
```diff
@@ -0,0 +1,117 @@
...
+        self._pending_invites = dict() # invited player -> current pending invite to this player
```
Sheeo: Same for type annotation as above; though the type of pending invite is unclear.

Do you imply with this that a player can only have a single pending invite? What if they are invited to multiple parties?

server/team_matchmaking_service.py
```diff
@@ -0,0 +1,117 @@
...
+
+    def invite_player_to_group(self, inviting_player: Player, invited_player: Player):
```
Sheeo: Is party now "group"?

server/lobbyconnection.py
```diff
@@ -888,6 +888,31 @@ def launch_game(self, game, port, is_host=False, use_map=None):
             else:
                 raise ValueError('invalid type argument')
 
+    async def command_invite_to_party(self, message):
+        invited_player = self.player_service.get_player(message["invited_player_id"])
```
Sheeo: Isn't it simpler to use words like `sender` and `target` or `recipient`?

server/team_matchmaking_service.py
```diff
@@ -0,0 +1,117 @@
...
+        if inviting_player not in self.player_parties:
+            self.player_parties[inviting_player] = PlayerParty(self, inviting_player)
+
+        inviting_party = self.player_parties.get(inviting_player)
+
+        if not inviting_party.owner == inviting_player:
```
Sheeo: I think `party` as a name suffices, given how the "inviting" bit is already inferred by there being an `inviting_player` above.

server/team_matchmaking_service.py
```diff
@@ -0,0 +1,117 @@
...
+            raise ClientError("You do not own this party.", recoverable=True)
+
+        if inviting_player not in invited_player.friends:
+            raise ClientError("This person hasn't befriended you.", recoverable=True)
+
+        self.clear_invites_to(invited_player)
```
Sheeo: As I guessed above; this means you can clear pending invites and theoretically DoS somebody out of party invites. Would it not be better to have a set of pending invites per player, which are cleared on acceptance rather than on creation of a new invite?

server/team_matchmaking_service.py
```diff
@@ -0,0 +1,117 @@
...
+
+        for party in disbanded_parties:
+            self.remove_party(party)
+
+    # - accept group invite:  client->server
+    # - invite to group:  client->server
+    # - kick from group:  client->server
+    # - leave group:  client->server
+    #
+    # - group update:  server->client
```
Sheeo: These comments ought to go in the protocol description markdown

Sheeo:
For tracking reasons see #19

server/players.py
Geosearchef: I agree, there needs to be some work to be done on this.it's for blocking joining a game,... while in queue.

We should probably use something like INQUEUE. Although there are many open questions like what do we do with the group after they are in game/finished, do we kick the player from the grouo on joining a game or just Block queueing. And most importantly, on matching, do we ask players if the are ready?

server/team_matchmaker/player_party.py
Geosearchef: You don't want to know how often I caught myself using camelcase ;), it's just natural to me and I don't notice it.

server/team_matchmaker/player_party.py
Geosearchef: That's was the plan, but actually that doesn't make any sense as the owner is alone when he makes the first invite and the party get's created. Also, I should probably handle offline players.

Rackover:
❤
I'm leaving this heart here because while I can't review this PR at the moment, I want to let you know how thankful I am for all the amazing work you're doing

server/team_matchmaking_service.py
Geosearchef: I do currently imply this to prevent leaking pending invites and overfloading the user with invites.

server/team_matchmaking_service.py
Geosearchef: Thx for catching that, at first it was named group

server/team_matchmaking_service.py
Geosearchef: It could also be the party the invited person is already in :)

server/team_matchmaking_service.py
Geosearchef: Yes, I thought about that aswell, if you allow multiple invites you can still dos a user out of a party by flooding them with notifications. They will foe you this way or the other and then you're safe.

I'm also now considering removing the requirement of being befriended by a person to invite them and replace it with not being foed

server/team_matchmaking_service.py
Geosearchef: Yeah, these are just notes left there after realizing my notes are gone in slack history.

Geosearchef:
Scoring/rating is intended to apply to the normal global rating as the intention of this was to get people out of playing single maps and make them try other maps out of a map people while reducing the lobby sim aspect.

server/team_matchmaking_service.py
Sheeo: You just want a uniqueness constraint on behalf of the sender per target -- that is, you cannot invite a user multiple times to the same party?

server/__init__.py
```diff
@@ -45,7 +46,8 @@
     'games',
     'control',
     'abc',
-    'protocol'
+    'protocol',
+    'TeamMatchmakingService'
```
Askaholic: Can we alphabetize these, and keep all the stuff with capital letters next to each other?

server/game_service.py
Askaholic: We should create services in `server.py`. There might be a better way to do this in the future, but for now putting them all in `server.py` at least ensures that all service creation happens in one place. 

Definitely do not create it in `update_data` here because that will cause the service to be re-created about every 10 minutes. (I told Sheeo already, but I can move LadderService to `server.py` once I find some time, so you don't need to worry about that)

server/lobbyconnection.py
```diff
@@ -888,6 +888,31 @@ def launch_game(self, game, port, is_host=False, use_map=None):
...
+        if invited_player is None:
+            raise ClientError("The invited player doesn't exist", recoverable=True)
+        else:
```
Askaholic: The `else` is unnecessary since raising an exception will exit the function.

server/lobbyconnection.py
```diff
@@ -888,6 +888,31 @@ def launch_game(self, game, port, is_host=False, use_map=None):
...
+
+    async def command_kick_player_from_party(self, message):
+        kicked_player = self.player_service.get_player(message["kicked_player_id"])
+        if kicked_player is None:
+            raise ClientError("The kicked player doesn't exist", recoverable=True)
+        else:
+            self.game_service.team_matchmaking_service.kick_player_from_party(self.player, kicked_player)
+
+    async def command_leave_party(self):
+        self.game_service.team_matchmaking_service.leave_party(self.player)
```
Askaholic: I think it would be better to remove some feature envy here and have `LobbyConnection` take `TeamMatchmakingService` in the constructor. Then we can just do `self.team_matchmaking_service...`

server/team_matchmaker/player_party.py
```diff
@@ -0,0 +1,43 @@
+from server.players import Player
+
+
+class PlayerParty:
+    def __init__(self, team_matchmaking_service: "TeamMatchmakingService", owner: Player):
```
Askaholic: Lets try our best to avoid the cyclic dependency of `TeamMatchmakingService <-> PlayerParty` if possible. Matchmmaking service can add parties to it's own member variables, and I think we should figure out a different way of passing messages to the player so that PlayerParty doesn't need to know about the lobbyconnection or protocol at all. Basically try to decouple this a bit more so it can function even if we remove lobbyconnection

server/lobbyconnection.py
Geosearchef: true, I'm busy grasping python and the server structure, would have never done such a thing in java ;)

server/lobbyconnection.py
Geosearchef: hmm, yeah, should've done that, gonna move teamMatchmakingService (notice me automatically writing this in camel case :) ) out of game conn to server.py, I was a bit shocked myself back then how ugly this line looked but went on doing sth else first

server/team_matchmaker/player_party.py
Geosearchef: teammatchmakingservice is a service, PlayerParty is an object. I don't see any issue with the objects managed by a service having a reference to call back that service. In this case it makes it extremly easy to trigger certain actions on adding/remove players. Also the team matchmaking service doesn't have to send data each time it changes something about the groups, the group updates are sent automatically.

server/team_matchmaking_service.py
```diff
@@ -0,0 +1,120 @@
...
+        recipient.lobby_connection.send({
+            "command": "party_invite",
+            "sender": sender.id
+        })
+
+    def accept_invite(self, recipient: Player, sender: Player):
+        if recipient not in self._pending_invites:
+            raise ClientError("You're not invited to a party", recoverable=True)
+
+        pending_invite = self._pending_invites.get((sender, recipient))
```
Askaholic: Maybe this should be combined with the line further down:
```python
if recipient in self._pending_invites:
    self._pending_invites.pop((sender, recipient))
```

Since right now if `(sender, recipient)` doesn't exist in the dict, `get()` will return `None` and you will get an `AttributeError` for `pending_invite.sender` on the following line.

Askaholic: Oh, just noticed you check for this existence on the previous line. So maybe you could do something like this:
```python
pending_invite = self._pending_invites.pop((sender, recipient), None)
if pending_invite is None:
    raise ClientError(...)
```

Or maybe this is better:
```python
if recipient not in self._pending_invites:
     raise ClientError("You're not invited to a party", recoverable=True)

pending_invite = self._pending_invites.pop((sender, recipient))
```

And then you don't need the extra existence check at the bottom

Geosearchef: which is why I check if (sender, recipient) is in self._pending_invites 2 lines above. Which I actually forgot to update when switching from recipient key to tuple key, so good you catched that

Askaholic: Hmm I see some cases where those suggestions might not work correctly. I guess maybe you just replace the `.get()` with a bracket operator, since you already check for existence so it will never raise a `KeyError` anyways

Geosearchef: the extra existence check is void anyways :)

Geosearchef: Which cases do you see, I just removed it? If the party doesn't exist, I want to remove the invite anyways, and checking if the sender sent the invite is also void due to the sender now being part of the key of the map.

server/team_matchmaker/player_party.py
Askaholic: Well here are specifically the things that bother me:
1. This function is directly modifying a member variable of `team_matchmaking_service` (feature envy again). Seems like it would be better for `team_matchmaking_service` to have a function called `add_player_to_party` which does this modification and then calls `party.add_player`.
```python
self.team_matchmaking_service.player_parties[player] = self
```

2. This object is directly coupled to the implementation of `LobbyConnection` by this line. Obviously this is convenient for quickly sending a message to the player, but it's bad design. Stuff like this is the main cause of the mess in the rest of the code, so I'd like to not contribute further to that mess. I'm not sure what the best way of handling this is, but maybe a callback function would be ok? A callback still doesn't feel elegant to me, but it feels a little better than directly coupling to lobbyconnection code. I will keep thinking about possible solutions to this.
```python
player.lobby_connection.send({
    "command": "update_party",
    "owner": self.owner,
     "members": {m.id for m in self._members}
})
```

server/lobbyconnection.py
```diff
@@ -888,6 +891,31 @@ def launch_game(self, game, port, is_host=False, use_map=None):
...
+            raise ClientError("The inviting player doesn't exist", recoverable=True)
+
+        self.team_matchmaking_service.accept_invite(self.player, sender)
+
+    async def command_kick_player_from_party(self, message):
+        kicked_player = self.player_service.get_player(message["kicked_player_id"])
+        if kicked_player is None:
+            raise ClientError("The kicked player doesn't exist", recoverable=True)
+
+        self.game_service.team_matchmaking_service.kick_player_from_party(self.player, kicked_player)
```
Askaholic: Missed on here

server/team_matchmaking_service.py
Askaholic: Looks good to me now

server/team_matchmaking_service.py
```diff
@@ -0,0 +1,117 @@
+from typing import NamedTuple
+
+from server import GameService
+from server.decorators import with_logger
+from server.lobbyconnection import ClientError
+from server.players import Player
+from server.team_matchmaker.player_party import PlayerParty
+
+MapDescription = NamedTuple('Map', [("id", int), ("name", str), ("path", str)])
```
Askaholic: No need to duplicate this. (As a side note the anti map repetition code should work with an arbitrary number of players, that is assuming it works in the first place :P )

server/team_matchmaking_service.py
```diff
@@ -0,0 +1,115 @@
...
+        party = self.player_parties.get(sender)
+
+        if not party.owner == sender:
+            raise ClientError("You do not own this party.", recoverable=True)
+
+        if sender in recipient.foes:
+            raise ClientError("This person doesn't accept invites from you.", recoverable=True)
+
+        self._pending_invites[(sender, recipient)] = GroupInvite(sender, recipient, party, time.time())
+        recipient.lobby_connection.send({
```
Askaholic: IMO, this should happen in LobbyConnection (it should send this message right after calling `invite_player_to_party`. Again, trying to decouple the code

server/team_matchmaking_service.py
```diff
@@ -0,0 +1,115 @@
...
+            raise ClientError("You're not invited to a party", recoverable=True)
+
+        pending_invite = self._pending_invites.pop((sender, recipient))
+
+        if pending_invite.party not in self.player_parties:
+            raise ClientError("The party you're trying to join doesn't exist anymore.", recoverable=True)
+
+        pending_invite.party.add_player(recipient)
+
+        self.remove_disbanded_parties()
```
Askaholic: Why do disbanded parties have to be checked here? Shouldn't they be checked here:

```python
if pending_invite.party not in self.player_parties:
    self.remove_disbanded_parties()
    raise ClientError("The party you're trying to join doesn't exist anymore.", recoverable=True)
```

server/team_matchmaking_service.py
```diff
@@ -0,0 +1,115 @@
...
+
+    def kick_player_from_party(self, owner: Player, kicked_player: Player):
+        if owner not in self.player_parties:
+            raise ClientError("You're not in a party.", recoverable=True)
+
+        party = self.player_parties.get(owner)
```
Askaholic: Can use `self.player_parties[owner]` here. When I see `.get()` that usually implies to me that we are expecting the possibility that the key does not exist, however, that isn't the case here since the code will already have errored out if the key doesn't exist.

server/team_matchmaking_service.py
```diff
@@ -0,0 +1,115 @@
...
+        if kicked_player not in party.members:
+            raise ClientError("The kicked player is not in your party.", recoverable=True)
+
+        party.remove_player(kicked_player)
+
+    def leave_party(self, player: Player):
+        if player not in self.player_parties:
+            raise ClientError("You are not in a party.", recoverable=True)
+
+        self.player_parties.get(player).remove_player(player)
```
Askaholic: Same as above

server/team_matchmaking_service.py
```diff
@@ -0,0 +1,115 @@
...
+        self.remove_disbanded_parties()
+
+    def clear_invites(self):
+        invites = {invite for invite in self._pending_invites.values() if time.time() - invite.created_at >= PARTY_INVITE_TIMEOUT or invite.sender not in self.player_parties}
```
Askaholic: Put on multiple lines

server/team_matchmaking_service.py
```diff
@@ -0,0 +1,115 @@
...
+
+        for invite in invites:
+            self._pending_invites.pop((invite.sender, invite.recipient))
+
+    def remove_party(self, party):
+        # party is removed from player_parties dict by disband() removing the key value pair for each player
+
+        party_invites = {invite for invite in self._pending_invites.values() if invite.party == party}
```
Askaholic: This probably wont make much of a difference since party sizes are going to be small (definitely less that 100 elements), but you're using the wrong data structure here. You should make it a list, not a set since you are iterating over it's elements. 

Here are some benchmarks (hopefully they are credible) for set vs list iteration performance. https://stackoverflow.com/questions/2831212/python-sets-vs-lists

server/team_matchmaking_service.py
```diff
@@ -0,0 +1,115 @@
...
+        for invite in party_invites:
+            self._pending_invites.pop((invite.sender, invite.recipient))
+
+        party.disband()
+
+    def remove_disbanded_parties(self):
+        disbanded_parties = {party for party in self.player_parties.values() if party.is_disbanded()}
```
Askaholic: Same as above

server/team_matchmaking_service.py
Askaholic: (use list instead of set)

Askaholic:
OK. Finished reading through it. Over all it's very clean nice job!

One other thing that occurred to be because of something Downlord mentioned, was that we should maybe think about how the clients could localize our server messages. So everywhere we have something like:

```python
raise ClientError("The invited player doesn't exist", recoverable=True)
```

Maybe we want to be sending some localizable error code instead? Just a thought.

server/team_matchmaker/player_party.py
Askaholic: Maybe instead of accessing `player.lobby_connection` directly to send messages, we could add a `send_message` method for `Player`. That way you can still send messages directly to a player if you have a reference to the player object, but you don't need to know about the implementation of lobby_connection.

Geosearchef:
- if a player leaves a party, the player remaining can't join any other parties
- same for being invited when having invited somebody
- when you are alone in a party and accept an invite, the party needs to be disbanded
- also, when leaving a party, you still get a party update about the party you're not in

server/lobbyconnection.py
```diff
@@ -830,19 +832,32 @@ async def command_game_join(self, message):
...
-
-        if self._attempted_connectivity_test:
-            raise ClientError("Cannot host game. Please update your client to the newest version.")
+        party = self.party_service.get_party(self.player)
 
         if state == "stop":
             await self.ladder_service.cancel_search(self.player, queue_name)
             return
 
+        if party is not None:
```
Askaholic, line 847: Additionally I think we can exit early if the requesting player is not the owner of their party.

PR #424, opened 2019-03-29
Run mypy

Sheeo:


scripts/run_tests_with_coverage.sh
```diff
@@ -1,3 +1,4 @@
 #!/usr/bin/env bash
 set -e
+mypy server.py
```
norraxx: Run with coverage?
Why not sparate?

Sheeo: This PR was a draft to test the github draft PR feature -- sorry I didn't make this clear :)

I think there's little value currently in running mypy in CI at all, I just added it to see how close we were to no type errors :)

Issue #399, opened 2019-02-11
Add tests for PlayerService


Askaholic:
`test_player_service.py` is currently an empty^ file. PlayerService its self has 80% coverage, but none of it is actually being tested directly.

^ it contains some imports and a duplicated fixture

PR #425, opened 2019-04-23
Add some tests for PlayerService

Askaholic:
Related to #399, but note that these tests only cover `fetch_player_data` and are therefore not comprehensive. More work remains to be done.

* Combined a bunch of sql queries into a single statement using the expression language.

### DIFF
--- a/server/player_service.py
+++ b/server/player_service.py
@@ -6,10 +6,10 @@ import marisa_trie
 import server.db as db
 from server.decorators import with_logger
 from server.players import Player
-from sqlalchemy import and_, select
+from sqlalchemy import select
 
 from .db.models import (avatars, avatars_list, clan, clan_membership,
-                        global_rating, ladder1v1_rating)
+                        global_rating, ladder1v1_rating, login)
 
 
 @with_logger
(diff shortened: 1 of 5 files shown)


server/player_service.py
```diff
@@ -48,35 +51,46 @@ def clear_dirty(self):
...
+                clan.c.tag
+            ], use_labels=True).select_from(
+                login
+                .join(global_rating)
+                .join(ladder1v1_rating)
+                .outerjoin(clan_membership)
+                .outerjoin(clan)
+                .outerjoin(avatars)
+                .outerjoin(avatars_list)
+            ).where(login.c.id == player.id)
```
norraxx: I like to write queries that way:
`select_from(login).join(...).outerjoin(...)` etc..

server/player_service.py
```diff
@@ -48,35 +51,46 @@ def clear_dirty(self):
...
+                row[global_rating.c.deviation]
+            )
+            player.numGames = row[global_rating.c.numGames]
+
+            player.ladder_rating = (
+                row[ladder1v1_rating.c.mean],
+                row[ladder1v1_rating.c.deviation]
+            )
+
+            player.clan = row.get(clan.c.tag)
```
norraxx: Can't you just get values from result `row` just accessing them like: `row.tag`,
same way before

Issue #422, opened 2019-03-29
TMM Implement party system and team matchmaking


Geosearchef:

Sheeo:
Suggested implementation and additional sideproject while we're at it (RFC):

Currently players queue up at random intervals and match with decreasing game quality over time (I cannot recall the threshold but you will get theoretically worse balanced games as time goes on in the queue). This doesn't work well when you don't have many people queuing up at once.

Instead I suggest we make queues "pop" at global times, say every 10-15 minutes for each queue (1v1, 2v2, 3v3 etc.).

This way--ideally anyway--the set of players to match is bigger and the chance of getting a balanced match is higher, at the cost of having to schedule your playing start time at queue pop interval, which might even be considered a benefit since you know exactly when your game will launch if you get a match.

Exact pop times should be variable with online population.

Pop times should be made very visible in the client so it's clear when you should queue and how much time is left to make it in this round
Ideally they're staggered as well, so there's a shorter timeframe between a 1v1 pop and a 2v2 pop than the 15 minutes to next 1v1 pop

I think the matchmaking system wholly overlaps though
There's no reason the same queue/matchmaking system cannot be used for 1v1/ladder and NvN[vN...] (edited) 

Rather, it can be described with a generic class configured with:
- Map pool
- Team configuration
- Game launch options
- Rating scheme
- (Queue pop timings) (edited) 
So, with what you've added now there's the ability to create and destroy parties. The matchmaking queue then needs to be generalized with the above support, and to deal with matching `List[Union[PlayerParty,Player]]` rather than just matching `List[Player]`
(Or to simplify things you enforce party size == queue team size in order to be able to queue your party)

Geosearchef:
@ sheeo, that's a really good idea and we should be option for that, some other notes from various debates/private discussions:

- group up with friends (party)
- join queue
- matching at dynamic intervals
- 20-30 secs before matching, everybody in a party has to ready up
- all ready parties are considered for matching

Askaholic:
I was thinking about making the server send the client more info during matchmaking to let the client know when a match is found. I was thinking it would look something like this:
```
{
    "command": "match_found",
    "queue": "ladder1v1",
    "opponents": [<opponent_player_id>],
    "factions": ["cybran"]
}
```

That way the client could let you know who you get matched up against, and maybe even show you a bit of their profile (finally a reason for those profile images). You could see their rank, and games played etc and what faction they're playing.

[...]

PR #426, opened 2019-04-25
Rework ladder for team matchmaking

Askaholic:
Rework `LadderService` and `LobbyConnection` to support adding multiple queues. Currently `LadderService` will enforce players only being in one queue at a time by only allowing players with player state `IDLE` to queue. This prevents lots of shenanigans with trying to initiate multiple conflicting asynchronus actions as long as the initiation of each of those actions appropriately changes the player state.

Notification of new ladder matches is done by a "pull" from `LadderService` rather than a "push" from `MatchmakerQueue`. This means `MatchmakerQueue` no longer needs to know how matches are started, and we will be able to reuse the same `MatchmakerQueue` implementation with a different match starting function.

Link #422


server/ladder_service.py
```diff
@@ -19,14 +23,57 @@ class LadderService:
     Service responsible for managing the 1v1 ladder. Does matchmaking, updates statistics, and
     launches the games.
     """
-    def __init__(self, games_service: "GameService"):
+    def __init__(self, games_service: GameService):
         self._informed_players: Set[Player] = set()
         self.game_service = games_service
 
+        # Hardcoded here until it needs to be dynamic
```
Rackover: No
Should be made dynamic in the first place
In fact, I think Ladder1v1 and LadderXvX should be exactly the same and using the same code (1v1 is no special case)
Let's think the "2v2" as "XvX" right now so that we can easily expand it in the future

server/ladder_service.py
```diff
@@ -71,6 +136,60 @@ def inform_player(self, player: Player):
...
+        )
+
+        game.map_file_path = map_path
+
+        for i, player in enumerate(all_players):
+            i += 1  # Game options are 1-indexed
+            player.game = game
+
+            # Configure game options
+            # TODO: Start spots might not make sense
```
Rackover: Odd for team 1, Even for team 2

server/matchmaker/matchmaker_queue.py
```diff
@@ -19,16 +19,70 @@ def __init__(
         self.queue_name = queue_name
         self.rating_prop = 'ladder_rating'
         self.queue = OrderedDict()
+        self._matches = deque()
+        self._shutdown = False
         self._logger.debug("MatchmakerQueue initialized for %s", queue_name)
 
-    def push(self, search: Search):
+    async def iter_matches(self):
+        """ Asynchronusly yields matches as they become available """
```
Rackover: asynchronously

server/matchmaker/matchmaker_queue.py
```diff
@@ -19,16 +19,70 @@ def __init__(
...
+
+        while not self._shutdown:
+            if not self._matches:
+                await asyncio.sleep(1)
```
Rackover: while not shutdown, sleep ? 
I'm not sure of what you're trying to do but it doesn't look very clean

server/matchmaker/search.py
```diff
@@ -76,7 +79,8 @@ def boundary_80(self):
 
         These are the mean, rounded to nearest 10, +/- 200, assuming sigma <= 100
         """
-        mu, _ = self.rating
+        # TODO: Figure out what do do with these boundaries
```
Rackover: Ask @Exotic-Retard

server/ladder_service.py
Askaholic: Dynamic means "happens at runtime." So dynamically adding queues would mean something like adding a control server endpoint to configure the queues without needing to restart the server. This is (IMO), to quote yourself, "massively overkill" and something that is not necessary for team matchmaking MVP.

I agree that in theory Ladder1v1 feels like it should be exactly the same as XvX. However, ladder1v1 requires that the game mode be set to `ladder1v1` which instantiates the game object as a `LadderGame` instead of a `CustomGame`, that the rating be using `ladder_rating`, and currently the game configures both ladder players to be in the FFA(-) team, which obviously won't work for XvX.

Basically, it means in order to combine ladder1v1 with ladderXvX, we will have to change how ladder1v1 games behave, and I didn't want to do that yet since that isn't the point of this PR.

server/matchmaker/matchmaker_queue.py
```diff
@@ -19,16 +19,70 @@ def __init__(
...
+                # If the queue was cancelled, or some other error occured,
+                # make sure to clean up.
+                self.game_service.mark_dirty(self)
+                if search in self.queue:
+                    del self.queue[search]
+
+    def find_match(self, search: Search) -> bool:
+        self._logger.debug("Searching for matchup for %s", search.players)
+        for other in self.queue.copy().values():
+            if other == search:
```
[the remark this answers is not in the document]
Askaholic: Yes it works. This is old code again, it shows up as new lines in git because I refactored it to a function, but it is exactly the same code that is running on prod right now

server/matchmaker/matchmaker_queue.py
```diff
@@ -58,51 +116,9 @@ def to_dict(self):
         """
         return {
             'queue_name': self.queue_name,
-            'boundary_80s': [search.boundary_80 for player, search in self.queue.items()],
-            'boundary_75s': [search.boundary_75 for player, search in self.queue.items()]
+            'boundary_80s': [search.boundary_80 for search in self.queue.values()],
+            'boundary_75s': [search.boundary_75 for search in self.queue.values()]
```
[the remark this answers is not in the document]
Askaholic: The values are different. Notice the first is `search.boundary_80` and the second is `search.boundary_75`. AFAIK these are to let clients know when other people of similar skill are searching ladder. That's how the Java client sends those annoying 1v1 notifications, and how the py client makes the little banner appear for quick joining a game.

Not sure why there are to different boundaries though. Maybe the 75 one is for new players since they will have lower quality matches.

server/matchmaker/search.py
```diff
@@ -29,9 +29,12 @@ def __init__(
         :param rating_prop: 'ladder_rating' or 'global_rating'
         :return: the search object
         """
+        assert isinstance(players, list)
+        for player in players:
+            assert getattr(player, rating_prop) is not None
```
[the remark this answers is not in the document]
Askaholic: New players have rating (1500, 500)

server/matchmaker/search.py
```diff
@@ -105,19 +110,23 @@ def match_threshold(self):
...
+        thresholds = []
+        for rating in self.ratings:
+            _, deviation = rating
+
+            for d, q in self._deviation_quality.items():
+                if deviation >= d:
+                    thresholds.append(max(q - self.search_expansion, 0))
+        return min(thresholds)
+
+    def quality_with(self, other: 'Search'):
```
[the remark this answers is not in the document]
Askaholic: Sadly, not on python 3.6, only on 3.7+

server/matchmaker/search.py
```diff
@@ -137,7 +146,7 @@ def matches_with(self, other: 'Search'):
         if not isinstance(other, Search):
             return False
         elif self.quality_with(other) >= self.match_threshold and \
-            other.quality_with(self) >= other.match_threshold:
+                other.quality_with(self) >= other.match_threshold:
```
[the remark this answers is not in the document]
Askaholic: Link. My linter complains about that:
https://lintlyci.github.io/Flake8Rules/rules/E125.html

server/ladder_service.py
```diff
@@ -35,11 +82,29 @@ def inform_player(self, player: Player):
                 progress = (500.0 - deviation) / 2.5
                 player.lobby_connection.sendJSON(dict(command="notice", style="info", text="The system is still learning you. <b><br><br>The learning phase is " + str(progress)+"% complete<b>"))
 
-    async def start_game(self, host: Player, guest: Player):
+    def start_queue_handlers(self):
+        for queue in self.queues:
+            if queue == "ladder1v1":
+                asyncio.ensure_future(self.handle_queue_matches_1v1())
+            else:
+                asyncio.ensure_future(self.handle_queue_matches(queue))
```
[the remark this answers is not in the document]
Askaholic: Different game mode, different rating, different teams, different maps

server/ladder_service.py
```diff
@@ -71,6 +136,60 @@ def inform_player(self, player: Player):
...
+        for player in team1:
+            game.set_player_option(player.id, 'Team', 2)
+
+        for player in team2:
+            game.set_player_option(player.id, 'Team', 3)
+
+        mapname = re.match('maps/(.*).zip', map_path).groups()[0]
+
+        host.lobby_connection.launch_game(game, is_host=True, use_map=mapname)
+        await asyncio.sleep(4)
```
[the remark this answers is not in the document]
Askaholic: Need to wait for host to set up lobby before guest can join.

server/lobbyconnection.py
```diff
@@ -291,7 +293,7 @@ def send_updated_achievements(self, updated_achievements):
                                     ban_fail = row[0]
                                 else:
                                     if period not in ["DAY", "WEEK", "MONTH"]:
-                                        self._logger.warn('Tried to ban player with invalid period')
+                                        self._logger.warning('Tried to ban player with invalid period')
```
Rackover: Is this part of the code even used anymore ?

server/gameconnection.py
```diff
@@ -92,7 +90,7 @@ def send_message(self, message):
         elif state == PlayerState.JOINING:
             pass
         else:
-            self._logger.exception("Unknown PlayerState")
+            self._logger.exception("Unknown PlayerState: %s", state)
```
Rackover: So here the server expects JOINING or HOSTING... but receives Initializing instead. that's interesting.

Askaholic: I think this was a different thing because the `Initializing` thing is a `GameState` not a `PlayerState`

server/matchmaker/matchmaker_queue.py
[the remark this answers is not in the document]
Askaholic: `copy` always does a shallow copy. For deep copy you have to use `deepcopy`

PR #429, opened 2019-05-07
Reduce server log spam

Askaholic:


Askaholic:
Adding another log level doesn't solve the problem. At that point we might as well just set the log level on prod to `WARN`, which IMO would be OK, since as it stands the logging isn't accomplishing anything at all because no one will send me any logs :P

Brutus5000:
I cannot hand out logs due to the sheer size, and this seems the main driver behind it.

PR #430, opened 2019-05-11
[RFC] Implement utf8 protocol and dynamic protocol switching

### LINKED ISSUE #132: Use UTF-8 encoding for main protocol
Currently the protocol uses a `QDataStream` compatible format, which implies text is encoded as `UTF-16BE`.

Since most of our data is English text, this means we incur an unnecessary 50% space overhead, and a small amount of conversion, for every message. For slow clients (Both in terms of processing and link-speed) this can especially be a problem.

**Acceptance Criteria**
- Protocol uses plain `UTF-8` encoded JSON with no other binary elements.

**Tips to get started**
- The protocol encoding is currently performed entirely by [this](https://github.com/FAForever/server/blob/develop/server/protocol/qdatastreamprotocol.py#L14)
- Replace this with a different implementation that only uses `UTF-8` and doesn't bother with legacy encoded messages
- Make sure to update explicit uses of `QDataStreamProtocol`, such as [here](https://github.com/FAForever/server/blob/develop/server/__init__.py#L65)

Askaholic:
Fixes #132.

Makes some minor adjustments to allow different protocol encoding schemes to be implemented and add a basic JSON/UTF-8 implementation. Currently ping messages are still encoded as QStrings, but these messages are also completely ignored by the server. Can they be removed entirely?

Please give feedback:
* How does the client tell the server to switch to a different protocol?
* Can we safely make UTF-8 be the default protocol, now that py client is deprecated?
* What should happen with ping messages?

### DIFF
--- a/server/__init__.py
+++ b/server/__init__.py
@@ -18,7 +18,7 @@ from .stats.game_stats_service import GameStatsService
 from .gameconnection import GameConnection
 from .ice_servers.nts import TwilioNTS
 from .lobbyconnection import LobbyConnection
-from .protocol import QDataStreamProtocol
+from .protocol import QDataStreamProtocol, Protocol
 from .servercontext import ServerContext
 from .geoip_service import GeoIpService
 from .player_service import PlayerService
@@ -26,6 +26,7 @@ from .game_service import GameService
 from .ladder_service import LadderService
 from .control import init as run_control_server
 from .matchmaker import MatchmakerQueue
+from .types import Address
 
 __version__ = '0.9.17'
 __author__ = 'Chris Kitching, Dragonfire, Gael Honorez, Jeroen De Dauw, Crotalus, Michael Søndergaard, Michel Jung'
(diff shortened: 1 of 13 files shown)


Brutus5000:
I see no relevant use in switching the protocol to UTF-8 without other fundamental changes. The break of compatibility for older clients is not justified for the purposes of "a little less protocol overhead".

From the reworks being done in the java server, we (Downlord and I) see the future in switching over to Websocket-based communication:
- adds encryption to the protocol if desired (-> on production)
- good library support
- can be tested and simulated with simple web-tools
Websocket would imply UTF-8 of course.

The java server would have supported both legacy-tcp and websocket protocols. As they are listening on different ports, it would have transformed both channels to the same internal message format.

Switching the protocol on the same port adds additional complexity to the server, where it is not necessary.

Regarding the Ping message I'd like to forward the question to @micheljung as he has investigated this question already, but I don't know the outcome.

Askaholic:
There are some asyncio websocket libraries which look pretty straight forward (Note to self: https://websockets.readthedocs.io/en/stable/)

> it would have transformed both channels to the same internal message format.

So this implies that there are no major changes in the protocol regarding the types of messages that are being sent? Maybe just the names of those messages?

> Switching the protocol on the same port adds additional complexity to the server, where it is not necessary.

This would only add complexity related to different message names, or a completely different message structure, which would be the case regardless of the port that it's listening on. The server doesn't care what port messages are coming from, and what their underlying encoding is. Switching dynamically is not inherently hard (5 lines of code in this PR), it would probably be a bit awkward with websockets though which are implemented in a different library.

PR #434, opened 2019-05-19
Report system integration

Rackover:
The server now uses the new report system instead of the teamkill-only report one

### DIFF
--- a/server/gameconnection.py
+++ b/server/gameconnection.py
@@ -11,6 +11,7 @@ from .player_service import PlayerService
 from .players import Player, PlayerState
 from .protocol import GpgNetServerProtocol, QDataStreamProtocol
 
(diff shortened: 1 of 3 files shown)


server/gameconnection.py
```diff
@@ -305,11 +305,18 @@ def send_message(self, message):
         """
 
         async with db.engine.acquire() as conn:
+            result = await conn.execute(
+                """ INSERT INTO `moderation_report` (`reporter_id`, `game_id`, `game_incident_timecode`)
```
Brutus5000: Description with missing telling that it is an auto-generated report for a teamkill from clicking in the game.

Rackover: you're right!

PR #435, opened 2019-05-20
Fix wrong player avatar being selected

Geosearchef:
Currently all selected avatars are the first avatar of the player due to the query not filtering the avatars if it is selected.

### DIFF
--- a/server/player_service.py
+++ b/server/player_service.py
@@ -68,7 +68,7 @@ class PlayerService:
                 .outerjoin(clan)
                 .outerjoin(avatars)
                 .outerjoin(avatars_list)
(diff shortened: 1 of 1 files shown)


Geosearchef:
How do you configure an outer join using the db engine? If there is no avatar present or selected, a row should still be returned.

Askaholic:
Here are the docs. https://docs.sqlalchemy.org/en/13/core/tutorial.html

I don’t have time to work on it right now but I might be able to make some time in a few hours. Also if you’re going to add a fix, then write unit tests

Geosearchef:
I thought about that unit test aswell. Don't have the time now though.

PR #439, opened 2019-06-01
Debug Ladder: Add some additional logging statements

Askaholic:
This should help make more sense of the prod logs


Brutus5000:
We should also add logging here: https://github.com/FAForever/server/blame/ae6d056b8d636c4c69b41f83f9d680655e29193d/server/lobbyconnection.py#L153

In case we send these without the client showing them.

Issue #427, opened 2019-04-28
Ladder matching hugely different ranked players too easily


JonnoHR31:
The ladder system currently seems to prioritise speed of matching over matching with someone of a close rank, resulting in many horribly unbalanced games. 

Today I experienced - insta-matching with someone 5 times my rank (200 vs 1000). I quit immediately (pointless game), only to then insta match with a 600 rank. 

All this whilst over 800 players were online. If the system simply waited a little bit longer, the rank deviation could be significantly reduced.

Askaholic:
Thanks for the report! This would hopefully be alleviated by Sheeo's proposed changes in #422.

PR #448, opened 2019-06-10
Make search expansion grow through time instead of shrink

### LINKED ISSUE #445: Search expansion increases threshold with time instead of decreasing it
Search expansion is currently defined as:

```python
@property
def search_expansion(self):
    """
    Defines how much to expand the search range of game quality due to waiting time
    """
    return 0.25 * min(1 / ((time.time() - self.start_time) / 300), 1)
```
Which means that it will decrease over time.
And it is used like this:
```python
thresholds.append(max(q - self.search_expansion, 0))
return min(thresholds)
```
Which means threshold will increase over time.

Search expansion basically boils down to `y=1/x` where x is time, but it would make more sense for it to follow a `y=sqrt(x)` or some sort of logistic function instead.

Askaholic:
Closes #445 
Might help with #427

Your threshold for matches will now decrease (it was increasing before) the longer you wait. Meaning that the longer you sit in queue without finding a match, the more likely it will become that the server can find a match for you at the cost of that match probably being uneven.


Rackover:
Started searching
```
DEBUG    Jun 19  17:56:25 MatchmakerQueue                Game quality with [Player(test, 7982, (1594.34, 297.648), (1500.0, 500.0))]: 0.244337 (other threshold: 0.400000)
```

After a few minutes
```
DEBUG    Jun 19  17:59:08 MatchmakerQueue                Game quality with [Player(Giebmasse, 128626, (2441.09, 95.087), (20.2615, 10.047))]: 0.244337 (other threshold: 0.800000)
```

If threshold increases it means that the matchmaker is **less** tolerant about rating stretch, not more tolerant, no ? Are you sure your PR is doing what it should ?

Rackover:
Tested, works as intended, although it's very slow and I would like the parameters to be more configurable

PR #453, opened 2019-06-18
Issue/#399 Add tests for PlayerService

### LINKED ISSUE #399: Add tests for PlayerService

Askaholic:
Closes #399


Brutus5000:
Checking blacklisted emails is not used anywhere in the server and should be removed. It is part of the registration and email change that is now handled by API.

Askaholic:
Good point, I totally didn't even check for that. I'll remove it in this PR

tests/unit_tests/test_player_service.py
```diff
@@ -43,3 +43,60 @@
...
+    assert player_service[0] is None
+    assert player_service.get_player(0) is None
+
+
+def test_mark_dirty(player_service):
+    player = Mock()
+    player_service[0] = player
+
+    player_service.mark_dirty(player)
+    player_service.mark_dirty(player)
```
Brutus5000: Why twice?

Askaholic: To test uniqueness. Marking the same player as dirty multiple times should still result in the player showing up in the dirties list only once. Maybe I should add a comment

Askaholic: Maybe it's pointless because the data structure is a set anyways, but I don't think it hurts to have it in there twice. You never know, maybe you get a list of sets if you mark dirty twice. It is Python after all, crazy stuff is possible.

Brutus5000: In that case I would put 2 asserts: one after the first mark_dirty, and one after the second.

Askaholic: Yep, that's what I did

PR #456, opened 2019-06-22
Send a warning to players logging in with the legacy client

Askaholic:


server/lobbyconnection.py
```diff
@@ -385,15 +385,25 @@ def send_updated_achievements(self, updated_achievements):
...
         version = message.get('version')
         server.stats.gauge('user.agents.None', -1, delta=True)
         server.stats.gauge('user.agents.{}'.format(self.user_agent), 1, delta=True)
 
+        if not self.user_agent or 'downlords-faf-client' not in self.user_agent:
+            self.send_warning(
+                "You are using an unofficial client version! "
+                "Some features might not work as expected. "
+                "If you experience any problems please download the latest "
+                f"version of the official client from {config.WWW_URL}"
```
Geosearchef: What's that f doing there?

Askaholic: `f` makes it a format string. It’s python 3.6+ syntax.

https://cito.github.io/blog/f-strings/

PR #457, opened 2019-06-23
Ladder matchmaking at dynamic intervals

### LINKED ISSUE #427: Ladder matching hugely different ranked players too easily

Askaholic:
Hopefully fixes #427

These are cherry picked commits of the "queue pop" code that I wrote in #428. This would let us deploy the queue popping for regular ladder 1v1 even before team matchmaking is completely finished.

This PR does NOT contain the team creation code for XvX games.


.travis.yml
```diff
@@ -23,7 +23,7 @@ before_install:
   - python -m pip install pipenv
 
 install:
-  - pipenv install --dev
+  - pipenv install --ignore-pipfile --dev
```
Sheeo: As per [this issue](https://github.com/pypa/pipenv/issues/3150) I think you want to be using `pipenv sync` instead?

Askaholic: Ah, ok. All I want is for Travis to install exactly the versions from pipfile.lock. So it sounds like `sync` is the right thing

PR #461, opened 2019-06-25
Check ban when hosting and joining a game

Rackover:


server/lobbyconnection.py
```diff
@@ -15,7 +15,7 @@
 import semver
 import server
 import server.db as db
-from sqlalchemy import and_, func, text
+from sqlalchemy import and_, func, text, select
```
Askaholic: Sort imports please

server/lobbyconnection.py
```diff
@@ -362,14 +362,14 @@ def send_updated_achievements(self, updated_achievements):
 
         player_id, real_username, dbPassword, steamid, create_time, ban_reason, ban_expiry = (row[i] for i in range(7))
 
+        now = datetime.now()
+        if ban_reason is not None and now < ban_expiry:
```
Askaholic: Put this after password check otherwise you will be leaking ban information about all users!

server/lobbyconnection.py
```diff
@@ -655,7 +655,7 @@ def command_ask_session(self, message):
             raise KeyError('invalid action')
 
     @timed
```
Askaholic: Not sure if timed works on async functions. You may need to remove it

server/lobbyconnection.py
```diff
@@ -664,6 +664,8 @@ def command_game_join(self, message):
         if self._attempted_connectivity_test:
             raise ClientError("Cannot join game. Please update your client to the newest version.")
 
+        await self.abort_connection_if_banned();
```
Askaholic: IS THAT A SEMICOLON!? BACK FOUL CREATURE! ⚰️👹

server/lobbyconnection.py
```diff
@@ -921,3 +925,24 @@ def sendJSON(self, data_dictionary):
...
+
+            data = await result.fetchone()
+
+            if data is not None:
+                self._logger.debug('Aborting connection of banned user: %s, %s, %s',
+                                   self.player.id, self.player.login, self.session)
+                self.send_ban_message_and_abort(data[ban.c.expires_at], data[ban.c.reason])
+
+    def send_ban_message_and_abort(self, expires_at, reason):
+                now = datetime.now()
```
Askaholic: Fix indentation

tests/unit_tests/test_lobbyconnection.py
```diff
@@ -120,14 +120,14 @@ def policy_server(loop):
     loop.run_until_complete(runner.cleanup())
 
 
-def test_command_game_host_creates_game(lobbyconnection,
+async def test_command_game_host_creates_game(lobbyconnection,
                                         mock_games,
```
Askaholic: Fix indentation

tests/unit_tests/test_lobbyconnection.py
```diff
@@ -120,14 +120,14 @@ def policy_server(loop):
...
                                         test_game_info,
                                         players):
     lobbyconnection.player = players.hosting
     players.hosting.in_game = False
     lobbyconnection.protocol = mock.Mock()
-    lobbyconnection.command_game_host(test_game_info)
+    await lobbyconnection.command_game_host(test_game_info)
```
Askaholic: I would replace all of these with `lobbyconnection.on_message_received` to make sure the message dispatch actually works. I think putting `timed` on an async function will cause this problem, but the way the tests are written, they will not detect it

tests/unit_tests/test_lobbyconnection.py
```diff
@@ -159,22 +159,22 @@ def test_launch_game(lobbyconnection, game, create_player):
...
     args_list = lobbyconnection.launch_game.call_args_list
     assert len(args_list) == 1
     args, kwargs = args_list[0]
     assert isinstance(args[0], CustomGame)
 
 
-def test_command_game_join_calls_join_game(mocker,
+async def test_command_game_join_calls_join_game(mocker,
                                            lobbyconnection,
```
Askaholic: Fix indentation

server/lobbyconnection.py
```diff
@@ -971,3 +968,30 @@ def send(self, message):
         if self.player:
             self._logger.debug("Lost lobby connection removing player {}".format(self.player.id))
             self.player_service.remove_player(self.player)
+
+    async def abort_connection_if_banned(self):
+        async with self._db.acquire() as conn:
+            now = datetime.now()
+            result = await conn.execute(
+                select([ban.c.reason, ban.c.expires_at]).where(
+                    ban.c.player_id == self.player.id and ban.c.expires_at < now
```
norraxx: Here is mistake, use `and_`
Use filter instead of where

tests/integration_tests/test_server.py
```diff
@@ -206,3 +207,40 @@
...
+            ban.insert().values(
+                player_id=player_id,
+                author_id=player_id,
+                reason="Test live ban",
+                expires_at=func.date_add(
+                    func.now(), text("interval 1000 year")
+                ),
+                level='GLOBAL'
+            )
+        )
```
norraxx: For such things rather make fixtures, everything that configures you database, should be in fixtures.
Like: give_me_ban fixture will insert ban for you :-)

cleborys: As I understand it the ban here has to be given *after* logging in and not during setup, so that you can log in first?
So if it was a fixture you would need to return a user that's logged in and banned, doing both steps in the fixture?
If that's too much for a fixture to do, then maybe it's enough to move that block of code into a helper function (in terms of readability).

norraxx: Maybe i would prefer test like action login, action ban, test message, instead of just putting query into the db.

Also i would like to see the test, where you have one user with ban, another without, and you ban first, but login second one and ask for ban. You will reveal that bug with `and` operator that is above.

tests/unit_tests/test_lobbyconnection.py
```diff
@@ -195,7 +202,11 @@ def policy_server(event_loop):
     lobbyconnection.player = players.hosting
     test_game_info['uid'] = 42
 
-    lobbyconnection.command_game_join(test_game_info)
```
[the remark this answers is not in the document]
Askaholic: I don't understand. I just replaced a direct call to `command_game_join` with an equivalent call to `on_message_received` to make sure that the message dispatch is working. I had a problem with this a while ago where all the tests passed even though there was a bug in the message dispatch which was preventing a certain handler function from ever being called in the first place.

tests/integration_tests/test_server.py
Askaholic: I can put it in a function. But it needs to happen after the player has logged in, otherwise the ban will simply be checked on login which is not what we're trying to test here.

server/lobbyconnection.py
```diff
@@ -352,8 +352,8 @@ def send_updated_achievements(self, updated_achievements):
             "lobby_ban.expires_at as expires_at "
             "FROM login "
             "LEFT JOIN lobby_ban ON login.id = lobby_ban.idUser "
-            "WHERE LOWER(login)=%s "
-            "ORDER BY expires_at DESC", (login.lower(), ))
+            "WHERE login=%s "
+            "ORDER BY expires_at DESC", login)
```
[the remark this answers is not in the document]
Askaholic: Maybe. `lobby_ban` is weird though because it's not a real table, it's just a view I think.

server/lobbyconnection.py
Askaholic: Good catch

tests/unit_tests/test_lobbyconnection.py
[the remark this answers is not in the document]
cleborys: I just thought that one would expect the direct call in the test, if the test is named `test_command_game_join_somrthing...`, but I am not aware of the all circumstances and if you feel that test_command_game_join_calls_join_game` is still a valid name for this test then please ignore me :)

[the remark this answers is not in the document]
Askaholic: Ok I see. It's maybe 1 step removed from DIRECTLY testing the function so I guess technically it's more of an integration test. But I don't see any downside to this. It's simply getting us more value out of these test, so unless you can convince me that there's a good reason not to do it like this, I'm just going to ignore you :P.

server/lobbyconnection.py
```diff
@@ -14,12 +14,13 @@
...
 from server.db import FAFDatabase
-from sqlalchemy import and_, func, text
+from sqlalchemy import and_, func, select, text
 
 from . import config
 from .abc.base_game import GameConnectionState
 from .config import FAF_POLICY_SERVER_BASE_URL, TRACE, TWILIO_TTL
-from .db.models import ban, friends_and_foes
+from .db.models import ban, friends_and_foes, lobby_ban
+from .db.models import login as t_login
```
norraxx: Let's rename `t_login` to `table_login`? it's not something what you would like to read.

tests/data/test-data.sql
```diff
@@ -5,7 +5,10 @@ insert into login (id, login, email, password, create_time) values
     (100, 'ladder1', 'ladder1@example.com', SHA2('ladder1', 256), '2000-01-01 00:00:00'),
     (101, 'ladder2', 'ladder2@example.com', SHA2('ladder2', 256), '2000-01-01 00:00:00'),
     (102, 'ladder_ban', 'ladder_ban@example.com', SHA2('ladder_ban', 256), '2000-01-01 00:00:00'),
-    (200, 'banme', 'banme@example.com', SHA2('banme', 256), '2000-01-01 00:00:00')
+    (200, 'banme', 'banme@example.com', SHA2('banme', 256), '2000-01-01 00:00:00'),
+    (201, 'ban_revoked', 'ban_revoked@example.com', SHA2('ban_revoked', 256), '2000-01-01 00:00:00'),
+    (202, 'ban_expired', 'ban_expired@example.com', SHA2('ban_expired', 256), '2000-01-01 00:00:00'),
+    (203, 'ban_long_time', 'ban_null_expiration@example.com', SHA2('ban_long_time', 256), '2000-01-01 00:00:00')
```
norraxx: you can leave `,` at the end of row :-)

Askaholic: In Python yes, but in SQL, sadly no you can't. Believe me I already tried. SQL is very picky about its syntax

server/db/models.py
```diff
@@ -145,6 +147,15 @@
     Column('is_active',     Boolean,    nullable=False)
 )
 
+# This is actually a view into the `ban` table with proper handling of ban
+# expiration and revocation
+lobby_ban = Table(
+    'lobby_ban', metadata,
+    Column('idUser',        Integer,    ForeignKey('login.id'), primary_key=True),
+    Column('reason',        Text,       nullable=False),
+    Column('expires_at',    DateTime)
```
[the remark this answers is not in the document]
Askaholic: Yea I want to try out black. Not going to change it for the models though. I think this is easier to read so I'm sticking to it.

server/lobbyconnection.py
Askaholic: I think it's clear from the context. When you see `select([t_login.c.login]...` you know what's going on. I would have left it as `login` but sadly that is already used to mean `username` in several places. I used `t_login` because that's what sqalcodegen does when it auto generates tables for you.

PR #463, opened 2019-07-13
Increase metrics collection

Askaholic:
- Collect number of failed logins as well as succeeded
- Tag hosted games with the game mode

### DIFF
--- a/server/fake_statsd.py
+++ b/server/fake_statsd.py
@@ -17,7 +17,7 @@ class DummyConnection:
     def gauge(self, a, b, delta=False):
         pass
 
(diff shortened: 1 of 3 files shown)


Askaholic:
I'm not attached to StatsD. If you know of an alternative (and have the time and all the usual reasons why stuff sits around and never gets changed) we could totally switch it out for something better. 

But I think having metrics is really valuable for finding out how code changes affect users, and StatsD is currently set up for it, so I'm using it to the best of it's abilities!

Proto violations and ICE messages are a great idea!

PR #466, opened 2019-07-14
Only tell authenticated clients about matchmaker info

### LINKED ISSUE #465: Server reports matchmaker_info to unauthenticated connections
The server sends matchmaker info to any active server connection even if that connection has not successfully completed the `command_hello` sequence. All other broadcasts are correctly sent only to authenticated clients.

Askaholic:
Closes #465

### DIFF
--- a/tests/integration_tests/test_matchmaker.py
+++ b/tests/integration_tests/test_matchmaker.py
@@ -1,7 +1,5 @@
 import asyncio
 
-from server import config
-
 from .conftest import connect_and_sign_in, read_until_command
 
 # Need to save the old sleep here otherwise the mocker recursively patches it
@@ -62,7 +60,7 @@ async def test_game_matchmaking(loop, lobby_server, mocker):
 
 async def test_matchmaker_info_message(lobby_server, mocker):
     mocker.patch('server.matchmaker.matchmaker_queue.time', return_value=1_562_000_000)
-    mocker.patch('server.matchmaker.matchmaker_queue.config.QUEUE_POP_TIME_MAX', return_value=1)
+    mocker.patch('server.matchmaker.matchmaker_queue.config.QUEUE_POP_TIME_MAX', 1)
 
     _, _, proto = await connect_and_sign_in(
         ('ladder1', 'ladder1'),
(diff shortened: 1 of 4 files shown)


tests/integration_tests/test_matchmaker.py
```diff
@@ -62,7 +60,7 @@
 
 async def test_matchmaker_info_message(lobby_server, mocker):
     mocker.patch('server.matchmaker.matchmaker_queue.time', return_value=1_562_000_000)
-    mocker.patch('server.matchmaker.matchmaker_queue.config.QUEUE_POP_TIME_MAX', return_value=1)
+    mocker.patch('server.matchmaker.matchmaker_queue.config.QUEUE_POP_TIME_MAX', 1)
```
Rackover: 🤔 
Why remove the additional precision for the unskilled reader?

tests/integration_tests/test_server.py
```diff
@@ -104,6 +106,29 @@
...
 
 
+@pytest.mark.slow
+async def test_info_broadcast_authenticated(loop, lobby_server):
+    proto1 = await connect_client(lobby_server)
+    proto2 = await connect_client(lobby_server)
+    proto3 = await connect_client(lobby_server)
+
+    await perform_login(proto1, ('test', 'test_password'))
+    await perform_login(proto2, ('Rhiza', 'puff_the_magic_dragon'))
```
Rackover: awesome password

tests/integration_tests/test_matchmaker.py
Askaholic: This actually changes the functionality. If you specify `return_value` then it will patch the object with an instance of `Mock` that when called as a function, returns that value. However if you supply it as a positional arg, then it actually replaces the object you are patching, with that new object. So...

Before (with `return_value=1`):
```python
QUEUE_POP_TIME_MAX = Mock(return_value=1)
```
After (with `1` as positional arg)
```python
QUEUE_POP_TIME_MAX = 1
```

The problem was that another part of the code was trying to compare `QUEUE_POP_TIME_MAX` and it was throwing some error like 'Mock object doesn't support < comparison with float'

Rackover:
Mergeable as soon as tested

PR #468, opened 2019-07-14
General code cleanup.

eforgacs:
Fixing whitespace/indentation/line length issues.
Ensuring compliance with PEP 8 where possible/practicable.

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -12,12 +12,12 @@ services:
 before_install:
   - sudo /etc/init.d/mysql stop
   - git clone https://github.com/FAForever/faf-stack.git faf-stack
-      && pushd faf-stack
-      && git checkout 8e9e3bd
-      && cp -r config.template config
-      && ./scripts/init-db.sh
-      && wget https://raw.githubusercontent.com/FAForever/db/$(grep -oP 'faforever/faf-db-migrations:\K(.*)$' ./docker-compose.yml)/test-data.sql
-      && popd
+    && pushd faf-stack
+    && git checkout 8e9e3bd
+    && cp -r config.template config
+    && ./scripts/init-db.sh
+    && wget https://raw.githubusercontent.com/FAForever/db/$(grep -oP 'faforever/faf-db-migrations:\K(.*)$' ./docker-compose.yml)/test-data.sql
+    && popd
   - docker exec -i faf-db mysql -uroot -pbanana faf < faf-stack/test-data.sql
   - docker exec -i faf-db mysql -uroot -pbanana faf -e "select * from login;"
   - python -m pip install pipenv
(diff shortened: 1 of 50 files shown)


Askaholic:
I think we would be best off using an auto formatter and committing a `setup.cfg` file. I've been using  [yapf](https://github.com/google/yapf) because it gives you quite a few customization options. 

For instance I really dislike this sort of thing:
```python
def some_function_with_a_long_name(self,
                                   param1,
                                   param2,
                                   param3):
```

Or this:
```python
self.some_function_with_a_long_name("Arg1 is a long string that takes up too much space", 
                                    arg2, arg3)
```

Because a) I think it looks awkward and b) it is annoying to change the formatting when you inevitably rename your function to something better.

So I would prefer (as you might have noticed from the code):
```python
def some_function_with_a_long_name(
    self, param1, param2, param3
):
self.some_function_with_a_long_name(
    "Arg1 is a long string that takes up too much space", 
    arg2, arg3
)
```
Where the closing paren aligns on the same level of indentation as the start of the expression that it's closing.

Anyways, all that is to say I think an auto formatter with a version controlled config file is the best option if we're going to commit to reformatting the entire code base. That way we can ensure consistency, and don't need to rely on humans reading, interpreting, and following a style guide document correctly. And If we use yapf then I already have a `setup.cfg` that I'm using for my work project with config for yapf and isort ;)

eforgacs:
Fair enough. I just thought I'd run it past you, to see if you preferred using this style (and also since my IDE was giving me messages about it). I'm just starting to contribute to the project and I figured I'd ramp up my commits in terms of content, starting with the most minor edits and moving upwards gradually.

I do think using the auto-formatter with the version controlled config file also makes a lot of sense, so if that's the style you prefer, then go for it. I think as long as it's consistent one way or another, it helps with readability, and thereby helps people find issues faster, etc. And of course, even Python's [PEP 8 style guide mentions that there are good reasons to ignore it sometimes](https://www.python.org/dev/peps/pep-0008/#a-foolish-consistency-is-the-hobgoblin-of-little-minds)! So I think common sense has to prevail.

There were even certain files that I couldn't bring myself to commit, even in this pull request. For example, the auto-formatter in my IDE, PyCharm, would have had me change server/db/models.py so that each of those columns you had nicely lined up would have the whitespace removed, which seemed worse than the way you had it, so I didn't commit that one. Or, in the case of the JSON files (e.g., tests/data/game_stats_single_player.json), it would have changed the concise JSON you had into spread-out, pretty-printed JSON, which is spacier but also less concise to read.

Anyway, I'll close this one out, but I did find some other issues I wanted to bring up, like making certain methods static, adding required packages to setup.py, and other small changes (I'll make a separate pull request for each of those). Thanks again for letting me contribute, and looking forward to making more pull requests soon!

Askaholic:
Always appreciate an extra set of eyes to look over the code! BTW we just started using ZulipChat here at FAF (migrated over from slack about a week ago), I'll get one of the admins to send you an invite using the email on your Github profile

PR #471, opened 2019-07-18
Listing required packages in setup.py.

eforgacs:
Adding required packages containing modules to setup.py.

### DIFF
--- a/setup.py
+++ b/setup.py
@@ -13,5 +13,7 @@ setup(
     author=server.__author__,
     author_email=server.__contact__,
     description='Lobby/game server project',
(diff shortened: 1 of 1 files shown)


Askaholic:
Any way of reading them in from the `Pipfile`? I think the Pipfile is basically some sort of TOML file.

I'm honestly not quite sure why the project even has a setup.py since it is only used for installing its self into the docker container. Not sure if there are any advantages to that vs just running from the source directory.

Askaholic:
I think this is bugged since it says `unknown repository`

PR #473, opened 2019-07-20
Feature/#440 Accept authentication via OAuth token

### LINKED ISSUE #440: Accept OAuth token instead of password hash
There should be only one system responsible for authentication and that is the API.
Instead of sending the hashed password to the server, the client should fetch a token from the API (which it does anyway) and send it to the server.
This would also improve client security, as a token can be stored rather than a hashed password.

Best case would be to do it in 2 steps: Add an additional command for Oauth login with parallel usage, then disable direct login.

Askaholic:
#440 (Can close once the client implementation is there and `command_hello` has been removed)

Adds a new command `auth` which allows users to log in with an OAuth token. Let me know if I should rework anything regarding the protocol.

Also introduces a primitive module system which we can use to refactor LobbyConnection.


Brutus5000:
In the API we would use ssh key format (see application.yml here https://github.com/FAForever/faf-java-api/pull/323/files#diff-f90a80a411765ea693b5d3f1125e1a35). That doesn't match the header defined here. Ping me if that is an issue. That was the easiest solution on java side.

Regarding IRC I'd suggest the server generates a new password on login and sends it in a separate response (or as extension of the hello message).

If possible we should extend the existing protocol without breaking compatibility. Otherwise we need to couple deployment of server and client(s) at the same time - which probably doesn't work out very well.

server/auth.py
```diff
@@ -0,0 +1,52 @@
...
+        try:
+            token = jwt.decode(token, mod.pub_key, algorithms="RS256")
+            user_id = token["user_id"]
+        except (InvalidTokenError, KeyError):
+            raise AuthenticationError("Token signature was invalid")
+
+        new_irc_password = b64encode(os.urandom(30)).decode()
+        conn.send({
+            "command": "irc_password",
+            "password": new_irc_password
```
axel1200: Correct me if I am wrong, but I think the way it works is. The first time somebody logs into IRC the username is associated with a password in our case this is 
grafik
Something pretty strange. And IRC will only let u log in if the second time it is still the same password. This way you can reserve your username. Now u send a new IRC password everytime the user logs in, so this won't work(if I got it right). So we either generate a random password per user and it stays the same or maybe the client saves it the first time he logs in to this has issues when he changes clients. Or like we do now we need to derive it from the password, which is bad because the sever is not supposed to know that. Probably storing a random password in db and handing that out on login would be best. @Brutus5000 correct me if I am wrong and you as well @Askaholic

axel1200: https://github.com/FAForever/docker-unrealircd
Yes i think i am right

axel1200: So way to go if u ask me is make a new column and store a random string there per user

axel1200: Also I think we might just do that in the java-api. We can make a database migration and add the column and fill it with a random string then just add the field in the java-api user entity... Then the client can ask the api for the irc password

Askaholic: The server updates each user's IRC password every time they log in. I guess this is in case the user changes their password. It would probably make the most sense for the API to handle this entirely, and only update the password when the user actually changes their password.

Brutus5000:
@Askaholic The corresponding changes in the API are now rolled out on test and production. The public key is in faf-stack under `config/faf-java-api/pki/public.key`

axel1200:
- tokens need to be removed from log
- on the test server the message appears:
```
 Failure updating NickServ password for axel12
```

axel1200:
Apart from that seems to work, my implmentation seems to struggel a little lot of small bugs because things changed

axel1200:
@Askaholic if you remove the tokens from log it should be good to go. Can we look into it again.

Askaholic:
Superseded by #591

PR #479, opened 2019-07-24
Issue/#166 Remove whitespace from json messages

### LINKED ISSUE #166: Server bothers with whitespace in json messages
Messages from the server come out like this:
`{"session": 1842478479, "command": "session"}`

We probably should be sending minified JSON. Would very slightly help performance in the presence of large messages (like the huge game info blob you get on login). Of course, UTF-8 would have a vastly larger effect, but is slightly trickier.

Askaholic:
Closes #166 

This will slightly decrease the network overhead of the protocol


Askaholic:
Yea, debug print doesn't use the `json` module. It just uses the python builtin `__str__` function on dicts.
https://github.com/FAForever/server/blob/b8910bd5dc2cc05afe775582318f5809637f0829/server/lobbyconnection.py#L906

PR #481, opened 2019-07-27
Smarter queue pop times

Askaholic:
The previous queue pop time calculations were pretty arbitrary. This method is based on trying to achieve a certain number of players in the queue for matching,  based on the observed rate of players queuing for matchmaker. In other words, the queue pop timer tries to adjust the pop times so that there are always a constant number of people in the queue.

### DIFF
--- a/server/matchmaker/matchmaker_queue.py
+++ b/server/matchmaker/matchmaker_queue.py
@@ -3,7 +3,6 @@ from collections import OrderedDict, deque
 from concurrent.futures import CancelledError
 from datetime import datetime, timezone
 from statistics import mean
-from time import time
 from typing import Deque, Dict
 
 import server
@@ -11,6 +10,7 @@ import server
 from .. import config
 from ..decorators import with_logger
 from .algorithm import stable_marriage
+from .pop_timer import PopTimer
 from .search import Match, Search
 
 
@@ -26,11 +26,9 @@ class MatchmakerQueue:
 
         self.queue: Dict[Search, Search] = OrderedDict()
         self._matches: Deque[Match] = deque()
-        self.last_queue_amounts: Deque[int] = deque()
-        self.last_queue_times: Deque[float] = deque()
         self._is_running = True
-        self._last_queue_pop = time()
-        self.next_queue_pop = self._last_queue_pop + config.QUEUE_POP_TIME_MAX
+
+        self.timer = PopTimer(self.queue_name)
         asyncio.ensure_future(self.queue_pop_timer())
         self._logger.debug("MatchmakerQueue initialized for %s", queue_name)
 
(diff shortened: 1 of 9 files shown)


server/matchmaker/pop_timer.py
```diff
@@ -0,0 +1,68 @@
...
+from typing import Callable, Deque
+
+import server
+
+from .. import config
+from ..decorators import with_logger
+
+
+@with_logger
+class PopTimer(object):
```
norraxx: It would be cool to have docstring, to not to read what that class is trying to solve, but just read it, why it's here.

server/matchmaker/matchmaker_queue.py
```diff
@@ -51,14 +50,7 @@ def __init__(
...
         while self._is_running:
-            time_remaining = self.next_queue_pop - time()
-            self._logger.info("Next %s wave happening in %is", self.queue_name, time_remaining)
-            server.stats.timing("matchmaker.queue.pop", int(time_remaining), tags={'queue_name': self.queue_name})
-            await asyncio.sleep(time_remaining)
-            server.stats.gauge(f"matchmaker.queue.{self.queue_name}.players", len(self))
-
-            self._last_queue_pop = time()
-            self.next_queue_pop = self._last_queue_pop + self.time_until_next_pop()
+            await self.timer.next_pop(lambda: len(self))
```
norraxx: `len(self)`? O_o which magic was used here?
let's avoid magic. ;-)

norraxx: `len(self.queue)` is much better. Explicit is better than implicit

norraxx: and it would be cool if we get rid of that `def __len__` method ;-)

server/matchmaker/pop_timer.py
```diff
@@ -0,0 +1,68 @@
...
+        self.next_queue_pop = self._last_queue_pop + self.time_until_next_pop(
+            num_players, time_remaining
+        )
+
+    def time_until_next_pop(self, num_queued: int, time_queued: float) -> float:
+        """ Calculate how long we should wait for the next queue to pop based
+        on the current rate of ladder queues
+        """
+        self.last_queue_amounts.append(num_queued)
+        self.last_queue_times.append(time_queued)
```
norraxx: do you clear that queues? or they are growing to infinity?

Askaholic: python `deque`'s have a builtin feature for max length which I'm setting in `__init__`:

https://docs.python.org/3/library/collections.html#collections.deque

norraxx: And what for is that deque?

If dequeue always has 10 elements, where are you working with that elements?

norraxx: If poptimer is speeding up, by number of people in queue, what happens, when there is not much people on queue? Will poptimer slow down?

Askaholic: The deques are used to calculate the moving average of players in the queue. This is basically to smooth out the pop time adjustments since there is a small player base. You can see on the next line it says:
```python
total_players = sum(self.last_queue_amounts)
```

If there are few players in the queue the time will increase up to `config.QUEUE_POP_TIME_MAX`

norraxx: Ok.

PR #484, opened 2019-07-31
Misc cleanup

Askaholic:
Various little things:
- Updated readme
- Added pipenv script for running server locally in development mode
- Removed some unused code
- Added auto formatter config and ran auto formatter on some files

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -13,7 +13,7 @@ before_install:
   - sudo /etc/init.d/mysql stop
   - git clone https://github.com/FAForever/faf-stack.git faf-stack
       && pushd faf-stack
-      && git checkout 8e9e3bd
+      && git checkout v19.8.12
       && cp -r config.template config
       && ./scripts/init-db.sh
       && wget https://raw.githubusercontent.com/FAForever/db/$(grep -oP 'faforever/faf-db-migrations:\K(.*)$' ./docker-compose.yml)/test-data.sql
--- a/server/players.py
+++ b/server/players.py
@@ -125,12 +125,16 @@ class Player(BasePlayer):
 
         return dict(
             filter(
-                filter_none,
-                (('id', self.id), ('login', self.login),
-                 ('global_rating', self.global_rating),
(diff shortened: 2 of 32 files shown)


server/players.py
```diff
@@ -118,31 +113,31 @@ def game_connection(self, value):
...
+        return dict(
+            filter(
+                filter_none,
+                (('id', self.id), ('login', self.login),
+                 ('global_rating', self.global_rating),
+                 ('ladder_rating', self.ladder_rating),
+                 ('number_of_games', self.numGames), ('avatar', self.avatar),
+                 ('country', self.country), ('clan', self.clan))
+            )
+        )
```
norraxx: Can we rewrite this code it to dict comprehension?

norraxx: This construction is bad and unreadable.
Also sometimes some tuples are 2x in row, some of them just 1x. it's confusing

setup.cfg
```diff
@@ -0,0 +1,10 @@
+[isort]
+multi_line_output = 5
+
+[yapf]
+based_on_style = pep8
+spaces_before_comment = 4
+split_before_dot = true
+coalesce_brackets = true
+dedent_closing_brackets = true
+spaces_around_default_or_named_assign = false
```
norraxx: ah, not it makes sense why that stuff before happend.

.travis.yml
```diff
@@ -13,7 +13,7 @@ before_install:
   - sudo /etc/init.d/mysql stop
   - git clone https://github.com/FAForever/faf-stack.git faf-stack
       && pushd faf-stack
-      && git checkout 8e9e3bd
+      && git checkout v19.8.12
```
norraxx: It would be cool, if there will be something like `latest` tag.

norraxx: But it's not for this PR

Askaholic: Hm. I'm not sure if that's a good idea, because then it would be possible for changes in faf-stack to break the server travis build. I prefer to do it manually to make sure that builds don't randomly break

PR #486, opened 2019-08-21
Support for multiple player ratings

Wesmania:
This PR will eventually add support for more player ratings in preparation for team matchmaking. For now though, it's a few misc cleanups.


server/players.py
```diff
@@ -35,18 +35,21 @@ def __init__(
...
-        if global_rating is None:
-            global_rating = (1500, 500)
-        if ladder_rating is None:
-            ladder_rating = (1500, 500)
-        self.global_rating = global_rating
-        self.ladder_rating = ladder_rating
+        if global_rating is not None:
+            self.global_rating = global_rating
+        if ladder_rating is not None:
+            self.ladder_rating = ladder_rating
```
norraxx: Bad naming for `global_rating`, `self.global_rating` and `self._global_rating`, how about to create and use `set_global_rating()`?

server/players.py
```diff
@@ -76,6 +79,39 @@ def __init__(
         self._game = lambda: None
         self._game_connection = lambda: None
 
+    @property
```
norraxx: Don't use `@property`, it doesn't give you anything, only hiding call functions + add wrappers, without proper IDE impossible to detect with naked eye.
Use property only if you're trying to not to break same API during refactoring.
Better to use `set_` methods. It's more explicit.

server/players.py
```diff
@@ -35,18 +35,21 @@ def __init__(
         permission_group: int = 0,
         lobby_connection: "LobbyConnection" = None
     ):
-        super().__init__(player_id, login)
+        self._faction = 0
```
norraxx: typing for local variables + typing for incoming variables like `clan`, `global_rating`, `ladder_rating`.

server/players.py
```diff
@@ -76,6 +79,39 @@ def __init__(
         self._game = lambda: None
         self._game_connection = lambda: None
 
+    @property
+    def global_rating(self):
+        return self._global_rating
+
+    @global_rating.setter
+    def global_rating(self, value: Rating):
```
norraxx: wrong typing, use `Union[Tuple[int, int], Rating]` https://docs.python.org/3/library/typing.html#typing.Union

server/players.py
```diff
@@ -76,6 +79,39 @@ def __init__(
...
+        if isinstance(value, Rating):
+            self._global_rating = (value.mu, value.sigma)
+        else:
+            self._global_rating = value
+
+    @property
+    def ladder_rating(self):
```
norraxx: missing return typing, if it would be just variable, not property, code would be much easier to read, and for outside world it behaves the same

tests/conftest.py
```diff
@@ -179,28 +179,25 @@ def make_game(uid, players):
...
-def create_player():
+def player_factory():
     from server.players import Player, PlayerState
 
-    def make(login='', id=0, port=6112, state=PlayerState.HOSTING, ip='127.0.0.1', global_rating=Rating(1500, 250), ladder_rating=Rating(1500, 250)):
-        p = mock.create_autospec(spec=Player(login))
-        p.global_rating = global_rating
-        p.ladder_rating = ladder_rating
-        p.ip = ip
+    # FIXME
```
norraxx: Describe what's wrong?

server/players.py
```diff
@@ -76,6 +79,39 @@ def __init__(
...
+    @property
+    def global_rating(self):
+        return self._global_rating
+
+    @global_rating.setter
+    def global_rating(self, value: Rating):
+        if isinstance(value, Rating):
+            self._global_rating = (value.mu, value.sigma)
+        else:
+            self._global_rating = value
```
norraxx: if `value` is `None`, you may achieve bad behavior.

server/players.py
norraxx: I wouldn't use fraction `0`, it doesn't looks like it's undefined. Maybe use `None`?

server/players.py
Wesmania: This commit does not change existent naming, and the `name` property backed by a `_name` attribute is pretty common practice. The property already acts like a setter and has no side effects beyond setting the backing value anyway.

server/players.py
Wesmania: It has no side effects beyond setting the backing value. Sounds like a perfectly fine usage of @property to me.

server/players.py
Wesmania: I had no reason to touch this code beyond copying it. Faction could use an enum anyway rather than just an obvious typing hint.

server/players.py
Wesmania: This is copied from BasePlayer, so I didn't bother touching it. Frankly you know my opinion on type hints, in trivial cases like this one they bring exactly zero benefit.

server/players.py
Wesmania: As above, didn't bother changing copied code. It deserves to be a named tuple anyway, we can change it then.

tests/conftest.py
Wesmania: Missed this one. Removed.

server/players.py
Wesmania: This is the original behaviour and it was not my intention to change it. It's not set to `None` internally and if there's an inconsistency with how other code uses it, then you can show and fix it.

server/players.py
Wesmania: Commit has no intention to change original behaviour.

server/players.py
Askaholic: Maybe it would be a good opportunity to add `assert value is not None`

server/players.py
norraxx: In fact it has. If you see below, you'll find condition that decides what to do with Rating instance. It is hiding behavior.

server/players.py
norraxx: I don't see benefit from moving code from one place to another, if we can make it better.

server/players.py
norraxx: You can make it better ;-) you're refactoring, isn't it?

server/players.py
Askaholic: @norraxx Because it's a lot of extra work and a lot higher chance that you mess something up on accident. I'd be happy if you wanted to make a PR though :)

server/players.py
Wesmania: I was talking about *side effects*. The only thing setting this value ever affects is the value returned by the getter. This is a valid use case for a property, what you're describing restricts `@property` to trivial cases.

server/players.py
Wesmania: Whenever possible, one commit is one feature.

server/games/custom_game.py
```diff
@@ -1,6 +1,8 @@
 import time
 
 from .game import Game, ValidityState
+from server.rating import RatingType
```
Askaholic: For now this makes sense, but it won't anymore once you finish multiple rating support :P

server/games/game.py
```diff
@@ -242,6 +247,16 @@ def connections(self):
 
     @property
     def teams(self):
+        """
+        A dictionary of lists representing teams
```
Askaholic: It returns a set not a dict. It would look like `{2, 3}` assuming that there are 2 teams

server/matchmaker/search.py
```diff
@@ -20,22 +21,22 @@ def __init__(
         self,
         players: List[Player],
         start_time: Optional[float]=None,
-        rating_prop: str='ladder_rating'
+        rating_type: str=RatingType.LADDER_1V1
```
Askaholic: So the type isn't `str` anymore

server/matchmaker/search.py
```diff
@@ -64,38 +65,34 @@ def ratings(self):
...
 
-        These are the mean, rounded to nearest 10, +/- 100, assuming sigma <= 200
-        """
-        # TODO: Figure out what to do with these boundaries
-        mu, _ = self.ratings[0]  # Takes the rating of the first player, only works for 1v1
-        rounded_mu = int(math.ceil(mu / 10) * 10)
-        return rounded_mu - 100, rounded_mu + 100
+    @property
+    def boundary_75(self):
+        """ Achieves roughly 75% quality. FIXME - why is it MORE restrictive??? """
```
Askaholic: I guess my TODO comment was poorly worded. But how are these boundaries going to make sense in a team context?

server/rating.py
```diff
@@ -0,0 +1,39 @@
+from enum import Enum
+from collections.abc import MutableMapping
+from trueskill import Rating
+
+
+# Values correspond to legacy table names. This will be fixed when db gets
```
Askaholic: :+1:

server/games/custom_game.py
Wesmania: You want something else than an enum? Is it supposed to be populated from the db?

server/games/game.py
Wesmania: Copypasted from BaseGame, didn't touch it yet.

server/matchmaker/search.py
Wesmania: Dunno, I didn't want to change behaviour here yet.

Askaholic:
Popped this onto the test server real quick and got the following error:
```
INFO     Aug 30  07:29:05 root                           Using StatsD server: 
ERROR    Aug 30  07:29:05 root                           Failure booting server connect() got an unexpected keyword argument 'loop'
  File "server.py", line 68, in <module>
    loop=loop
TypeError: connect() got an unexpected keyword argument 'loop'
```

Wesmania:
Server launch issue should be fixed now.

server/games/game.py
Askaholic: Can you please fix or delete this comment, because it's wrong

PR #489, opened 2019-08-29
Add endpoint for requesting matchmaker info

Askaholic:
Need this so the client can display the queue time immediately.

https://github.com/FAForever/downlords-faf-client/pull/1364

### DIFF
--- a/server/lobbyconnection.py
+++ b/server/lobbyconnection.py
@@ -209,8 +209,8 @@ class LobbyConnection():
                 maps.append(json_to_send)
 
         self.protocol.send_messages(maps)
-]
-    async def send_matchmaker_info(self):
+
+    async def command_matchmaker_info(self, message):
         self.send({
             'command': 'matchmaker_info',
             'queues': [queue.to_dict() for queue in self.ladder_service.queues.values()]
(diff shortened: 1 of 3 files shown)


Katharsas:
Im not sure about this approach. Does the server respond to a ladder search stop by sending matchmaker info? Then the client could send matchmaking stop when he needs the frist message. Then we don't need to rely on timings to be correct and Ladder1v1Controller in client to be initialized at the right time.

Askaholic:
The only thing I don't like about that, is it will cause the queue info to be broadcast to every player whenever someone else logs in. Also as it stands right now, what if you log in just before the queue pops anyways? There's no guarantee that you wont get a message immediately on login

Katharsas:
I don't understand the last two sentences. Getting a message too early is not a problem in terms of client stability. But that message will be ignored. So the client might not get the message on login, which will effectivly look like a bug from a user perspective.

But i understand that the broadcast problem. When this gets deployed to the test server i will check if it works reliably. The proper solution would be do to have some endpoint/command where the client can request a matchmaker info for itself. But i guess thats much more work.

Issue #487, opened 2019-08-28
New players don't get 1v1 ladder matches


cleborys:
It seems to me that new players with none or very few ladder games cannot get matched in the 1v1 queue. I can observe this personally on mine and a friend's account and something similar seems to be indicated [on this reddit thread](https://www.reddit.com/r/FAF/comments/cq0z9c/how_does_the_ranked_1v1_works/).
I myself tried the ladder once on July 7 before the [Better 1v1 Matchmaking release](https://github.com/FAForever/server/releases/tag/v1.1) and was matched immediately. I tried next two weeks later in late July and can ever since not get a match. (I am running a matchmaking search right now while I type this.) A friend of mine cannot find matches either, even after half an hour of waiting.

Here is what I understand of the current matchmaking algorithm as it pertains to new players:

A new player will be assigned rating 1500 and deviation 500:
https://github.com/FAForever/server/blob/35a2927b27cdb5a91f4421574d3ecbef806f4401/server/players.py#L45-L46

For new players with less than `NEWBIE_MIN_GAMES` however, this rating will be adjusted to the weighted average with `NEWBIE_BASE_MEAN`:
https://github.com/FAForever/server/blob/35a2927b27cdb5a91f4421574d3ecbef806f4401/server/matchmaker/search.py#L57-L59
(In release v1.1 this seems to be calculated on `numGames` which is the number of global games, not ladder games,  but this has since been addressed in ###. Is the server running on v1.1 or a later fix, currently?)
These config values above have default values of ten games and a rating of 500:
https://github.com/FAForever/server/blob/35a2927b27cdb5a91f4421574d3ecbef806f4401/server/config.py#L41-L42
I am not sure why the `NEWBIE_BASE_MEAN` was chosen to be different from the inital rating of 1500, but I guess this means that new players start with a high rating for trueskill purposes but arematched with a low rating that is increasing from 500 to 1500 (or wherever it ends up) during the first ten games.

For potential matches, the rating is converted to a game quality projection via trueskill. This takes the adjusted rating into account but asserts that the raw rating is not None or zero:
https://github.com/FAForever/server/blob/35a2927b27cdb5a91f4421574d3ecbef806f4401/server/matchmaker/search.py#L143-L150
(I think the assert fails if any of the players has a rating of exactly zero, since Python evaluates the `int` 0 to `False`. Is that intentional?)

A match is acceptable if it is above the `match_threshold` for both searches.
[...]

PR #490, opened 2019-08-29
Fix players with high deviations not getting 1v1 ladder matches

cleborys:
See issue #487.

Two main changes:
1. Matchmaking graph for stable marriage will include top five matches for every player *of those who would accept the match* 
2. Minimum acceptable match quality is calculated as 80% of quality in a game against yourself, not hard-coded in `_deviation_quality`

### DIFF
--- a/server/matchmaker/algorithm.py
+++ b/server/matchmaker/algorithm.py
@@ -121,5 +121,11 @@ def _rank_all(searches: List[Search]) -> Dict[Search, List[Search]]:
 
 
 def _rank_partners(search: Search, others: Iterable[Search]) -> List[Search]:
-    matchable_others = [other for other in others if search.matches_with(other)]
-    return heapq.nlargest(SM_NUM_TO_RANK, matchable_others, key=lambda other: search.quality_with(other))
+    return heapq.nlargest(
+        SM_NUM_TO_RANK, 
+        filter(
+            lambda other: search.matches_with(other),
+            others
+        ), 
+        key=lambda other: search.quality_with(other)
+    )
--- a/server/matchmaker/algorithm.py
+++ b/server/matchmaker/algorithm.py
@@ -29,7 +29,9 @@ class StableMarriage(object):
         self.searches = searches
 
     def find(self) -> List[Match]:
-        "Perform stable matching"
+        """Perform SM_NUM_TO_RANK runs of the stable matching algorithm. 
+        Assumes that _rank_all only returns edges whose matches are acceptable
+        to both parties."""
         ranks = _rank_all(self.searches)
         self.matches: Dict[Search, Search] = {}
 
@@ -55,8 +57,6 @@ class StableMarriage(object):
                     search, preferred, search.quality_with(preferred),
                     search.match_threshold, preferred.match_threshold
                 )
-                if not search.matches_with(preferred):
-                    continue
 
                 self._propose(search, preferred)
 
--- a/tests/unit_tests/test_matchmaker_queue.py
+++ b/tests/unit_tests/test_matchmaker_queue.py
@@ -46,15 +46,29 @@ def test_search_threshold(mocker, loop, matchmaker_players):
     assert s.match_threshold >= 0
 
 
-def test_search_threshold_of_old_players_is_high(mocker, loop):
+def test_search_threshold_of_single_old_players_is_high(mocker, loop):
     old_player = Player('experienced_player', player_id=1, ladder_rating=(1500, 50), ladder_games=(config.NEWBIE_MIN_GAMES + 1))
     s = Search([old_player])
     assert s.match_threshold >= 0.6
 
 
-def test_search_threshold_of_new_players_is_low(mocker, loop):
+def test_search_threshold_of_team_old_players_is_high(mocker, loop):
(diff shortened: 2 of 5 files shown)


cleborys:
Code coverage decreased because I removed some now obsolete lines, so the overall file got shorter, while the previously uncovered line remains uncovered. I am not sure what `Search.done` does, exactly, so I can't cover that line right now.

server/matchmaker/algorithm.py
```diff
@@ -121,4 +121,5 @@ def _rank_all(searches: List[Search]) -> Dict[Search, List[Search]]:
 
 
 def _rank_partners(search: Search, others: Iterable[Search]) -> List[Search]:
-    return heapq.nlargest(SM_NUM_TO_RANK, others, key=lambda other: search.quality_with(other))
+    matchable_others = [other for other in others if search.matches_with(other)]
+    return heapq.nlargest(SM_NUM_TO_RANK, matchable_others, key=lambda other: search.quality_with(other))
```
Askaholic:
```suggestion
    return heapq.nlargest(SM_NUM_TO_RANK, filter(lambda other: search.matches_with(other), others), key=lambda other: search.quality_with(other))
```
You could use a filter

server/matchmaker/search.py
```diff
@@ -128,17 +118,15 @@ def search_expansion(self) -> float:
...
-        thresholds = []
-        for rating in self.ratings:
-            _, deviation = rating
 
-            for d, q in self._deviation_quality.items():
-                if deviation >= d:
-                    thresholds.append(max(q - self.search_expansion, 0))
-        return min(thresholds)
+        quality_of_game_against_yourself = self.quality_with(self)
+        return max(0.8 * quality_of_game_against_yourself - self.search_expansion, 0)
```
Askaholic: Keep in mind that a `Search` object may represent an entire party of players, and not necessarily just 1 player. We will need this to work for team matchmaking as well.

tests/integration_tests/test_server.py
```diff
@@ -53,7 +53,7 @@
     assert msg == {
         'command': 'notice',
         'style': 'error',
-        'text': 'You are banned from FAF for 981 years.\n Reason :\n Test permanent ban'}
+        'text': 'You are banned from FAF for 980 years.\n Reason :\n Test permanent ban'}
```
Askaholic: Yea this is annoying. Mazor fixed it in his PR so it will say "forever" once the time reaches a certain amount.

tests/unit_tests/test_matchmaker_queue_algorithm.py
```diff
@@ -43,6 +56,19 @@ def test_stable_marriage():
...
+    old1 = Search([p(2300, 75, name='old1', ladder_games=100)])
+    old2 = Search([p(2350, 75, name='old2', ladder_games=200)])
+
+    searches = [new1, new2, old1, old2]
+
+    matches = algorithm.stable_marriage(searches)
+
+    assert (new1, new2) in matches
+    assert (old1, old2) in matches
+
```
Askaholic: Out of curiosity, what about when 4 players have the same mean but different deviation? 

```
new1 = Search([p(1500, 500, name='new1', ladder_games=0)])
new2 = Search([p(1500, 500, name='new2', ladder_games=0)])
old1 = Search([p(500, 75, name='old1', ladder_games=100)])
old2 = Search([p(500, 75, name='old2', ladder_games=100)])
```
500 would be the adjusted mean for players with 0 games.

I would guess the new players should match with each other and the experienced players with each other.

server/matchmaker/search.py
cleborys: I believe this should work as is but will add some tests.

tests/integration_tests/test_server.py
cleborys: So I will leave it as is until that is pushed.

tests/unit_tests/test_matchmaker_queue_algorithm.py
cleborys: This test should be expanded, you're right.

tests/integration_tests/test_server.py
cleborys: So I will leave it as above to make the tests pass and adapt if their request is pulled.

server/matchmaker/search.py
cleborys: See `test_search_threshold_of_team_old_players_is_high`.

tests/unit_tests/test_matchmaker_queue_algorithm.py
cleborys: See `test_stable_marriage_matches_new_players_with_new_and_old_with_old_if_same_mean`.

tests/unit_tests/test_matchmaker_queue.py
```diff
@@ -46,6 +46,32 @@ def test_search_threshold(mocker, loop, matchmaker_players):
     assert s.match_threshold >= 0
 
 
+def test_search_threshold_of_single_old_players_is_high(mocker, loop):
```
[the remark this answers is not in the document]
cleborys: Yes, I was wondering but thought I might be missing something. I'll remove what I think makes sense and see if it passes :)

PR #491, opened 2019-09-01
Increase search expansion per queue pop

cleborys:
Not an essential change, but avoids weird boundary cases where searches will update their match threshold several times or not at all between having the chance to be matched with players.
I chose the rate so that it will reach the maximum expansion after five queue pops, which is after maximum 15 minutes.
This would hopefully cut down queueing times of players with few quality match opportunities slightly.

(@Askaholic as the diffs are hard to read before the base branch is merged, I invited you to my fork.)


server/matchmaker/search.py
```diff
@@ -110,26 +111,25 @@ def search_expansion(self) -> float:
...
-        if elapsed > END:
-            return MAX
-
-        return (MAX / (END - START)) * (elapsed - START)
+        The threshold will expand linearly with every failed matching attempt
+        until it reaches the specified MAX.
+        """
+        
+        return min(
+                self._failed_matching_attempts * config.LADDER_SEARCH_EXPANSION_STEP,
```
Askaholic: Extra indent

server/matchmaker/search.py
```diff
@@ -110,26 +111,25 @@ def search_expansion(self) -> float:
...
+
+    def register_failed_matching_attempt(self) -> int:
+        """
+        Signal that matchmaker tried to match this search but was unsuccessful
+        and increase internal counter by one.
+        Returns number of failed matching attempts after increase.
+        """
+
+        self._failed_matching_attempts += 1
+        return self._failed_matching_attempts
```
Askaholic: Not a fan of the return value, especially because it's just used for logging. I'd say just access the value directly when you need it. That way there is no confusion as to whether the return value is the value before increment or after. It will be completely clear.

PR #492, opened 2019-09-01
Match newbies randomly

cleborys:
As suggested: match newbie players randomly if the stable marriage didn't match them.

Prefers to match two unmatched newbies with each other before matching with a "normal" player.

### DIFF
--- a/server/config.py
+++ b/server/config.py
@@ -1,7 +1,7 @@
 import logging
+import os
 
 import trueskill
-import os
 
 # Logging setup
 TRACE = 5
@@ -62,10 +62,11 @@ LADDER_SEARCH_EXPANSION_START = int(os.getenv('LADDER_SEARCH_EXPANSION_START', 6
 LADDER_SEARCH_EXPANSION_END = int(os.getenv('LADDER_SEARCH_EXPANSION_END', 60 * 20))
 LADDER_SEARCH_EXPANSION_MAX = float(os.getenv('LADDER_SEARCH_EXPANSION_MAX', 0.25))
 
-# The maximum amount of time (in seconds) to wait if no one is searching.
+# The maximum amount of time (in seconds) to wait between pops.
 QUEUE_POP_TIME_MAX = int(os.getenv('QUEUE_POP_TIME_MAX', 60 * 3))
-# The number of searches in the queue required for the queue time to be cut in
-# half. See https://www.desmos.com/calculator/v3tdrjbqub.
-QUEUE_POP_TIME_SCALE_FACTOR = int(os.getenv('QUEUE_POP_TIME_SCALE_FACTOR', 20))
+# The number of players we would like to have in the queue when it pops. The
+# queue pop time will be adjusted based on the current rate of players queuing
+# to try and hit this number.
+QUEUE_POP_DESIRED_PLAYERS = int(os.getenv('QUEUE_POP_DESIRED_PLAYERS', 8))
 # How many previous queue sizes to consider
 QUEUE_POP_TIME_MOVING_AVG_SIZE = int(os.getenv('QUEUE_POP_TIME_MOVING_AVG_SIZE', 5))
--- a/server/matchmaker/algorithm.py
+++ b/server/matchmaker/algorithm.py
@@ -121,5 +121,11 @@ def _rank_all(searches: List[Search]) -> Dict[Search, List[Search]]:
 
 
 def _rank_partners(search: Search, others: Iterable[Search]) -> List[Search]:
-    matchable_others = [other for other in others if search.matches_with(other)]
-    return heapq.nlargest(SM_NUM_TO_RANK, matchable_others, key=lambda other: search.quality_with(other))
+    return heapq.nlargest(
+        SM_NUM_TO_RANK, 
+        filter(
+            lambda other: search.matches_with(other),
+            others
+        ), 
+        key=lambda other: search.quality_with(other)
+    )
--- a/server/matchmaker/algorithm.py
+++ b/server/matchmaker/algorithm.py
@@ -1,6 +1,5 @@
 import heapq
-from collections import deque
-from typing import Deque, Dict, Iterable, List, Set
+from typing import Dict, Iterable, List, Set
 
 from ..decorators import with_logger
 from .search import Match, Search
@@ -15,8 +14,6 @@ SM_NUM_TO_RANK = 5
 ################################################################################
 #                                Implementation                                #
 ################################################################################
-
-last_queue_amounts: Deque[int] = deque()
 
 
 def stable_marriage(searches: List[Search]) -> List[Match]:
--- a/server/matchmaker/algorithm.py
+++ b/server/matchmaker/algorithm.py
@@ -57,7 +57,33 @@ class StableMarriage(object):
 
                 self._propose(search, preferred)
 
+        self._forcefully_match_unmatched_newbies()
+
         return self._remove_duplicates()
+
+    def _forcefully_match_unmatched_newbies(self):
+        unmatched_newbies = [
+            search in self.searches 
+            if search.is_single_ladder_newbie()
+            and not search in self.matches
+        ]
+
+        while 2 <= len(unmatched_newbies):
+            newbie1 = unmatched_newbies.pop()
+            newbie2 = unmatched_newbies.pop()
+            self._match(newbie1, newbie2)
+
+        if len(unmatched_newbies) == 1:
+            newbie = unmatched_newbies[0]
+            opponent = next(
+                (search for search in  self.searches
+                if search != newbie
+                and not search in self.matches),
+                None
+            )
+            if opponent is not None:
+                self._match(newbie, opponent)
+
 
     def _remove_duplicates(self) -> List[Match]:
         matches_set: Set[Match] = set()
--- a/server/matchmaker/algorithm.py
+++ b/server/matchmaker/algorithm.py
@@ -29,7 +29,9 @@ class StableMarriage(object):
         self.searches = searches
 
     def find(self) -> List[Match]:
-        "Perform stable matching"
+        """Perform SM_NUM_TO_RANK runs of the stable matching algorithm. 
+        Assumes that _rank_all only returns edges whose matches are acceptable
+        to both parties."""
         ranks = _rank_all(self.searches)
         self.matches: Dict[Search, Search] = {}
 
@@ -55,8 +57,6 @@ class StableMarriage(object):
                     search, preferred, search.quality_with(preferred),
                     search.match_threshold, preferred.match_threshold
                 )
-                if not search.matches_with(preferred):
-                    continue
 
                 self._propose(search, preferred)
 
--- a/server/matchmaker/algorithm.py
+++ b/server/matchmaker/algorithm.py
@@ -121,4 +121,5 @@ def _rank_all(searches: List[Search]) -> Dict[Search, List[Search]]:
 
 
 def _rank_partners(search: Search, others: Iterable[Search]) -> List[Search]:
-    return heapq.nlargest(SM_NUM_TO_RANK, others, key=lambda other: search.quality_with(other))
+    matchable_others = [other for other in others if search.matches_with(other)]
+    return heapq.nlargest(SM_NUM_TO_RANK, matchable_others, key=lambda other: search.quality_with(other))
--- a/tests/unit_tests/test_matchmaker_queue.py
+++ b/tests/unit_tests/test_matchmaker_queue.py
@@ -68,26 +68,26 @@ def test_search_threshold(matchmaker_players):
     assert s.match_threshold >= 0
 
 
-def test_search_threshold_of_single_old_players_is_high(mocker, loop):
+def test_search_threshold_of_single_old_players_is_high():
     old_player = Player('experienced_player', player_id=1, ladder_rating=(1500, 50), ladder_games=(config.NEWBIE_MIN_GAMES + 1))
     s = Search([old_player])
     assert s.match_threshold >= 0.6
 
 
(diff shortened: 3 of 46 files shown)


JaggedAppliance:
I'm delighted you're working on this

tests/unit_tests/test_matchmaker_queue_algorithm.py
```diff
@@ -82,3 +124,36 @@ def test_stable_marriage_unmatch():
...
+
+
+def test_unmatched_newbies_forcefully_match_pros():
+    newbie = Search([p(0, 500, ladder_games=0)])
+    pro = Search([p(2500, 10, ladder_games=100)])
+
+    searches = [newbie, pro]
+    matches = algorithm.stable_marriage(searches)
+
+    assert (newbie, pro) in matches or (pro, newbie) in matches
```
Askaholic: Hmm. Now I'm not so sure if it's a good idea to match pros with noobs. I think it's common for pros to sit in queue while trying to find an opponent in aeolus. it would be unfortunate for a new player to get matched against one of the top ladder players. Neither person in that match would have a good time.

Maybe instead of doing a completely random match there could be some sort of heuristic like "any player with rating lower than 2 standard deviations plus my mean" or something like that. Maybe it shouldn't even use standard deviation and should just be a hardcoded number say 200?

Either way I think it's probably better to avoid this situation where a complete new player gets matched with a top ladder player. Good test though!

server/matchmaker/algorithm.py
```diff
@@ -52,13 +54,37 @@ def find(self) -> List[Match]:
...
-                if not search.matches_with(preferred):
-                    continue
 
                 self._propose(search, preferred)
 
+        self._forcefully_match_unmatched_newbies()
+
         return self._remove_duplicates()
 
+    def _forcefully_match_unmatched_newbies(self):
```
Askaholic: Is it practical to move this out of the `StableMarriage` class? This isn't part of the stable marriage implementation any more. I had refactored to a class to cut down on the explicit parameter passing, but maybe the class should be renamed?

server/matchmaker/algorithm.py
```diff
@@ -52,13 +54,37 @@ def find(self) -> List[Match]:
...
+        unmatched_newbies = [
+            search for search in self.searches 
+            if search.is_single_ladder_newbie()
+            and not search in self.matches
+        ] 
+
+        while 2 <= len(unmatched_newbies):
```
Askaholic: Any reason not to do
```suggestion
        while len(unmatched_newbies) >= 2:
```
? I feel like this is more readable

server/matchmaker/algorithm.py
```diff
@@ -52,13 +54,37 @@ def find(self) -> List[Match]:
...
+            newbie2 = unmatched_newbies.pop()
+            self._match(newbie1, newbie2)
+
+        if len(unmatched_newbies) == 1:
+            newbie = unmatched_newbies[0]
+            opponent = next(
+                (search for search in  self.searches
+                if search != newbie
+                and not search in self.matches),
+                None
```
Askaholic: Woah. I didn't know `next` had a default parameter! Maybe passing it as a keyword argument could increase readability.

server/matchmaker/algorithm.py
```diff
@@ -52,13 +54,37 @@ def find(self) -> List[Match]:
...
+
+        while 2 <= len(unmatched_newbies):
+            newbie1 = unmatched_newbies.pop()
+            newbie2 = unmatched_newbies.pop()
+            self._match(newbie1, newbie2)
+
+        if len(unmatched_newbies) == 1:
+            newbie = unmatched_newbies[0]
+            opponent = next(
+                (search for search in  self.searches
```
Askaholic:
```suggestion
                (search for search in self.searches
```

tests/unit_tests/test_matchmaker_queue_algorithm.py
cleborys: Then I'd suggest we only match with with players who have rating mean < 2000? Looking at the "Rating distribution" graph in the 1v1 lobby that should be "basically everyone" :D

server/matchmaker/algorithm.py
cleborys: I once read a nice blog post on style suggesting to only use "<" and never ">" because it makes things somewhat natural to think about when reading (since they are always sorted from left to right increasing/non-decreasing).
But I agree that it looks weird if one isn't used to it and it's certainly not important here ;)

tests/unit_tests/test_matchmaker_queue_algorithm.py
```diff
@@ -82,3 +124,36 @@ def test_stable_marriage_unmatch():
...
+    pro = Search([p(2500, 10, ladder_games=100)])
+
+    searches = [newbie1, pro, newbie2]
+    matches = algorithm.stable_marriage(searches)
+
+    assert (newbie1, newbie2) in matches or (newbie2, newbie1) in matches
+
+
+def test_unmatched_newbies_forcefully_match_pros():
+    newbie = Search([p(0, 500, ladder_games=0)])
```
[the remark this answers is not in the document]
cleborys: I think I just wanted to make sure no later change on the matching algorithm will match these two from the start (but the low deviation on the pro probably still ensures that for any sensible matchmaking strategy.)

server/matchmaker/algorithm.py
cleborys: I am not sure how to do that :( Trying to pass it as `next(iterator, default=None)` tells me that `TypeError: next() takes no keyword arguments`.

To make it clear that it's a default the following would work, but I'm not convinced it's clean...
 ```suggestion
           default_if_no_available_opponent = None
            opponent = next(
                (search for search in  self.searches
                if search != newbie
                and not search in self.matches),
                default_if_no_available_opponent
            )
            if opponent is not default_if_no_available_opponent:
                self._match(newbie, opponent)
```

Comment (cleborys) [reply server/matchmaker/algorithm.py]: (An alternative is of course to try/except catch the `StopIterationError`, but that seemed less readable)

File: tests/unit_tests/test_matchmaker_queue_algorithm.py (hunk 1 above)
Comment (cleborys) [reply tests/unit_tests/test_matchmaker_queue_algorithm.py]: (side node: I needed a word for "not a newbie in the tests so I called them "pros". I now call everyone with less than ten games a "newbie", everyone with rating > 2500 a "top" player and everyone else a "pro". Maybe they should be `normal_player` instead.)

Comment (Askaholic) [reply tests/unit_tests/test_matchmaker_queue_algorithm.py]: Ok, that’s a good solution too. Note that the ratings in the ladder tab are calculated as `mean - 3 * deviation` so they will be a bit lower than the true skill mean

File: server/matchmaker/algorithm.py (hunk 3 above)
Comment (Askaholic) [reply server/matchmaker/algorithm.py]: Interesting. I don’t think I agree with that reasoning :P, unless maybe you establish it as some standard for the entire code base

File: server/matchmaker/algorithm.py (hunk 2 above)
Comment (cleborys) [reply server/matchmaker/algorithm.py]: These are some slightly bigger changes, so could you tell me what you think about this https://github.com/cleborys/server/pull/5 before I include them in the PR? :)

File: server/matchmaker/algorithm.py (hunk 4 above)
Comment (Askaholic) [reply server/matchmaker/algorithm.py]: Well it's not a big deal then.

File: server/matchmaker/algorithm.py [hunk 7]
```diff
@@ -118,4 +150,11 @@ def _rank_all(searches: List[Search]) -> Dict[Search, List[Search]]:
 
 
 def _rank_partners(search: Search, others: Iterable[Search]) -> List[Search]:
-    return heapq.nlargest(SM_NUM_TO_RANK, others, key=lambda other: search.quality_with(other))
+    return heapq.nlargest(
+        SM_NUM_TO_RANK, 
+        filter(
```
Askaholic: Ok, so the only problem with this is it makes the debugging output kindof pointless. Because before it would print out all of the qualities and thresholds between people and you could see why they weren't matching. Now people don't match and there's no debugging output to show for it. If they are going to match anyways, then it prints out their qualities.

server/matchmaker/algorithm.py
Askaholic: Oh yea that looks pretty nice!

server/matchmaker/algorithm.py
```diff
@@ -87,42 +93,112 @@ def _propose(self, search: Search, preferred: Search):
...
+            for search in searches
+        }
+
+
+    @staticmethod
+    def _get_top_matches(search: Search, others: Iterable[Search]) -> List[Search]:
+        def is_possible_match(other: Search) -> bool:
+            quality_log_string = (
+                f"Quality between {search} and {other}: {search.quality_with(other)}"
+                + f" thresholds: [{search.match_threshold}, {other.match_threshold}]."
```
Askaholic: Don't need the plus
```suggestion
                f" thresholds: [{search.match_threshold}, {other.match_threshold}]."
```

server/matchmaker/algorithm.py
```diff
@@ -87,42 +93,112 @@ def _propose(self, search: Search, preferred: Search):
...
+                return True
+            else:
+                _MatchingGraph._logger.debug(
+                    quality_log_string + f" Will be discarded."
+                )
+                return False
+
+        return heapq.nlargest(
+            SM_NUM_TO_RANK, 
+            filter(
```
Askaholic: This can fit on one line now

tests/unit_tests/test_matchmaker_queue.py
```diff
@@ -34,10 +34,32 @@ def matchmaker_players_all_match():
...
+    pro, _, _, _, _, newbie = matchmaker_players
+
+    single_newbie = Search([newbie])
+    single_pro = Search([pro])
+    two_newbies = Search([newbie, newbie])
+    two_pros = Search([pro, pro])
+    two_mixed = Search([newbie, pro])
+
+    assert single_newbie.is_single_ladder_newbie()
+    assert single_pro.is_single_ladder_newbie() == False
```
Askaholic: Use `is` for singletons like `True`, `False` and `None`
```suggestion
    assert single_pro.is_single_ladder_newbie() is False
```

server/matchmaker/algorithm.py
```diff
@@ -87,42 +93,112 @@ def _propose(self, search: Search, preferred: Search):
...
+        self.searches = searches
+        self.matches: Dict[Search, Search] = {}
+
+    def find(self) -> List[Match]:
+        self.matches.update(StableMarriage(self.searches).find())
 
-def _rank_all(searches: List[Search]) -> Dict[Search, List[Search]]:
-    """ Returns searches with best quality for each search.
+        remaining_searches = [
+                search for search in self.searches
```
Askaholic: I think there's an extra indent here
```suggestion
            search for search in self.searches 
```

server/matchmaker/algorithm.py
```diff
@@ -87,42 +93,112 @@ def _propose(self, search: Search, preferred: Search):
...
+    def _get_top_matches(search: Search, others: Iterable[Search]) -> List[Search]:
+        def is_possible_match(other: Search) -> bool:
+            quality_log_string = (
+                f"Quality between {search} and {other}: {search.quality_with(other)}"
+                + f" thresholds: [{search.match_threshold}, {other.match_threshold}]."
+            )
+
+            if search.matches_with(other):
+                _MatchingGraph._logger.debug(
+                    quality_log_string + f" Will be considered during matchmaking."
```
Askaholic: In general string formatting is better than concatenation because it will require less copying. If you want to go down the concatenation route, then the string doesn't need to be an f-string.
```suggestion
                    f"{quality_log_string} Will be considered during matchmaking."
```

server/matchmaker/algorithm.py
```diff
@@ -87,42 +93,112 @@ def _propose(self, search: Search, preferred: Search):
...
+                )
+                return True
+            else:
+                _MatchingGraph._logger.debug(
+                    quality_log_string + f" Will be discarded."
```
Askaholic:
```suggestion
                    f"{quality_log_string} Will be discarded."
```

server/matchmaker/algorithm.py
cleborys: Oh, that's nice, thank you!

cleborys:
I'll give YAPF a try! Normally I use `black` and `flake8`, but I didn't want to impose their doctrine on all of the codebase... The YAPF approach of optimizing instead of only making it PEP compliant looks nice, too!

server/config.py
```diff
@@ -40,6 +40,7 @@
 
 NEWBIE_BASE_MEAN = int(os.getenv('NEWBIE_BASE_MEAN', 500))
 NEWBIE_MIN_GAMES = int(os.getenv('NEWBIE_MIN_GAMES', 10))
+TOP_PLAYER_MIN_RATING = int(os.getenv('TOP_PLAYER_MIN_RATING', 2000))
```
Askaholic: Talked to FTXCommando about this for a while, we're thinking 1600 is probably better.

cleborys: Sure!

Askaholic:
I haven't heard of `black` before. The only reason I started using YAPF was because it has a lot of configuration options and I hoped I would be able to configure it to do what I wanted, but that doesn't seem to be the case. What I was really looking for was a formatter that would actually be able to rewrite some of the code by for example converting this:

```python
d = dict(item1=value1, item2=value2)
```
into this
```python
d = {
    "item1": value1,
    "item2: value2
}
```

But unfortunately yapf doesn't mess with the token stream. Looks like black does a few things like replacing string literals that use quotes with double quote versions (which I think would be super nice because I always find myself being indecisive over which type of quote to use) so I'll give it a shot.

PR #493, opened 2019-09-07
Install aiomysql from git and re-enable warnings

Askaholic:
I submitted a PR to `aiomysql` to fix the deprecation warnings https://github.com/aio-libs/aiomysql/pull/403 which was merged, but hasn't been deployed as apparently the people with the ability to do so have disappeared into the void. So instead, we can just install from git to get the latest version

### DIFF
--- a/Dockerfile
+++ b/Dockerfile
@@ -3,7 +3,7 @@ FROM python:3.6-slim
 # Apt-install c++ build tools and cleanup temporary files afterwards
 RUN apt-get update
 RUN apt-get install -y --no-install-recommends \
-        build-essential && \
+        build-essential git && \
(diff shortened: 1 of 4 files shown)


norraxx:
What do you miss in aiomysql, that is needed for you?
Did you know, that pipenv can install from git?

norraxx:
Do you know, that you can uninstall things that aren't used in production? For example git?

Askaholic:
Look at the PR I linked. It generates hundreds of warnings from inside aiomysql that make it very hard to find warnings that actually are caused by our code

norraxx:
Lol! Np :-)

PR #494, opened 2019-09-08
Auto-rollback DB changes after every test

Wesmania:
This allows us to remove a few DB clearing methods and queries.

### DIFF
--- a/tests/utils.py
+++ b/tests/utils.py
@@ -1,6 +1,8 @@
 import functools
 import asyncio
 import asynctest
+from asyncio import Lock, Event
+from aiomysql.sa import create_engine
 
 
 # Copied over from PR #113 of pytest-asyncio. It will probably be available in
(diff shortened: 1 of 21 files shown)


tests/utils.py
```diff
@@ -66,3 +68,77 @@ def deco(f):
...
+    async def close(self):
+        if self.engine is None:
+            return
+
+        async with self._lock:
+            self._done.set()
+            await self._keep
+            self.engine.close()
+            await self.engine.wait_closed()
+            self.engine = None
```
norraxx: Did you know, that you don't need that class at all?
See here, how to work with database as yield fixture:
https://github.com/norraxx/faf-aio-replayserver/blob/master/tests/fixtures/db_fixtures.py#L7
And if you'll need cursor, you can use that one fixture directly.
https://github.com/norraxx/faf-aio-replayserver/blob/master/tests/fixtures/db_fixtures.py#L48

You don't need to "overwrite" database connector behavior, let it be. Just ensure call `begin` at the beginning of the test, and `rollback` at the end.

That will allow you access to the database, make queries, etc.. during whole test, and it will clean everything else at the end of the each test.

Wesmania: This gives me a single connection. The server expects to have access to an entire pool. The class is intended to mock that pool with a single connection.

norraxx: Check what I've send again and think how it works and then think what you've wrote.

And then you'll understand that if you'll use my way, you won't need to touch `async with self._db.engine.acquire() as conn` and your tests will work properly.

Anyway: I'm impressed how you understand code, very fast, in 9 minutes you did great analysis to approve your solution to not to change and do anything else. ;-)

I think, I'll make Mazor's reject counter, and add there time per rejection.

Wesmania: Your solution creates a connection with a different interface - not the sqlalchemy-wrapped connection that the lobby server uses, but one used to directly execute SQL strings. The only way I found in the API to create the former is to create an engine with a pool, then grab a connection from that pool. If you know of a better way to create and provide an `SAConnection` instance, please suggest one.

Furthermore, your solution:
* Utilizes module-level globals for testing, which risks inter-test interaction,
* Has separate code paths in non-testing code for running tests, which is a horrible practice.

I don't intend to do either of these things.

I'd also like you to use an impersonal tone during reviews, as I do - the purpose of these is to be informative and direct so code can be improved. I have no time and no interest in guessing what I'm "supposed to understand" that you're withholding from me, nor do I care about any clever remarks unrelated to code itself.

PR #495, opened 2019-09-14
Clean up game outcome handling

Wesmania:
This moves the logic determining game's outcome for players into a separate class. Methods for choosing an outcome from multiple reports are merged into one. TODO: discuss if the method chosen is appropriate.

The PR is based on top of #494, so it has some extra commits.

### DIFF
--- a/server/games/game.py
+++ b/server/games/game.py
@@ -725,7 +725,7 @@ class Game:
                 options['Color'],
                 options['Team'],
                 options['StartSpot'],
-                mean, dev, 0, -1
+                mean, dev, 0, 0
             ))
         if not query_args:
             self._logger.warning("No player options available!")
--- a/server/games/game.py
+++ b/server/games/game.py
@@ -442,13 +442,9 @@ class Game:
         scores = {}
         for player in self.players:
             army = self.get_player_option(player.id, 'Army')
-            try:
-                score = self.get_army_score(army)
-                scores[player] = score
-                self._logger.info('Result for army %s, player: %s: %s', army, player, score)
-            except KeyError:
-                # Default to -1 if there is no result
-                scores[player] = -1
+            score = self.get_army_score(army)
+            scores[player] = score
+            self._logger.info('Result for army %s, player: %s: %s', army, player, score)
 
         async with self._db.acquire() as conn:
             rows = []
@@ -797,11 +793,7 @@ class Game:
             else:
                 if team not in team_scores:
                     team_scores[team] = 0
-                try:
-                    team_scores[team] += self.get_army_score(army)
-                except KeyError:
-                    team_scores[team] += 0
-                    self._logger.warning("Missing game result for %s: %s", army, player)
+                team_scores[team] += self.get_army_score(army)
         ranks = [-score for team, score in sorted(team_scores.items(), key=lambda t: t[0])]
         rating_groups = []
         for team in sorted(self.teams):
--- a/server/games/game.py
+++ b/server/games/game.py
@@ -130,7 +130,7 @@ class Game:
         game_mode: str='faf'
     ):
         self._db = database
-        self._results = GameResults()
+        self._results = GameResults(id_)
         self._army_stats = None
         self._players_with_unsent_army_stats = []
         self._game_stats_service = game_stats_service
@@ -759,7 +759,7 @@ class Game:
                 "WHERE id = %s", (new_validity_state.value, self.id))
 
     def get_army_score(self, army):
-        return self._results.score(army, self.id)
+        return self._results.score(army)
 
     def get_army_result(self, player):
         army = self.get_player_option(player.id, 'Army')
--- a/server/games/game.py
+++ b/server/games/game.py
@@ -13,6 +13,7 @@ from trueskill import Rating
 from ..abc.base_game import GameConnectionState, InitMode
 from ..players import Player, PlayerState
 from server.rating import RatingType
+from server.games.game_results import GameOutcome, GameResult, GameResults
 
 FFA_TEAM = 1
 
@@ -103,28 +104,6 @@ class ValidityState(Enum):
     OTHER_UNRANK = 24
 
 
-@unique
-class GameOutcome(Enum):
-    VICTORY = 1
-    DEFEAT = 2
-    DRAW = 3
-    MUTUAL_DRAW = 4
-
-    @staticmethod
-    def from_string(value: str) -> Optional["GameOutcome"]:
-        """
-        :param value: The string to convert from
-
-        :return: VisibilityState or None if the string is not valid
-        """
-        return {
-            "victory": GameOutcome.VICTORY,
-            "defeat": GameOutcome.DEFEAT,
-            "draw": GameOutcome.DRAW,
-            "mutual_draw": GameOutcome.MUTUAL_DRAW
-        }.get(value)
-
-
 class GameError(Exception):
     pass
 
@@ -151,7 +130,7 @@ class Game:
         game_mode: str='faf'
     ):
         self._db = database
-        self._results = {}
+        self._results = GameResults()
         self._army_stats = None
         self._players_with_unsent_army_stats = []
         self._game_stats_service = game_stats_service
@@ -210,17 +189,7 @@ class Game:
 
     @property
     def is_mutually_agreed_draw(self) -> bool:
-        # Don't count non-reported games as mutual draws
-        if not len(self._results):
-            return False
-        for army in self.armies:
-            if army in self._results:
-                for result in self._results[army]:
-                    if result[1] != 'mutual_draw':
-                        return False
-            else:
-                return False
-        return True
+        return self._results.is_mutually_agreed_draw(self.armies)
 
     @property
     def players(self):
@@ -323,7 +292,7 @@ class Game:
         if not self._is_hosted.done():
             self._is_hosted.set_result(value)
 
-    def outcome(self, player: Player) -> Optional[GameOutcome]:
+    def outcome(self, player: Player) -> GameOutcome:
         """
         Determines what the game outcome was for a given player. Did the
         player win, lose, draw?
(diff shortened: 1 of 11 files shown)


server/games/game.py
```diff
@@ -462,7 +406,7 @@ def host_left_lobby() -> bool:
                     await self.mark_invalid(ValidityState.MUTUAL_DRAW)
                     return
 
-                if len(self._results) == 0:
+                if not self._results:
```
cleborys: When is this `True`? Does this maybe rely on `collections.abc.Mapping.__len__` being nonzero or `GameResults.__len__` being nonzero? Or just `self._results` not being `None`? I am uncertain how trucity/casting to bool is handled for user-defined class if there is no `__bool__`.

server/games/game.py
```diff
@@ -836,49 +757,20 @@ def sanitize_name(self, name: str) -> str:
...
-        if len(scores) == 1:
-            return scores.popitem()[0]
-
-        # Return the highest score with the most votes
-        self._logger.info("Conflicting scores (%s) reported for game %s", scores, self)
-        score, _votes = max(scores.items(), key=lambda kv: kv[::-1])
-        return score
+        return self._results.score(army, self.id)
 
     def get_army_result(self, player):
```
cleborys: Maybe this should be called `get_player_outcome`?

Brutus5000:
If you want to clean up game outcomes please write into the db table player_stats into a new column an enum of type VICTORY, DRAW, DEFEAT. That would save me froma lot of trouble in the future

server/games/game_results.py
```diff
@@ -0,0 +1,136 @@
...
+        # Everyone has to agree to a mutual draw
+        for army in player_armies:
+            if army not in self:
+                continue
+            if any(r.outcome is not GameOutcome.MUTUAL_DRAW
+                   for r in self[army]):
+                return False
+        return True
+
+    def outcome(self, army: int) -> GameOutcome:
```
cleborys: Maybe `get_outcome`, making it a verb?

server/games/game.py
Wesmania: According to Python docs, in absence of `__bool__` truthiness is decided by `__len__` being non-zero. So, it's perfectly fine to idiomatically check for non-emptiness.

server/games/game.py
Wesmania: It does return a score (a 'score' value from the game) instead of an outcome ('victory', 'defeat' etc.).

server/games/game_results.py
Wesmania: It returns a value without side effects, so I think it's up to personal taste.

server/games/game_results.py
```diff
@@ -0,0 +1,136 @@
...
+        """
+        if army not in self:
+            return GameOutcome.UNKNOWN
+
+        outcomes = set(r.outcome for r in self[army]
+                       if r.outcome is not GameOutcome.UNKNOWN)
+
+        if len(outcomes) == 1:
+            return outcomes.pop()
+        else:
```
Askaholic: Can drop the else

server/games/game_results.py
```diff
@@ -0,0 +1,136 @@
...
+            return 0
+
+    # FIXME: Should we gather such methods in one class, or place them near
+    # corresponding types?
+    @classmethod
+    async def from_db(cls, database, game_id):
+        results = cls()
+        async with database.acquire() as conn:
+            rows = await conn.execute(
```
Askaholic: Could rewrite this with sqlalchemy models

server/games/game_results.py
```diff
@@ -0,0 +1,136 @@
...
+    VICTORY = 'victory'
+    DEFEAT = 'defeat'
+    DRAW = 'draw'
+    MUTUAL_DRAW = 'mutual_draw'
+    UNKNOWN = 'unknown'
+
+    @classmethod
+    def from_message(cls, msg: str):
+        try:
+            return cls(msg.lower())
```
Askaholic: Interesting, I didn't know that `Enum`'s work this way

server/games/game.py
cleborys: Sotty, I meant `get_army_result`, which takes a `Player` as argument and returns a `GameOutcome` (not `get_army_score`)

server/games/game.py
cleborys: Good to know, thanks!

Askaholic:
> If you want to clean up game outcomes please write into the db table player_stats into a new column an enum of type VICTORY, DRAW, DEFEAT. That would save me froma lot of trouble in the future

Yea, we should really add that. Otherwise the only way of knowing how the server rated a game, is to run the exact version of the rating code that was deployed at the time against the same game results. And as we know, that rating code is flawed at the moment

server/games/game.py
Wesmania: I didn't want to touch `Game` class public methods, though I could rename that if you want.

server/games/game_results.py
Wesmania: The else clause is a one-liner, so I thought it's clearer to keep it.

server/games/game_results.py
Wesmania: Sure. I kept it to make it easier to verify that the code is equivalent, we can change that in a separate commit.

server/games/game.py
cleborys: It just looked off to me, so I'm all for it. I can do it in my later PR if you don't want to :)

tests/unit_tests/test_game.py
```diff
@@ -405,6 +406,8 @@ def custom_game(event_loop, database, game_service, game_stats_service):
             assert new_rating != Rating(*player.ratings[RatingType.GLOBAL])
 
 
+# FIXME - discuss correct get_army_result behaviour
```
Askaholic: So this test now fails because the first result (with one reported victory and one reported defeat) now is reported as an unknown outcome? It might be a good idea to unrank the game due to `UNKNOWN_RESULT` in this case as well.

server/games/game_results.py
```diff
@@ -0,0 +1,136 @@
...
+            return outcomes.pop()
+        else:
+            return GameOutcome.UNKNOWN
+
+    def score(self, army: int, game_id=None):
+        """
+        Pick and return most frequently reported score for an army. If multiple
+        scores are most frequent, pick the largest one.
+        """
+        if army not in self:
```
Askaholic: I think some pieces of the code actually expect this to raise a `KeyError`. Have a look at:
https://github.com/FAForever/server/blob/6b9aec7f04698da4b0843c2a26a82705c5faf503/server/games/game.py#L515
and
https://github.com/FAForever/server/blob/6b9aec7f04698da4b0843c2a26a82705c5faf503/server/games/game.py#L897
It even mentions it in the docstring
https://github.com/FAForever/server/blob/6b9aec7f04698da4b0843c2a26a82705c5faf503/server/games/game.py#L841

We need some tests for `persist_results` because I don't think it ever writes `-1` into the database any more. As a matter of fact that code path is not covered by the tests: https://coveralls.io/builds/25850926/source?filename=server%2Fgames%2Fgame.py#L449
which is probably why I missed this when I rewrote that function in the first place

server/games/game_results.py
```diff
@@ -0,0 +1,136 @@
...
+        return self._back[key]
+
+    def __iter__(self):
+        return iter(self._back)
+
+    def __len__(self):
+        return len(self._back)
+
+    def add(self, result: GameResult):
+        army_results = self._back.setdefault(result.army, [])
```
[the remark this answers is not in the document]
Wesmania: This line is just used to append to or add a list in one line. I'd rather the class behaved like a regular dict.

server/games/game_results.py
```diff
@@ -0,0 +1,136 @@
...
+        """
+        if army not in self:
+            return 0
+
+        if any(r.outcome is GameOutcome.VICTORY for r in self[army]):
+            return 1
+        else:
+            return 0
+
+    # FIXME: Should we gather such methods in one class, or place them near
```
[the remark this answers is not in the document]
Wesmania: I think it's okay for it to be here - `database` is just a handle to the DB used in many places, and `game_id` is a game identifier in it, not a `Game` class internal.

tests/unit_tests/test_game.py
Wesmania: I'll change the test to do that.

server/games/game_results.py
Wesmania: I think I won't restore these KeyErrors and will just fix these catch blocks - we have good neutral values (`0`, and `GameOutcome.UNKNOWN`) for when an army isn't there. It also makes sense in that it's not an exceptional condition - we might have not received any results for a given army yet, so we need a sensible value for an army with no results.

Askaholic: Yea I agree with that approach. Like I mentioned, the code used to write `-1` for unknown scores, although I'm not convinced that this is unambiguous since negative scores are possible. Is it possible to have a score of 0?

server/games/game_results.py
Askaholic: Now that it's not a one liner you should probably get rid of the else. It's just an unnecessary level of indentation

server/games/game_results.py
```diff
@@ -0,0 +1,141 @@
...
+        yet, return None.
+        """
+        if army not in self:
+            return None
+
+        scores = Counter(r.score for r in self[army])
+        if len(scores) == 1:
+            return scores.popitem()[0]
+
+        # FIXME - we pass game id just for logging
```
Askaholic: Remove comment? Seems that you fixed this by making it a member

server/games/game_results.py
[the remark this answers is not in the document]
Askaholic: In that case let's remove the comment

Askaholic:
I think this PR is very close to being ready, but I'd still like to see a regression test that verifies unknown results are reported as score `-1` in the database. This is mostly so that the next time someone changes the code and accidentally modifies that behaviour, at least they will know.

Also I think you have a few `FIXME`' comments that can be removed now, as well as a now extraneous else statement.

Wesmania:
Now that I look at the old code, I'm somewhat confused. Originally -1 was entered when `get_army_score` threw a KeyError, but I can't see how that could happen in old code. It seems to me we never hit that '-1' branch, and that I should modify the new `get_army_score` to just return 0 when there are no results.

PR #497, opened 2019-10-16
Refactor message sending interface to be a coroutine

Askaholic:
This refactors the old callback interface for message sending to use async/await native instead. Not only is the coroutine interface cleaner and more intuitive, it also helps to ensure that the code behaves as we expect it to (for instance backpressure handling) without us needing to even think about edge cases.

Also adds a few additional test cases.

### DIFF
--- a/server/__init__.py
+++ b/server/__init__.py
@@ -27,6 +27,8 @@ from .player_service import PlayerService
 from .game_service import GameService
 from .ladder_service import LadderService
 from .control import init as run_control_server
+from .timing import at_interval
+
 
 __version__ = '0.9.17'
 __author__ = 'Askaholic, Chris Kitching, Dragonfire, Gael Honorez, Jeroen De Dauw, Crotalus, Michael Søndergaard, Michel Jung'
@@ -95,10 +97,7 @@ def run_lobby_server(
     Run the lobby server
     """
 
-    def report_dirties():
-        asyncio.ensure_future(do_report_dirties())
-        loop.call_later(DIRTY_REPORT_INTERVAL, report_dirties)
-
+    @at_interval(DIRTY_REPORT_INTERVAL)
     async def do_report_dirties():
         try:
             dirty_games = games.dirty_games
@@ -153,9 +152,9 @@ def run_lobby_server(
 
     ping_msg = encode_message('PING')
 
-    def ping_broadcast():
-        asyncio.ensure_future(ctx.broadcast_raw(ping_msg))
-        loop.call_later(45, ping_broadcast)
+    @at_interval(45)
+    async def ping_broadcast():
+        await ctx.broadcast_raw(ping_msg)
 
     def make_connection() -> LobbyConnection:
         return LobbyConnection(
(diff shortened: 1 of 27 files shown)


server/__init__.py
```diff
@@ -103,10 +104,20 @@ def report_dirties():
             player_service.clear_dirty()
 
             if len(dirty_queues) > 0:
-                ctx.broadcast_raw(encode_queues(dirty_queues), lambda lobby_conn: lobby_conn.authenticated)
+                asyncio.ensure_future(
+                    ctx.broadcast_raw(
+                        encode_queues(dirty_queues),
+                        lambda lobby_conn: lobby_conn.authenticated
+                    )
+                )
```
Wesmania: This will cause `report_dirties` to not wait for these broadcasts. Worst case scenario, we'll start doing these in parallel. We need to change `report_dirties` to a coroutine as well. Also an opportunity to make its infinite loop sane.

server/__init__.py
```diff
@@ -134,7 +150,7 @@ def report_dirties():
     ping_msg = encode_message('PING')
 
     def ping_broadcast():
-        ctx.broadcast_raw(ping_msg)
+        asyncio.ensure_future(ctx.broadcast_raw(ping_msg))
```
Wesmania: As above, won't wait for this, need to be made a coro.

server/gameconnection.py
```diff
@@ -513,7 +517,7 @@ def abort(self, log_message: str=''):
             self._logger.debug("%s.abort(%s)", self, log_message)
 
             if self.game.state == GameState.LOBBY:
-                self.disconnect_all_peers()
+                asyncio.ensure_future(self.disconnect_all_peers())
```
Wesmania: As above.

server/__init__.py
Askaholic: I just took a look at `aiocron` and this is pretty much exactly how it is implemented. It has a synchronous function which fires off a coroutine without waiting for it to complete, and then schedules its self to be called again via `loop.call_at`. I tried working on refactoring it to use more or less of a sleep loop, but then you need to have a way of signaling the loop to shut down when the server shuts down, which is really annoying to do. With the call_at method, you don't need a shutdown signal, because if the server shuts down then the coroutine just won't be called again.

server/__init__.py
```diff
@@ -103,39 +107,54 @@ def report_dirties():
...
-                    validation_func = lambda lobby_conn: lobby_conn.player.id not in game.host.foes
+                    def validation_func(lobby_conn):
+                        return lobby_conn.player.id not in game.host.foes
+
+                tasks.append(ctx.broadcast_raw(
+                    message,
+                    lambda lobby_conn: lobby_conn.authenticated and validation_func(lobby_conn)
+                ))
+
+            await asyncio.gather(*tasks)
```
Wesmania: Is it safe to run all these broadcasts in parallel? Is it possible for data they send to get interleaved?

Askaholic: Well you have to remember that with asyncio, nothing actually runs in parallel. There is no scheduler that will interrupt a coroutine, so once a coroutine starts, it keeps running until it decides it’s ready to stop. So I don’t imagine that it’s possible for any messages to become corrupted. I would also guess that they’ll likely execute in the order that they’re created, although it wouldn’t be guaranteed.

Wesmania: Only chunks of code that don't contain any await calls are guaranteed to be atomic, so if we sent any command in chunks using multiple await calls, then they could get interleaved. It does seem that we're only writing whole commands, though.

Askaholic: Yea, I don't think the code does that anywhere. The `QDataStreamProtocol.send_message` function always encodes an entire message and writes it with a single call to `StreamWriter.write`

Issue #499, opened 2019-10-20
Game close message wording


Katharsas:
Close game message
![close](https://user-images.githubusercontent.com/1922573/67159117-badccd80-f340-11e9-87e8-45709cbe6f0d.JPG)
should be "by a moderator" instead of "by an administrator"

Relevant code:
https://github.com/FAForever/server/blob/develop/server/lobbyconnection.py#L272

Askaholic:
We should move this message entirely to the client to take advantage of it's localization capabilities. Instead of sending English text let's send something like:
```
{
    "command": "close_lobby",
    "initiator": "PhilipJFry",
    ... any other fields that we need
}
```

See issue  #444.

PR #501, opened 2019-10-20
game closed message is 'by moderator', not 'adminstrator'

cleborys:
Simple enough: #499


Askaholic:
We really shouldn't be sending English text from the server in the first place. If someone is working on this in the client right now, it would be a great chance to replace this with a proper command.

Issue #444, opened 2019-06-06
Improve localization of server messages


Askaholic:
The server sends a number of messages as strings in English. It would be better to send only command representations or maybe localization keys. In some cases the messages might even be moved completely to the client and not sent by the server at all.

For instance in `LadderService.inform_player`, the server will send a lengthy explanation of the matchmaker to new players. This message could probably be moved entirely to the client, and not need a server trigger at all.

Brutus5000:
The player has no way to choose his language as of now.

Askaholic:
I don't think I explained my idea well enough. I don't mean that the server should "do localization" its self, I meant that it should not send as much English text, and rather send commands containing the information necessary for the client to localize those commands into an appropriate message.

Brutus5000:
Ah yes, that should have been obvious to me.

PR #502, opened 2019-10-24
Let client do localization for Close FA message

### LINKED ISSUE #499: Game close message wording

Askaholic:
Since client work is being done on this anyways, might as well do it right. See FAForever/downlords-faf-client#1472

Related to #444.
Closes #499.

### DIFF
--- a/server/lobbyconnection.py
+++ b/server/lobbyconnection.py
@@ -258,11 +258,10 @@ class LobbyConnection():
                 subject_id=subject_id,
             ))
 
-    def kick(self, initiator):
+    def kick(self):
         self.send({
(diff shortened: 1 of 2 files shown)


Geosearchef:
The notice message is not supposed to have an initiator field.

Askaholic:
Should I make it a new command? Like `administrative_action`?

Geosearchef:
I don't think it's needed.

PR #505, opened 2019-11-24
Issue/#444 localize authentication_failed message

Askaholic:
Related #444.

This is a small start on getting rid of English display strings from the server. Now that the client can run on Java 11, I can actually implement this in the client myself.

Closes #183.


Askaholic:
Waiting on https://github.com/FAForever/downlords-faf-client/pull/1512

PR #507, opened 2019-11-29
General code cleanup

eforgacs:
Removing unnecessary parentheses, test for membership should be "not in", typo fixes, indentation/spacing, changing from use of deprecated functions to non-deprecated functions, capitalization, fixing syntax of Bash scripts, line length, added "await" to coroutine that is not awaited.

### DIFF
--- a/scripts/pipenv_run_tests.sh
+++ b/scripts/pipenv_run_tests.sh
@@ -1,2 +1,2 @@
 #! /bin/bash
-docker exec -i faf-db mysql faf < test-data.sql && scripts/run_tests_with_coverage.sh $@
+docker exec -i faf-db mysql faf < test-data.sql && scripts/run_tests_with_coverage.sh "$@"
--- a/server/__init__.py
+++ b/server/__init__.py
@@ -81,14 +81,14 @@ def encode_queues(queues):
 
 
 def run_lobby_server(
-        address: (str, int),
-        database: FAFDatabase,
-        player_service: PlayerService,
-        games: GameService,
-        loop,
-        nts_client: Optional[TwilioNTS],
-        geoip_service: GeoIpService,
-        ladder_service: LadderService
+    address: (str, int),
+    database: FAFDatabase,
+    player_service: PlayerService,
+    games: GameService,
+    loop,
+    nts_client: Optional[TwilioNTS],
+    geoip_service: GeoIpService,
+    ladder_service: LadderService
 ) -> ServerContext:
     """
     Run the lobby server
--- a/server/__init__.py
+++ b/server/__init__.py
@@ -33,7 +33,6 @@ __contact__ = 'admin@faforever.com'
 __license__ = 'GPLv3'
 __copyright__ = 'Copyright (c) 2011-2015 ' + __author__
 
-
 __all__ = (
     'GameConnection',
     'GameStatsService',
@@ -52,6 +51,7 @@ stats = None
 
 if not config.ENABLE_STATSD:
     from . import fake_statsd
+
     stats = fake_statsd.DummyConnection()
 else:
     stats = aiomeasures.StatsD(config.STATSD_SERVER)
@@ -81,14 +81,14 @@ def encode_queues(queues):
 
 
 def run_lobby_server(
-    address: (str, int),
-    database: FAFDatabase,
-    player_service: PlayerService,
-    games: GameService,
-    loop,
-    nts_client: Optional[TwilioNTS],
-    geoip_service: GeoIpService,
-    ladder_service: LadderService
+        address: (str, int),
+        database: FAFDatabase,
+        player_service: PlayerService,
+        games: GameService,
+        loop,
+        nts_client: Optional[TwilioNTS],
+        geoip_service: GeoIpService,
+        ladder_service: LadderService
 ) -> ServerContext:
     """
     Run the lobby server
@@ -121,7 +121,8 @@ def run_lobby_server(
                 # allowed to see them.
                 if game.visibility == VisibilityState.FRIENDS:
                     # To see this game, you must have an authenticated connection and be a friend of the host, or the host.
-                    validation_func = lambda lobby_conn: lobby_conn.player.id in game.host.friends or lobby_conn.player == game.host
+                    validation_func = lambda \
+                        lobby_conn: lobby_conn.player.id in game.host.friends or lobby_conn.player == game.host
                 else:
                     validation_func = lambda lobby_conn: lobby_conn.player.id not in game.host.foes
 
@@ -146,6 +147,7 @@ def run_lobby_server(
             players=player_service,
             ladder_service=ladder_service
         )
+
     ctx = ServerContext(make_connection, name="LobbyServer")
     loop.call_later(DIRTY_REPORT_INTERVAL, report_dirties)
     loop.call_soon(ping_broadcast)
--- a/server/gameconnection.py
+++ b/server/gameconnection.py
@@ -1,7 +1,7 @@
 import asyncio
 
 from server.db import FAFDatabase
-from sqlalchemy import text, select
+from sqlalchemy import text, select, or_
 
 from .abc.base_game import GameConnectionState
 from .config import TRACE
@@ -311,8 +311,8 @@ class GameConnection(GpgNetServerProtocol):
             """
 
             check = await conn.execute(select([login.c.id]).where(
-                login.c.id == teamkiller_id or
-                login.c.login == teamkiller_name
+                or_(login.c.id == teamkiller_id,
+                login.c.login == teamkiller_name)
             ))
 
             row = await check.fetchone()
@@ -331,8 +331,8 @@ class GameConnection(GpgNetServerProtocol):
             """
 
             check = await conn.execute(select([login.c.id]).where(
-                login.c.id == reporter_id or
-                login.c.login == reporter_name
+                or_(login.c.id == reporter_id,
+                login.c.login == reporter_name)
             ))
 
             row = await check.fetchone()
--- a/server/gameconnection.py
+++ b/server/gameconnection.py
@@ -305,10 +305,9 @@ class GameConnection(GpgNetServerProtocol):
         """
                 
         async with self._db.acquire() as conn:
-            """
-                Sometime the game sends a wrong ID - but a correct player name
-                We need to make sure the player ID is correct before pursuing
-            """
+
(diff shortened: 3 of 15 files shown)


tests/unit_tests/test_decorators.py
```diff
@@ -7,7 +7,7 @@
 
 def test_with_logger():
     @with_logger
-    class testclass():
+    class testclass:
```
norraxx: Can be camel case

scripts/pipenv_run_tests.sh
```diff
@@ -1,2 +1,2 @@
 #! /bin/bash
-docker exec -i faf-db mysql faf < test-data.sql && scripts/run_tests_with_coverage.sh $@
+docker exec -i faf-db mysql faf < test-data.sql && scripts/run_tests_with_coverage.sh "$@"
```
Askaholic: Are you sure this will still work? The point is to forward all command line arguments to pytest so that for example you can do `pipenv run tests tests/integration_tests -s --cli-log-level 5`. I would think that adding quotes would mash all of your arguments into a single string, which presumably doesn't get parsed correctly.

tests/unit_tests/test_decorators.py
eforgacs: Thanks! Just changed it.

scripts/pipenv_run_tests.sh
eforgacs: Here's the warning from ShellCheck that I get when I leave it in its current state:

"Double quote array expansions to avoid re-splitting elements."

> Rationale:
> Double quotes around $@ (and similarly, ${array[@]}) prevents globbing and word splitting of individual elements, while still expanding to multiple separate arguments.
> 
> Let's say you have three arguments: baz, foo bar and *
> 
> "$@" will expand into exactly that: baz, foo bar and *
> 
> $@ will expand into multiple other arguments: baz, foo, bar, file.txt and otherfile.jpg
> 
> Since the latter is rarely expected or desired, ShellCheck warns about it.

https://github.com/koalaman/shellcheck/wiki/SC2068

eforgacs:
@Askaholic I had never heard about `black` until now - is this the right link?

https://black.readthedocs.io/en/stable/

I really like the description! Can't wait to try it out!

server/__init__.py
```diff
@@ -121,7 +121,8 @@ def report_dirties():
                 # allowed to see them.
                 if game.visibility == VisibilityState.FRIENDS:
                     # To see this game, you must have an authenticated connection and be a friend of the host, or the host.
-                    validation_func = lambda lobby_conn: lobby_conn.player.id in game.host.friends or lobby_conn.player == game.host
+                    validation_func = lambda \
+                        lobby_conn: lobby_conn.player.id in game.host.friends or lobby_conn.player == game.host
```
norraxx: Imho, you should leave first lambda param on same line as lambda. How do you read code? I read first input params then logic

norraxx:
+1 for black, but if we will use black, we need to reformat whole project.

scripts/pipenv_run_tests.sh
Askaholic: I just tried it on my machine, and it seems to still work so I guess there's no problem.

server/__init__.py
```diff
@@ -81,14 +81,14 @@ def encode_queues(queues):
...
 def run_lobby_server(
-    address: (str, int),
-    database: FAFDatabase,
-    player_service: PlayerService,
-    games: GameService,
-    loop,
-    nts_client: Optional[TwilioNTS],
-    geoip_service: GeoIpService,
-    ladder_service: LadderService
+        address: (str, int),
```
Askaholic: What's with the double indents? I prefer single, and for what it's worth so does `black`

eforgacs: That's a good question. PyCharm formatted it that way, I think it's just conforming to PEP8, but I'm amenable to either single or double, no strong feelings one way or the other.

server/__init__.py
eforgacs: That's a good suggestion. I think this was a "line too long" formatting change, so I can change this to be on one line.

server/__init__.py
Askaholic: Ok, let's keep it single then since that's what I've been doing in the rest of the code.

server/__init__.py
Askaholic: How about just defining the function with a `def`? I don't see the point of using a lambda function here anyways.

tests/unit_tests/test_matchmaker_queue.py
```diff
@@ -227,7 +227,7 @@ def matchmaker_players_all_match(player_factory):
     matchmaker_queue._matches = deque(matches)
 
     async def call_shutdown():
-        asyncio.sleep(1)
+        await asyncio.sleep(1)
```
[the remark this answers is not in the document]
Askaholic: I guess not. TBH I'm not sure any more why I was calling shutdown like that.

server/gameconnection.py
```diff
@@ -129,24 +133,25 @@ def send_message(self, message):
...
         :return: None
         """
-        self.send_ConnectToPeer(player_name=peer.player.login,
-                                player_uid=peer.player.id,
-                                offer=True)
-        peer.send_ConnectToPeer(player_name=self.player.login,
-                                player_uid=self.player.id,
-                                offer=False)
+        self.send_ConnectToPeer(
+            player_name=peer.player.login, player_uid=peer.player.id, offer=True
```
norraxx: once you have it on one line, another on 3. decide how to format, dude ;-)

server/gameconnection.py
```diff
@@ -246,20 +257,33 @@ def send_message(self, message):
...
+                [
+                    result.startswith(possible_result)
+                    for possible_result in [
+                        "score",
+                        "defeat",
+                        "victory",
+                        "draw",
+                    ]
+                ]
+            ):
```
norraxx: well, the code, that was before was valid and nice, rewritting loop like that makes me unhappy. in fact, you want code to be compact, because of reading.

server/gameconnection.py
```diff
@@ -303,23 +343,26 @@ def send_message(self, message):
...
             """
 
-            check = await conn.execute(select([login.c.id]).where(
-                login.c.id == teamkiller_id or
-                login.c.login == teamkiller_name
-            ))
+            check = await conn.execute(
+                select([login.c.id]).where(
+                    login.c.id == teamkiller_id
+                    or login.c.login == teamkiller_name
```
norraxx: this is very invalid query., please, if you are touching it, fix it. or condition in sqlalchemy is `or_`

norraxx: see more https://stackoverflow.com/questions/7942547/using-or-in-sqlalchemy

server/gameconnection.py
```diff
@@ -330,17 +373,19 @@ def send_message(self, message):
                 for the same reasons
             """
 
-            check = await conn.execute(select([login.c.id]).where(
-                login.c.id == reporter_id or
-                login.c.login == reporter_name
-            ))
+            check = await conn.execute(
+                select([login.c.id]).where(
+                    login.c.id == reporter_id or login.c.login == reporter_name
```
norraxx: same here

norraxx: and decide how to format ;-)

server/gameconnection.py
```diff
@@ -399,14 +444,14 @@ def send_message(self, message):
...
             )
             return
 
-        game_connection.send_message({
-            "command": "IceMsg",
-            "args": [int(self.player.id), ice_msg]
-        })
+        game_connection.send_message(
+            {"command": "IceMsg", "args": [int(self.player.id), ice_msg]}
+        )
```
norraxx: again, don't agree with formatting, before it was - each line is one value, now it's unreadable.

server/matchmaker/algorithm.py
```diff
@@ -171,13 +182,15 @@ def build_sparse(searches: List[Search]) -> Dict[Search, List[Search]]:
...
+                key=lambda other: search.quality_with(other),
             )
             for search in searches
         }
 
     @staticmethod
-    def _get_top_matches(search: Search, others: Iterable[Search]) -> List[Search]:
+    def _get_top_matches(
+        search: Search, others: Iterable[Search]
+    ) -> List[Search]:
```
norraxx: what's wrong with formatting before?

server/matchmaker/matchmaker_queue.py
```diff
@@ -113,7 +113,9 @@ def match(self, s1: Search, s2: Search) -> bool:
         :param s2:
         :return:
         """
-        if (s1.is_matched or s2.is_matched) or (s1.is_cancelled or s2.is_cancelled):
+        if (s1.is_matched or s2.is_matched) or (
+            s1.is_cancelled or s2.is_cancelled
+        ):
```
norraxx: I don't understand what algorithm do you use for such changes?

server/matchmaker/matchmaker_queue.py
```diff
@@ -95,8 +96,7 @@ def find_matches(self) -> None:
 
         # Call self.match on all matches and filter out the ones that were canceled
         new_matches = filter(
-            lambda m: self.match(m[0], m[1]),
-            make_matches(self.queue.values())
+            lambda m: self.match(m[0], m[1]), make_matches(self.queue.values())
```
norraxx: solution before was defenitwly better

Askaholic:
@norraxx this formatting comes from black. I’m not sure I really like it either.

Also since I’m the person who spends the most time working on this code base at the moment, I feel that it is fully within reason for me to require a max line length of 80 because my screen is not as big as yours, and I prefer not to squint or constantly scroll to read full lines when I’m coding.

server/gameconnection.py
Askaholic: Good catch we should definitely fix that

norraxx:
well, at my job, we are up to migrate to black... we do have code changes, so we are in same situation now, and we're not so happy :-)))
80 width is rough restriction from 30 years ago

eforgacs:
Hey guys, sorry about the mess. I'm thinking about undoing this commit entirely and starting over (without using black). What are your thoughts?

norraxx:
If you want, make `git reset <hash of first non yours commit>`

eforgacs:
> If you want, make `git reset <hash of first non yours commit>`, then git commit and git push force

Yes, that's actually what @Askaholic recommended as well. I'll give that a shot.

PR #508, opened 2019-11-30
Register UID even if user is steam linked

Rackover:


server/lobbyconnection.py
```diff
@@ -525,9 +525,9 @@ def check_version(self, message):
                     "player_id": player_id
                 })
 
-            if not self.player_service.is_uniqueid_exempt(player_id) and steamid is None:
```
Geosearchef: That won't work. `check_policy_conformity` sends an error to the user when the system is already registered, meaning steam bypass won't work anymore, this method will continue running, but the user gets an error anyways.

Rackover: Ok let's try this way

Rackover:
Welp :p And now, to update tests..

Askaholic:
Superseded by #510.

PR #509, opened 2019-12-15
Migrate to new permission system

Brutus5000:
The legacy access level is already removed from the API. We need to check for the group permission.

New permissions:
- [x] `ADMIN_KICK_SERVER`
- [x] `ADMIN_BROADCAST_MESSAGE`
- [x] `ADMIN_JOIN_CHANNEL`


axel1200:
https://github.com/FAForever/db/pull/211

server/db/models.py
```diff
@@ -298,3 +316,23 @@
     Column('game_id',       Integer, ForeignKey('game_stats.id')),
     Column('gametime',      Integer),
 )
+
+user_group = Table(
+    'user_group', metadata,
+    Column('id',                Integer,        primary_key=True),
+    Column('technical_name',    String(100),    nullable=False),
```
cleborys, line 323: Here `technical name` and `name_key` are shorter than before (`String(100)` vs `String(255)`).
That's probably long enough either way, but I wonder why they are different, given that they probably refer to keys in the same table of descriptions?

server/lobbyconnection.py
```diff
@@ -287,8 +287,8 @@ async def send_updated_achievements(self, updated_achievements):
     async def command_admin(self, message):
         action = message['action']
 
-        if self.player.admin:
-            if action == "closeFA":
+        if action == "closeFA":
+            if await self.player_service.has_permission_role(self.player, 'ADMIN_KICK_SERVER'):
```
cleborys, line 291: You could think about moving these in the same condition i.e.
```python
if (
  action == "closeFA"
  and await self.player_service.has_permission_role(self.player, 'ADMIN_KICK_SERVER')
):
```
upside: Loses one indentation level
downside: A few more `elif` checks if it fails because the permissions are denied.
It will short-circuit though, so only ever one `has_permission_role` is awaited.

server/player_service.py
```diff
@@ -64,6 +64,15 @@ def clear_dirty(self):
 
     async def fetch_player_data(self, player):
         async with self._db.acquire() as conn:
+            result = await conn.execute(
```
cleborys, line 67: Could this be added to the query below (`sql = ...`)?

server/player_service.py
```diff
@@ -167,20 +176,37 @@ async def _fetch_player_legacy_rating(self, player, rating_types, conn):
...
 
-
     def remove_player(self, player: Player):
         if player.id in self._players:
             del self._players[player.id]
             metrics.players_online.set(len(self._players))
 
-    def get_permission_group(self, user_id: int) -> int:
-        return self.privileged_users.get(user_id, 0)
+    async def has_permission_role(self, player: Player, role_name: str) -> bool:
```
cleborys, line 189: Given that the `user_groups` are already pre-fetched and stored in the player object, maybe this could be quicker / work with only one join?

As far as I can see right now, all `user_groups` are loaded on `fetch_player_data`, but then only used to check if the player is a moderator (in the `mod` property), while checking any other permission uses this query?

server/players.py
```diff
@@ -76,6 +74,10 @@ def __init__(
         self._game = lambda: None
         self._game_connection = lambda: None
 
+    @property
+    def mod(self):
```
cleborys, line 78: Maybe `is_moderator` (or even `is_moderator()` without `property`), since "mod" could also mean a "game mod"?

tests/unit_tests/test_lobbyconnection.py
```diff
@@ -501,194 +500,20 @@ async def test_command_admin_closelobby(mocker, lobbyconnection):
...
-
-    async with database.acquire() as conn:
-        result = await conn.execute(select([ban]).where(ban.c.player_id == banme.id))
-
-        bans = [row['reason'] async for row in result]
-
-    assert len(bans) == 0
-
-
-async def test_command_admin_closelobby_with_ban_injection(mocker, lobbyconnection, database):
```
cleborys, line 647: Do I even want to know why this test was necessary at some point? :sweat_smile:

Brutus5000:
> As far as I can see it was previously possible to include a "ban" in a "kick" action and that is no longer the case? I assume there is an alternative way to ban people, since kicking someone is otherwise merely forcing them to log out and back inagain?

Banning is a feature of the API.

server/db/models.py
Askaholic: Don't know, that's just how it is in the database :P

Askaholic: Actually I should fix the `group_permission` ones. Those should be 100 as well I think.

server/player_service.py
Askaholic: Yea I can if I also grab the group id's.

tests/unit_tests/test_lobbyconnection.py
Askaholic: You can look at the deleted code, it was using string formatting in the `text` call because mysql has no way of parametrizing interval durations.

server/player_service.py
Askaholic: Not very easily because both queries return a list. In order to get both lists in a single query I'm pretty sure you have to return the cartesian product of both lists which doesn't scale well at all.

cleborys: I see

server/player_service.py
```diff
@@ -64,6 +64,18 @@ def clear_dirty(self):
...
+            result = await conn.execute(
+                select([
+                    user_group.c.technical_name,
+                    user_group.c.id
+                ])
+                .select_from(user_group_assignment.join(user_group))
+                .where(user_group_assignment.c.user_id == player.id)
+            )
+            player.user_groups = {
+                entry async for row in result for entry in row.as_tuple()
```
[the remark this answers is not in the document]
norraxx: I don't get that line, why we have to call for each entry as `row.as_tuple()?` when result row is always named tuple?
and how is it possible, to have loop in loop. Like, when i read the result from sqlalchemy, i will get result row as named tuple and it's it, no need to iterate. And when i wanna have the values, i'll just write something like:
```
player.user_groups {(row.technical_name, row.id) for row in result}
```
Also I don't get, how do you wanna work with that set? Do you flatten whole result to `player.user_groups`? Like there will be the `{technical_name, id, technical_name2, id2}`? And is there any reason why? Because only place where you're using it is in `if row.group_id in player.user_groups`, which could be solved on database side, like just query database is there is result with proper tables join.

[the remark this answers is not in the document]
norraxx: To be continued in 5-6 hours

server/player_service.py
```diff
@@ -167,20 +179,33 @@ async def _fetch_player_legacy_rating(self, player, rating_types, conn):
...
+                        group_permission.c.id ==
+                        group_permission_assignment.c.permission_id
+                    ))
+                )
+                .where(group_permission.c.technical_name == role_name)
+            )
+            async for row in result:
+                if row.group_id in player.user_groups:
+                    return True
+            return False
```
norraxx, line 208: As I mentioned before, you don't need to have `player.user_groups`, because you're making another query to the database here. just join proper tables and get rid of `player.user_groups`. It's not hard for MySQL to have one more join.

What you will check is only that you have a row, so you can select `text(1).label("right_exists"). .... .first()`.

In the end, fewer queries to the database, and less code to maintain.

server/player_service.py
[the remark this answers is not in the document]
norraxx: I've found peace of code that is working with `player.user_groups` below. And I'm thinking that you can leave that `player.user_groups` variable.

server/lobbyconnection.py
```diff
@@ -287,8 +287,8 @@ async def send_updated_achievements(self, updated_achievements):
     async def command_admin(self, message):
         action = message['action']
 
-        if self.player.admin:
-            if action == "closeFA":
+        if action == "closeFA":
```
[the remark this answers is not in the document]
Askaholic: I'm not going to make the actions constants for now. We may find a better solution for this when we work on refactoring the protocol for auto documentation. But I'm not going to do it in this PR.

As for the permissions. I'm not really sure. Seems a bit overkill since they're only used in this one function.

server/players.py
```diff
@@ -134,6 +132,19 @@ def game_connection(self, value):
...
 
+    def power(self):
+        """An artifact of the old permission system. The client still uses this
+        number to determine if a player gets a special category in the user list
+        such as "Moderator"
+        """
+        if self.is_moderator():
+            return 1
+
+        return 0
```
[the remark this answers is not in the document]
Askaholic: Might as well just do `int(self.is_moderator())` at that point. I think I need to fix this to check for server admin too though.

server/player_service.py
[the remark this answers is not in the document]
Askaholic: Well it's not just a named tuple. It's a `RowProxy` object from sqlalchemy which does a bunch of stuff. I'm pretty sure just iterating it did not work so I had to call `as_tulpe()`.

> Do you flatten whole result to player.user_groups

yes exactly. I'm grabbing ids and technical_names so that I can use either to check permissions. This lets me write the permission query using only 1 join instead of 2.

server/player_service.py
Askaholic: It's still the same number of queries because we need the technical_names for `Player.is_moderator()`. It's just that the way I have it, we use 1 fewer join.

server/lobbyconnection.py
```diff
@@ -298,64 +298,19 @@ async def command_admin(self, message):
...
-                            await player.lobby_connection.kick()
-                    if ban_fail:
-                        raise ClientError("Kicked the player, but he was already banned!")
-
-            elif action == "broadcast":
-                message_text = message.get('message')
-                if not message_text:
-                    return
+                if player and player.lobby_connection is not None:
+                    self._logger.warning('Administrative action: %s closed client for %s', self.player, player)
```
axel1200, line 305: Not sure what your logging policy is but I'd say that's info

Askaholic: True

server/player_service.py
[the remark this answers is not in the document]
norraxx: Ok.

server/player_service.py
norraxx: yeah, np.

server/player_service.py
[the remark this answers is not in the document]
Askaholic: Btw I reverted this so it uses 2 joins but doesn’t pull as much data from the db each time.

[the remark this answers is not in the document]
Askaholic: Oh maybe this part was in a different commit. I’ll need to check again

PR #512, opened 2019-12-28
Add a login throttler

Wesmania:
Inspired by talks on Zulip, this should help with large login spikes thrashing the server irrespective of client behaviour.

TODO: add integration tests.


server/throttler.py
```diff
@@ -0,0 +1,34 @@
...
+        if self._quota < 0:
+            async with self._wait_lock:
+                await self._wait_lock.wait()
+
+    async def _leak(self):
+        while True:
+            self._quota += self._rate_per_s
+            self._quota = min(self._quota, self._capacity)
+            async with self._wait_lock:
+                self._wait_lock.notify(self._rate_per_s)
```
cleborys: Just checking i understand this correctly:
At the start of every second, the quota is increased by `rate_per_s`, and the first `rate_per_s` waiting login requests are released, so there will be `rate_per_s` logins granted at the start of the second and  an additional `rate_per_s` logins granted for new incoming requests during the second?
Then the effective rate would be up to `2 * rate_per_s`?

Intuitively, if you stick with this implementation I would have increased `self._quota` only by `rate_per_s` minus the number of requests being notified.

It's nice that this doesn't need any retries!
I cannot say if it's a drawback to release all of them at once at the start of a second though.

server/throttler.py
```diff
@@ -0,0 +1,34 @@
...
+            await asyncio.sleep(1)
+
+    async def close(self):
+        self._leak_coro.cancel()
+        self._closing = True
+        async with self._wait_lock:
+            self._wait_lock.notify_all()
```
cleborys: Does this give a load spike when closing the server?

server/config.py
```diff
@@ -71,3 +71,6 @@
 QUEUE_POP_DESIRED_PLAYERS = int(os.getenv('QUEUE_POP_DESIRED_PLAYERS', 8))
 # How many previous queue sizes to consider
 QUEUE_POP_TIME_MOVING_AVG_SIZE = int(os.getenv('QUEUE_POP_TIME_MOVING_AVG_SIZE', 5))
+
+LOGIN_THROTTLER_CAPACITY = int(os.getenv('LOGIN_THROTTLER_CAPACITY', -1))
+LOGIN_THROTTLER_RATE_PER_S = int(os.getenv('LOGIN_THROTTLER_RATE_PER_S', -1))
```
cleborys: What happens if I set `LOGIN_THROTTLER_CAPACITY` but forget to set `LOGIN_THROTTLER_RATE_PER_S`?

Maybe this should have a reasonable positive default value?

server/servercontext.py
```diff
@@ -12,10 +13,11 @@ class ServerContext:
     Base class for managing connections and holding state about them.
     """
 
-    def __init__(self, connection_factory, name='Unknown server'):
+    def __init__(self, connection_factory, throttler, name='Unknown server'):
```
cleborys: Maybe `throttler=None` could be the default, then you don't need to explicitly pass `None` in the integration tests.

server/throttler.py
Wesmania: The whole thing implements this:
https://en.wikipedia.org/wiki/Leaky_bucket

server/throttler.py
Wesmania: Not if the server is under load. I dunno if we care.

Wesmania:
That PR is a draft really, and we don't need it after server migration. Unless we start having server issues again, it'll probably stay unmerged.

cleborys:
Thank you for your comments, anyways :)

server/config.py
norraxx: I would use full name `LOGIN_THROTTLER_RATE_PER_SECOND`

Askaholic:
Haven't seen a need for this

Brutus5000:
Additional side note: We have login throttling on acquiring an OAuth token now.

PR #514, opened 2020-01-07
Issue/#513 New geoip database download URL

### LINKED ISSUE #513: Download GeoIP database with Maxmind account token
Maxmind recently pulled their public download link because of CCPA: https://blog.maxmind.com/2019/12/18/significant-changes-to-accessing-and-using-geolite2-databases/

We need to:
1. [x] Create a maxmind account for FAF
2. [x] Add the license key as a config parameter
3. [x] Change the url and add license key as a get parameter

Askaholic:
Updated the geoip url and added a license key parameter. The license key does not have a default and must be set by a server admin in order for the new database to be downloaded.

The database is now served via a `tar.gz` which contains a copyright and a license text as well as the database. The server now downloads the archive and compares its checksum against the md5 hash available at the same url. Once the archive has been successfully downloaded, the database file is extracted from the archive and copied over the old version of the file (if it exists).

Closes #513

### DIFF
--- a/server/geoip_service.py
+++ b/server/geoip_service.py
@@ -1,7 +1,8 @@
 import asyncio
-import gzip
+import hashlib
 import os
 import shutil
+import tarfile
 from datetime import datetime
 
 import aiocron
@@ -69,23 +70,45 @@ class GeoIpService(object):
             Download the geoip database to a file. If the downloaded file is not
         a valid gzip file, then it does NOT overwrite the old file.
         """
+        if not config.GEO_IP_LICENSE_KEY:
+            self._logger.warning(
+                "GEO_IP_LICENSE_KEY not set! Unable to download GeoIP database!"
+            )
+            return
 
         self._logger.info("Downloading new geoip database")
 
         # Download new file to a temp location
-        temp_file_path = "/tmp/geoip.mmdb.gz"
-        await self._download_file(config.GEO_IP_DATABASE_URL, temp_file_path)
+        temp_file_path = "/tmp/geoip.mmdb.tar.gz"
+        await self._download_file(
+            config.GEO_IP_DATABASE_URL,
+            config.GEO_IP_LICENSE_KEY,
+            temp_file_path
+        )
+
+        def get_db_member(tar: tarfile.TarFile) -> str:
+            for name in tar.getnames():
+                if name.endswith("GeoLite2-Country.mmdb"):
+                    return name
+
+            # Because we verified the checksum earlier, this should only be
+            # possible if maxmind actually served us a bad file
+            raise Exception("Tar archive did not contain the database file!")
 
         # Unzip the archive and overwrite the old file
         try:
-            with gzip.open(temp_file_path, 'rb') as f_in:
+            with tarfile.open(temp_file_path, 'r:gz') as tar:
+                name = get_db_member(tar)
                 with open(self.file_path, 'wb') as f_out:
+                    f_in = tar.extractfile(name)
+                    assert f_in is not None
                     shutil.copyfileobj(f_in, f_out)
-        except OSError:    # pragma: no cover
-            self._logger.warning("Failed to unzip downloaded file!")
+        except (tarfile.TarError) as e:    # pragma: no cover
+            self._logger.warning("Failed to extract downloaded file!")
+            raise e
         self._logger.info("New database download complete")
 
-    async def _download_file(self, url: str, file_path: str) -> None:
+    async def _download_file(self, url: str, license_key: str, file_path: str) -> None:
         """
             Download a file using aiohttp and save it to a file.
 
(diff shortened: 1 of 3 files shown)


server/geoip_service.py
```diff
@@ -69,23 +70,45 @@ def __init__(self):
...
+                "GEO_IP_LICENSE_KEY not set! Unable to download GeoIP database!"
+            )
+            return
 
         self._logger.info("Downloading new geoip database")
 
         # Download new file to a temp location
-        temp_file_path = "/tmp/geoip.mmdb.gz"
-        await self._download_file(config.GEO_IP_DATABASE_URL, temp_file_path)
+        temp_file_path = "/tmp/geoip.mmdb.tar.gz"
```
norraxx: maybe use https://docs.python.org/3/library/tempfile.html ?

server/geoip_service.py
```diff
@@ -69,23 +70,45 @@ def __init__(self):
...
+        await self._download_file(
+            config.GEO_IP_DATABASE_URL,
+            config.GEO_IP_LICENSE_KEY,
+            temp_file_path
+        )
+
+        def get_db_member(tar: tarfile.TarFile) -> str:
```
norraxx: is there a reason for inner function? Like i mean, you don't use local variables from outer scope, etc...

server/geoip_service.py
```diff
@@ -94,14 +117,42 @@ def __init__(self):
...
+        async with aiohttp.ClientSession(raise_for_status=True) as session:
+            checksum, our_hash = await asyncio.gather(
+                get_checksum(session),
+                get_db_file_with_checksum(session)
+            )
+
+        if checksum != our_hash:
+            raise Exception(
+                f"Hashes did not match! Expected {checksum} got {our_hash}"
+            )
```
norraxx: shouldn't you log?

server/geoip_service.py
```diff
@@ -69,23 +70,45 @@ def __init__(self):
...
+            raise Exception("Tar archive did not contain the database file!")
 
         # Unzip the archive and overwrite the old file
         try:
-            with gzip.open(temp_file_path, 'rb') as f_in:
+            with tarfile.open(temp_file_path, 'r:gz') as tar:
+                name = get_db_member(tar)
                 with open(self.file_path, 'wb') as f_out:
+                    f_in = tar.extractfile(name)
+                    assert f_in is not None
```
norraxx: what's the reason behind of it?

norraxx: Like, you `get_db_member` raising if there is no file, isn't it?

server/geoip_service.py
Askaholic: Yes, it's for readability. If I put it inline it looks like this:
```python
with tarfile.open(temp_file_path, 'r:gz') as tar:
    for name in tar.getnames():
        if name.endswith("GeoLite2-Country.mmdb"):
            with open(self.file_path, 'wb') as f_out:
                f_in = tar.extractfile(name)
```
Which is too much indentation IMO

server/geoip_service.py
Askaholic: It will get logged by the exception handler

server/geoip_service.py
Askaholic: Because `tar.extractfile` has return type `Optional[IO[bytes]]` I get a type error (warning) if I use it directly since the type checker doesn't know about the relationship between `tar.getnames` and `tar.extractfile`. 

The assert should theoretically always succeed (unless there is a bug in `tarfile` I guess) but it makes the type checker happy.

server/geoip_service.py
Askaholic: I don't see a reason to. The purpose is just so it doesn't overwrite the existing file while downloading. There's no need to use a module just to add `/tmp` to the beginning of a file path :P

Askaholic:
I did think about whether to move the functions `get_db_file_with_checksum` and `get_checksum` somewhere else, but I think it makes more sense to leave them there. They aren't being used anywhere else in the code so I feel that it would just be annoying to read if you had to go find them in a different file. They are basically like lambda functions except that python doesn't allow async lambdas, so they have to be real functions (well coroutines actually).

The other one as I explained is for readability because I didn't want to have 5 levels of nested indentation.

norraxx:
How about to extract them from class and leave in this file?

server/geoip_service.py
norraxx: If something goes wrong, file could stay locked for a while. I just propose the tempfile.

server/geoip_service.py
norraxx: =^_^= couldn't see that without analyzer. Yes, i thought about wrong tar too, but i don't know, if there is such possibility, when we already could read name.. like it's highly non possible to not to have whole file, isn't it? :-)

Askaholic: Yea, like I said. I don't think it's possible for that assert to fail.

server/geoip_service.py
Askaholic: How would it get locked? It's opened in a context manager and this function is supposed to be called once every 3 weeks...

norraxx: Ok. NP.

server/geoip_service.py
Wesmania: We could also do `name = next((n for n in tar.getnames() if n.endswith(foo)), None)`, then throw if it's None.

server/geoip_service.py
Wesmania: Unless someone's feeling malicious and the tar has a directory with `GeoLite2-Country.mmdb` in it. IMHO it does not hurt to just throw.

server/geoip_service.py
Askaholic: Oh yea, that's pretty clean. There's also a function for getting `TarInfo` objects instead of just names which we can use to make sure that we only look at files and not directories as well.

PR #515, opened 2020-01-11
Await async sending functions in command_admin

Askaholic:
While merging a different PR I noticed that I forgot to await these and there were no tests to catch this error. So now I'm fixing it and adding some tests.

I also refactored the `join_channel` command. I noticed that the logic would not allow admins to use that command (only mods without admin power could use it) and that the method of extracting the user_id's seemed questionable. Before the message would have had to look like:
```json
{
    "user_ids": [1, 2, 3],
    1: 1,
    2: 2,
    3, 3
}
``` 
which seems insane to me and I'm not even sure is valid json, so now it just looks like:

```json
{
    "user_ids": [1, 2, 3]
}
```

### DIFF
--- a/tests/unit_tests/test_lobbyconnection.py
+++ b/tests/unit_tests/test_lobbyconnection.py
@@ -752,6 +752,50 @@ async def test_broadcast(lobbyconnection: LobbyConnection, mocker):
     tuna.lobby_connection.send_warning.assert_called_with("This is a test message")
 
 
+async def test_broadcast_during_disconnect(lobbyconnection: LobbyConnection, mocker):
+    player = mocker.patch.object(lobbyconnection, 'player')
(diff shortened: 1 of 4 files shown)


tests/unit_tests/test_lobbyconnection.py
```diff
@@ -752,6 +752,50 @@ def policy_server(event_loop):
...
+    lobbyconnection.player_service = [player, tuna]
+
+    # This should not leak any exceptions
+    await lobbyconnection.on_message_received({
+        'command': 'admin',
+        'action': 'broadcast',
+        'message': "This is a test message"
+    })
+
+    player.lobby_connection.send_warning.assert_called_with("This is a test message")
```
norraxx: :+1:

PR #517, opened 2020-01-18
Send game configuration options to the client in game_launch

Askaholic:
Pulling some stuff out of #428 that doesn't need to wait. I think these changes are backwards compatible with the current client. A client implementation exists on the gw-on-ice branch: https://github.com/FAForever/downlords-faf-client/commit/97f72147d48537ca98a7cc37faad591f612fc6b8

### DIFF
--- a/server/lobbyconnection.py
+++ b/server/lobbyconnection.py
@@ -844,6 +844,11 @@ class LobbyConnection:
             "uid": game.id,
             "mod": game.game_mode,
             "mapname": use_map,
+            # Following parameters are not used by the client yet. They are
+            # needed for setting up auto-lobby style matches such as ladder, gw,
+            # and team machmaking where the server decides what these game
+            # options are. Currently, options for ladder are hardcoded into the
+            # client.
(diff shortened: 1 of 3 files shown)


server/lobbyconnection.py
```diff
@@ -840,13 +840,20 @@ def on_connection_made(self, protocol: QDataStreamProtocol, peername: Address):
...
             "uid": game.id,
-            "args": ["/numgames " + str(self.player.game_count[RatingType.GLOBAL])]
+            "mod": game.game_mode,
+            "mapname": use_map,
+            "name": game.name,
+            "team": game.get_player_option(self.player, "Team"),
+            "faction": game.get_player_option(self.player, "Faction"),
+            "expected_players": len(game.players) or None,
+            "map_position": game.get_player_option(self.player, "StartSpot"),
+            "init_mode": game.init_mode.value,
```
Geosearchef: I'd add a comment that those are not used yet so nobody is confused why changing them doesn't do anything.

PR #524, opened 2020-01-30
Bugfix: Handle connection errors

Askaholic:
`broadcast_raw` is now guaranteed not to raise any `ConnectionError`s. We really don't care if our broadcast failed to a player because they closed their connection before we finished writing the message. 

Also ignored connection errors in some other places where a manual broadcast was happening such as admin commands.

### DIFF
--- a/server/async_functions.py
+++ b/server/async_functions.py
@@ -2,26 +2,18 @@
 Some helper functions for common async tasks.
 """
 import asyncio
-from typing import Callable, List, Optional
+from typing import List
 
 
 async def gather_without_exceptions(
     tasks: List[asyncio.Task],
     *exc_classes: type,
-    callback: Optional[Callable[[Exception], None]] = None
 ) -> None:
     """
     Call gather on a list of tasks, raising the first exception that dosen't
-    match a any of the specified exception classes.
-
-    If callback is set, then it will be called on any non-matching exceptions.
-    This is useful if you need to process every exception and not just the
-    first one.
+    match any of the specified exception classes.
     """
     results = await asyncio.gather(*tasks, return_exceptions=True)
     for result in results:
         if isinstance(result, Exception) and result.__class__ in exc_classes:
-            if callback is None:
-                raise result
-            else:
-                callback(result)
+            raise result
--- a/server/async_functions.py
+++ b/server/async_functions.py
@@ -2,18 +2,23 @@
 Some helper functions for common async tasks.
 """
 import asyncio
-from typing import List
+from typing import Any, List
 
 
 async def gather_without_exceptions(
     tasks: List[asyncio.Task],
-    *exc_classes: type,
-) -> None:
+    *exceptions: type,
+) -> List[Any]:
     """
     Call gather on a list of tasks, raising the first exception that dosen't
     match any of the specified exception classes.
     """
     results = await asyncio.gather(*tasks, return_exceptions=True)
     for result in results:
-        if isinstance(result, Exception) and result.__class__ not in exc_classes:
-            raise result
+        if isinstance(result, Exception):
+            # Check if this exception is an instance (maybe subclass) that
+            # should be ignored
+            for exc_type in exceptions:
+                if not isinstance(result, exc_type):
+                    raise result
+    return results
--- a/server/async_functions.py
+++ b/server/async_functions.py
@@ -1,0 +1,27 @@
+"""
+Some helper functions for common async tasks.
+"""
+import asyncio
+from typing import Callable, List, Optional
+
+
+async def gather_without_exceptions(
+    tasks: List[asyncio.Task],
+    *exc_classes: type,
+    callback: Optional[Callable[[Exception], None]] = None
+) -> None:
+    """
+    Call gather on a list of tasks, raising the first exception that dosen't
+    match a any of the specified exception classes.
+
+    If callback is set, then it will be called on any non-matching exceptions.
+    This is useful if you need to process every exception and not just the
+    first one.
+    """
+    results = await asyncio.gather(*tasks, return_exceptions=True)
+    for result in results:
+        if isinstance(result, Exception) and result.__class__ in exc_classes:
+            if callback is None:
+                raise result
+            else:
+                callback(result)
--- a/server/async_functions.py
+++ b/server/async_functions.py
@@ -15,5 +15,5 @@ async def gather_without_exceptions(
     """
     results = await asyncio.gather(*tasks, return_exceptions=True)
     for result in results:
-        if isinstance(result, Exception) and result.__class__ in exc_classes:
+        if isinstance(result, Exception) and result.__class__ not in exc_classes:
             raise result
--- a/server/gameconnection.py
+++ b/server/gameconnection.py
@@ -536,13 +536,7 @@ class GameConnection(GpgNetServerProtocol):
             if peer == self:
                 continue
 
-            try:
-                tasks.append(peer.send_DisconnectFromPeer(self.player.id))
-            except AttributeError:
-                # In case player.lobby_connection was None
-                self._logger.debug("Failed to send DisconnectFromPeer to %s", peer)
-            except Exception:
-                self._logger.exception("Failed to send DisconnectFromPeer to %s", peer)
+            tasks.append(peer.send_DisconnectFromPeer(self.player.id))
 
         for fut in asyncio.as_completed(tasks):
             try:
--- a/server/gameconnection.py
+++ b/server/gameconnection.py
@@ -4,6 +4,7 @@ from server.db import FAFDatabase
 from sqlalchemy import or_, select, text
 
 from .abc.base_game import GameConnectionState
+from .async_functions import gather_without_exceptions
 from .config import TRACE
 from .db.models import login, moderation_report, reported_user
 from .decorators import with_logger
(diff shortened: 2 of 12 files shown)


server/async_functions.py
```diff
@@ -0,0 +1,27 @@
+"""
+Some helper functions for common async tasks.
+"""
+import asyncio
+from typing import Callable, List, Optional
+
+
+async def gather_without_exceptions(
```
Wesmania: If we want to handle exceptions in the caller, then we can use `asyncio.as_completed` like this:

```
import asyncio


async def fail(t):
    await asyncio.sleep(t)
    print(f"Failing {t}")
    raise Exception


async def do():
    jobs = [fail(i) for i in range(10)]
    for f in asyncio.as_completed(jobs):
        try:
            await f
        except Exception:
            print("Caught an exception!")


asyncio.get_event_loop().run_until_complete(do())
```

Otherwise, we could just wrap our tasks with exception catching handlers instead of passing in callbacks.

server/gameconnection.py
```diff
@@ -536,13 +537,21 @@ def _mark_dirty(self):
...
-            tasks.append(peer.send_DisconnectFromPeer(self.player.id))
-
-        for result in await asyncio.gather(*tasks, return_exceptions=True):
-            if isinstance(result, Exception):
-                self._logger.exception(
-                    "peer_sendDisconnectFromPeer failed for player %i",
-                    self.player.id)
+            try:
+                tasks.append(peer.send_DisconnectFromPeer(self.player.id))
+            except AttributeError:
```
Wesmania: This looks very fishy, is it possible to fix lower down? If it's difficult, we can let it slide as a hotfix, but then let's add a FIXME here.

server/lobbyconnection.py
```diff
@@ -339,11 +340,12 @@ def on_connection_made(self, protocol: QDataStreamProtocol, peername: Address):
                             player.lobby_connection.send_warning(message_text)
                         )
                     except AttributeError:
```
Wesmania: idem

server/gameconnection.py
Askaholic: Not sure. Basically what happens sometimes (it's pretty rare) is that a player can disconnect causing their `lobby_connection` object to be deleted. Since players only have a weak reference to the lobby connection, calls like `player.lobby_connection.send` can fail with an `AttributeError` caused by `None.send`.

Wesmania: Alright, let's just put a FIXME comment here then.

server/lobbyconnection.py
```diff
@@ -339,11 +340,12 @@ def on_connection_made(self, protocol: QDataStreamProtocol, peername: Address):
...
+                        # In case player.lobby_connection was None
                         self._logger.debug("Failed to send broadcast to %s", player)
                     except Exception:
                         self._logger.exception("Failed to send broadcast to %s", player)
 
-                await asyncio.gather(*tasks, return_exceptions=True)
```
Wesmania: It's a matter of preference, but I'd catch ConnectionError either directly with `as_completed` or by wrapping each task with a try-catch. Just to avoid that 'pass exception type to function in order to catch it' thing :/

Askaholic: I dunno. To me it feels quite similar to the `contextlib.suppress` function. It helps you write your try...catch...ignore block in one line instead of 5 lines. I don't really see a benefit to writing the whole thing out each time when we're just ignoring a specific type of exception.

```python
with contextlib.suppress(AssertionError, ConnectionError):
    assert False
```

I don't think there is inherently anything wrong with passing exception classes as types.

server/gameconnection.py
Askaholic: So what exactly is the fixme? I see this as the same problem as the classic "check if file exists, then do something" problem where state can change between the check happening and the logic.

```python
if player.lobby_connection:
    # Maybe the GIL saves us here, but I still see this as bad practice
    await player.lobby_connection.send("msg")
```
vs
```python
try:
    await player.lobby_connection.send("msg")
except AttributeError:
    print("Player has disconnected while we tried to contact them")
```

Askaholic: I think in the party system PR, we refactored this to add a `send_message` function to player which can handle this error in one place. So instead of `player.lobby_connection.send` you would just call `player.send_message` which we could guarantee not to throw an exception.

Wesmania: Catching `AttributeError` can hide genuine programmer errors. We *need* to check for `None` further down.

Wesmania: There's also no TOCTOU here, since we're not awaiting between checking the weakref and using it.

server/async_functions.py
```diff
@@ -0,0 +1,19 @@
...
+    tasks: List[asyncio.Task],
+    *exc_classes: type,
+) -> None:
+    """
+    Call gather on a list of tasks, raising the first exception that dosen't
+    match any of the specified exception classes.
+    """
+    results = await asyncio.gather(*tasks, return_exceptions=True)
```
norraxx: From docs
```
If return_exceptions is True, exceptions are treated the same as successful results, and aggregated in the result list.
```
Do we cancel on first exception here?

server/gameconnection.py
norraxx: Also, I'd think, why won't we check `player.lobby_connection` for `isinstance`/`None` on place, from where attribute error is raised?
Like, yes: we can handle exception, but it's not the fix. Problem might be hidden else.

server/async_functions.py
Askaholic: Ok nice. That's much better than a callback.

server/async_functions.py
Askaholic: The docs are a little vague, but I've used gather enough to be able to explain how it works. Basically, `gather` always runs every coroutine. Even if one raises an exception, the rest all still get to complete normally. The difference is in how these exceptions are handled. If `return_exceptions` is false, then the call to `gather` will raise the first exception that it encounters. If `return_exceptions` is true, then all exceptions will be aggregated into the result list so you can iterate through them later.

server/gameconnection.py
Askaholic: Ok, from the docs it says that checking for None is in fact the right thing to do.
```python
# r is a weak reference object
o = r()
if o is None:
    # referent has been garbage collected
    print("Object has been deallocated; can't frobnicate.")
else:
    print("Object is still live!")
    o.do_something_useful()
```
`the idiom shown above is safe in threaded applications as well as single-threaded applications.`

server/async_functions.py
```diff
@@ -0,0 +1,23 @@
...
+    for result in results:
+        if isinstance(result, Exception):
+            # Check if this exception is an instance (maybe subclass) that
+            # should be ignored
+            for exc_type in exc_classes:
+                if not isinstance(result, exc_type):
+                    raise result
```
norraxx: I think, we should return `results`, if it's possible, even if it's not needed now. Maybe we would like process/log that results..

tests/conftest.py
```diff
@@ -25,7 +25,7 @@
 from server.rating import RatingType
 from tests.utils import MockDatabase
 
-logging.getLogger().setLevel(logging.DEBUG)
+logging.getLogger().setLevel(TRACE)
```
norraxx: What that trace does? I didn't get it O_o

tests/integration_tests/test_server.py
```diff
@@ -238,6 +232,103 @@
...
+    """
+    Causes connection errors in `do_report_dirties` which in turn will cause
+    closed games not to be cleaned up if the errors aren't handled properly.
+    """
+    # This test causes way to much logging output otherwise
+    caplog.set_level(logging.WARNING)
+
+    NUM_HOSTS = 10
+    NUM_PLAYERS_DC = 20
+    NUM_TIMES_DC = 10
```
norraxx: `NUM_PLAYERS_DC` + `NUM_TIMES_DC` what that `DC` is?

tests/integration_tests/test_server.py
```diff
@@ -238,6 +232,103 @@
...
+
+    # Set up our game hosts
+    host_protos = []
+    for _ in range(NUM_HOSTS):
+        _, _, proto = await connect_and_sign_in(
+            await tmp_user("Host"), lobby_server
+        )
+        host_protos.append(proto)
+    await asyncio.gather(*(
+        read_until_command(proto, "game_info") for proto in host_protos
```
norraxx: For better reading make for on next row

norraxx:
> Needs some commit squashing, but otherwise looks fine.
I made this before review.
`git rebase -i 0ea040f66f2b35b99401db00da8555b432134cb5`

server/async_functions.py
```diff
@@ -0,0 +1,23 @@
+"""
+Some helper functions for common async tasks.
+"""
+import asyncio
+from typing import List
+
+
+async def gather_without_exceptions(
+    tasks: List[asyncio.Task],
+    *exc_classes: type,
```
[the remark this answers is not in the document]
Askaholic: `type` is basically the type of a class in python. I read about it somewhere when I was researching metaclasses. Heres the docs page https://docs.python.org/3.6/library/functions.html?highlight=type#type.

I think I'll name the param like contextlib does `contextlib.suppress(*exceptions)`.

server/async_functions.py
```diff
@@ -0,0 +1,23 @@
...
+) -> None:
+    """
+    Call gather on a list of tasks, raising the first exception that dosen't
+    match any of the specified exception classes.
+    """
+    results = await asyncio.gather(*tasks, return_exceptions=True)
+    for result in results:
+        if isinstance(result, Exception):
```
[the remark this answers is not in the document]
Askaholic: That will cause it to ignore all exceptions if you don't pass any classes in.

tests/conftest.py
Askaholic: It is used for logging the contents of every message that is sent/received by lobbyconnection.

tests/integration_tests/test_server.py
Askaholic: Disconnect

Askaholic: Notice this line 
```python
# Set up our players that will disconnect
dc_players = [await tmp_user("Disconnecter") for _ in range(NUM_PLAYERS_DC)]
```

server/async_functions.py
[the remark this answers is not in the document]
norraxx: true

Issue #474, opened 2019-07-22
Fix broken stats collection


Askaholic:
In particular the user agent collection is a bit buggy and often shows negative values.
Instead of using separate guages for each user agent, we should:
- [ ] Use 1 "client connected" metric and tag it with the user agent
- [ ] Tag it with user_agent version as in: https://github.com/FAForever/server/pull/260

Other things:
- [ ] Remove `connections.aborted`. It doesn't seem to provide any useful info
- [ ] Figure out a better data format for ladder match quality
- [ ] Add a guage or counter for the number of active games
- [ ] Player count is only updated when someone logs in, not when someone disconnects

Maybe ditching `aiomeasures` statsd and switching to https://github.com/prometheus/client_python could help improve things. It would also give us CPU and memory usage as well.

cleborys:
I am not sure what is meant by "a better data format for ladder match quality" @Askaholic.
Is this already solved by making it a `Histogram` or `Summary` from the prometheus_client module?

Given that it is just a stream of floats between 0 and 1, I don't see why the format needs to be complicated.

Askaholic:
Yea I think that will cover it.

PR #526, opened 2020-02-02
Issue/#525 update python3.7

### LINKED ISSUE #525: Update to Python 3.7
Python 3.7 added some new features and optimizations to `asyncio`. 

Currently `aiomysql` gets installed from the master branch on github because it includes a fix that stops it from generating tons of warnings and also allows it to run in python3.7. If the folks at `aiomysql` have finally released that version on pypi, we should install from pypi again.

Askaholic:
Bump version numbers everywhere. Also updates dependencies.

Closes #525

### DIFF
--- a/Dockerfile
+++ b/Dockerfile
@@ -1,4 +1,4 @@
-FROM python:3.6-slim
+FROM python:3.7-slim
 
 # Apt-install c++ build tools and cleanup temporary files afterwards
 RUN apt-get update
--- a/Dockerfile
+++ b/Dockerfile
@@ -1,9 +1,9 @@
 FROM python:3.7-slim
 
-# Apt-install c++ build tools and cleanup temporary files afterwards
+# Need git for installing aiomysql
 RUN apt-get update
 RUN apt-get install -y --no-install-recommends \
-        build-essential git && \
+        git && \
     apt-get clean
 RUN rm -rf /var/lib/apt/lists/* /tmp/* /var/tmp/*
 
--- a/server/core/service.py
+++ b/server/core/service.py
@@ -1,0 +1,11 @@
+from abc import abstractmethod
+
+
+class Service(object):
+    @abstractmethod
+    async def initialize(self) -> None:
+        """
+        Called once before the service is used to run initial asynchronus tasks
+        such as loading data from the database.
+        """
+        pass
--- a/server/core/service.py
+++ b/server/core/service.py
@@ -1,6 +1,3 @@
-from abc import abstractmethod
-
-
 class Service(object):
     """
     All services should inherit from this class.
@@ -8,9 +5,14 @@ class Service(object):
     Services are singleton objects which manage some server task.
     """
 
-    @abstractmethod
     async def initialize(self) -> None:
         """
         Called once while the server is starting.
         """
-        pass
+        pass  # pragma: no cover
+
+    async def shutdown(self) -> None:
+        """
+        Called once after the server received the shutdown signal.
+        """
+        pass  # pragma: no cover
--- a/server/core/service.py
+++ b/server/core/service.py
@@ -2,10 +2,15 @@ from abc import abstractmethod
 
 
 class Service(object):
+    """
+    All services should inherit from this class.
+
+    Services are singleton objects which manage some server task.
+    """
+
     @abstractmethod
     async def initialize(self) -> None:
         """
-        Called once before the service is used to run initial asynchronus tasks
-        such as loading data from the database.
+        Called once while the server is starting.
         """
         pass
--- a/server/geoip_service.py
+++ b/server/geoip_service.py
@@ -12,12 +12,13 @@ import geoip2.database
 from maxminddb.errors import InvalidDatabaseError
 
 from . import config
+from .core import Service
 from .decorators import with_logger
 from .timing import Timer
 
 
 @with_logger
-class GeoIpService(object):
+class GeoIpService(Service):
     """
         Service for managing the GeoIp database. This includes an asyncio crontab
     which periodically checks if the current file is out of date. If it is, then
@@ -32,13 +33,15 @@ class GeoIpService(object):
         self.db_update_time = None
 
         self.check_geoip_db_file_updated()
+
+    async def initialize(self):
+        await self.check_update_geoip_db()
         # crontab: min hour day month day_of_week
         # Run every Wednesday because GeoLite2 is updated every first Tuesday
         # of the month.
         self._update_cron = aiocron.crontab(
             '0 0 0 * * 3', func=self.check_update_geoip_db
         )
-        asyncio.ensure_future(self.check_update_geoip_db())
         self._check_file_timer = Timer(
             60 * 10, self.check_geoip_db_file_updated, start=True
         )
--- a/server/geoip_service.py
+++ b/server/geoip_service.py
@@ -34,7 +34,7 @@ class GeoIpService(Service):
 
         self.check_geoip_db_file_updated()
 
-    async def initialize(self):
+    async def initialize(self) -> None:
         await self.check_update_geoip_db()
         # crontab: min hour day month day_of_week
         # Run every Wednesday because GeoLite2 is updated every first Tuesday
--- a/server/ladder_service.py
+++ b/server/ladder_service.py
@@ -294,6 +294,6 @@ class LadderService(Service):
         if player in self._informed_players:
             self._informed_players.remove(player)
 
-    def shutdown_queues(self):
+    async def shutdown(self):
         for queue in self.queues.values():
             queue.shutdown()
--- a/server/ladder_service.py
+++ b/server/ladder_service.py
@@ -10,6 +10,7 @@ from sqlalchemy import and_, func, select, text
 
 from .async_functions import gather_without_exceptions
 from .config import LADDER_ANTI_REPETITION_LIMIT
+from .core import Service
 from .db.models import game_featuredMods, game_player_stats, game_stats
 from .decorators import with_logger
 from .game_service import GameService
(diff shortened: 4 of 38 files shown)


Askaholic:
Looks like `aiomeasures` does not support python3.7. I guess this will need to be combined with #474 switching to `prometheus_client`.

```
future: <Task finished coro=<StatsD.send() done, defined at /usr/local/lib/python3.7/site-packages/aiomeasures/clients/statsd/client.py:37> exception=RuntimeError('generator raised StopIteration')>
  File "/usr/local/lib/python3.7/site-packages/aiomeasures/collectors.py", line 12, in flush
    metric = self.popleft()
IndexError: pop from an empty deque

During handling of the above exception, another exception occurred:

  File "/usr/local/lib/python3.7/site-packages/aiomeasures/collectors.py", line 28, in flush
    raise StopIteration
StopIteration

The above exception was the direct cause of the following exception:

  File "/usr/local/lib/python3.7/site-packages/aiomeasures/clients/statsd/client.py", line 45, in send
    yield from self.reporter.send(metrics)
  File "/usr/local/lib/python3.7/asyncio/coroutines.py", line 120, in coro
    res = func(*args, **kw)
  File "/usr/local/lib/python3.7/site-packages/aiomeasures/reporters/statsd_reporter.py", line 27, in send
    for i, metric in enumerate(metrics, start=1):
RuntimeError: generator raised StopIteration
```

setup.py
```diff
@@ -1,12 +1,27 @@
...
+    return m.groups()[0]
+
+
 setup(
     name='Forged Alliance Forever Server',
-    version=server.__version__,
+    version=get_version(),
     packages=['server'] + find_packages(),
     url='http://www.faforever.com',
     license=server.__license__,
```
norraxx: do we need that file at all?

setup.py
```diff
@@ -1,12 +1,27 @@
...
 import server
 
+
+def get_version() -> str:
+    output = subprocess.run(
+        ["git", "describe", "--abbrev=0", "--tags"],
+        capture_output=True
+    ).stdout.decode().strip()
+
+    m = re.match(r"v?(\d+\.\d+\.\d+)", output)
```
norraxx: I would say that second and third are not alwyas used.. `r"v?(\d+(\.\d+\(.\d+)?)?)"` o shorter `r"v?(\d+\.?)?"`
Like, somebody can write `v1` and somebody `v1.1`, and somebody `v1.1.1`

tests/unit_tests/test_lobbyconnection.py
```diff
@@ -927,6 +928,31 @@ def policy_server(event_loop):
     lobbyconnection.ladder_service.cancel_search.assert_called_with(lobbyconnection.player)
 
 
+async def test_command_matchmaker_info(lobbyconnection, ladder_service):
```
norraxx: Why do we have new test on matchmaker in python upgrade? O_o

setup.py
Askaholic: You mean `setup.py`? It is used to by the dockerfile to install the server. Not sure what the advantage is of doing it this way, but it doesn't hurt.

setup.py
Askaholic: Hm. That's probably true. I would prefer to use semver but I guess I can't guarantee that other people will always adhere to that.

tests/unit_tests/test_lobbyconnection.py
Askaholic: Hm, yea I'm not sure anymore. Maybe I just noticed it was uncovered when looking through coveralls or something.

server/geoip_service.py
```diff
@@ -32,13 +33,15 @@ def __init__(self):
         self.db_update_time = None
 
         self.check_geoip_db_file_updated()
+
+    async def initialize(self):
+        await self.check_update_geoip_db()
```
norraxx: :+1:

server/ladder_service.py
```diff
@@ -21,24 +22,35 @@
...
         # Hardcoded here until it needs to be dynamic
         self.queues = {
-            'ladder1v1': MatchmakerQueue('ladder1v1', games_service)
+            'ladder1v1': MatchmakerQueue('ladder1v1', games_service, loop=loop)
         }
 
         self.searches: Dict[str, Dict[Player, Search]] = defaultdict(dict)
 
-        asyncio.ensure_future(self.handle_queue_matches())
+        loop.create_task(self.handle_queue_matches())
```
norraxx: why not to just `await` for result?

server/matchmaker/matchmaker_queue.py
```diff
@@ -48,7 +49,9 @@ def __init__(
         self._is_running = True
 
         self.timer = PopTimer(self.queue_name)
-        asyncio.ensure_future(self.queue_pop_timer())
+        if not loop:
+            loop = asyncio.get_running_loop()
+        loop.create_task(self.queue_pop_timer())
```
norraxx: same here, why not to `await`?

setup.py
```diff
@@ -13,7 +13,7 @@ def get_version() -> str:
         capture_output=True
     ).stdout.decode().strip()
 
-    m = re.match(r"v?(\d+\.\d+\.\d+)", output)
+    m = re.match(r"v?(\d+(\.\d+(\.\d+)?)?)", output)
```
norraxx: :+1:

norraxx:
I've got warnings during `pipenv install`:

```
Ignoring idna-ssl: markers 'python_version < "3.7"' don't match your environment
Ignoring typing-extensions: markers 'python_version < "3.7"' don't match your environment
```
After `pipenv update`, it didn't happen.
Do we have actual `Pipfile.lock`? Or maybe i did something wrong?

norraxx:
Ignore, i've checked pipfile.lock, and they're removed, `pipenv --rm` ftw :-)

server/matchmaker/matchmaker_queue.py
norraxx: Imho, it's our opportunity to migrate everything to await, reason for `create_task` for me is only "run from nonasync code", otherwise it's almost the same, but more readable.

setup.py
norraxx: BTW: I've got idea! Find all `\d+` in version, and don't care about version format :-), then do everything you want with it. 

```python3
import re
re.findall("\d+", version)
```

server/ladder_service.py
Askaholic: `__init__` is not async

server/matchmaker/matchmaker_queue.py
Askaholic: Because this is in `__init__` it is not async

server/core/service.py
```diff
@@ -0,0 +1,11 @@
+from abc import abstractmethod
+
+
+class Service(object):
```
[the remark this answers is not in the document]
Askaholic: Yea I'm not quite sure how I want to do services. Another option is with a decorator, but ideally we'd refactor it so that services are created automatically, and all the initialize code is created automatically. I was also thinking maybe `update_data` should be a function on the base class since most of the services seem to use it.

setup.py
Askaholic: Oh yea that's not bad.

```python
".".join(re.findall(r"\d+", version)) or "dev"
```

server/ladder_service.py
norraxx: `=T_T=` True

server/core/service.py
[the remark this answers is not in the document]
norraxx: Well, if you have it that way, just write what do you define now, and how it should be used, it's interface what you're doing :-)

server/matchmaker/matchmaker_queue.py
Askaholic: Also in this case `create_task` is what we want. We can't `await` it because the coroutine runs until the server is shut down.

Dockerfile
```diff
@@ -1,9 +1,9 @@
-FROM python:3.6-slim
+FROM python:3.7-slim
 
-# Apt-install c++ build tools and cleanup temporary files afterwards
+# Need git for installing aiomysql
 RUN apt-get update
 RUN apt-get install -y --no-install-recommends \
-        build-essential git && \
+        git && \
```
norraxx: You can remove unused packages, after they're not needed.

server/core/dependency_injector.py
```diff
@@ -0,0 +1,137 @@
...
+            ]
+            if not leaves:
+                if not dep:
+                    return instances
+
+                # Find which dependencies are missing
+                missing = {
+                    d for dependencies in dep.values()
+                    for d in dependencies if d not in dep
+                }
```
norraxx: this double for loop is not good idea. It's unreadable.

server/core/dependency_injector.py
```diff
@@ -0,0 +1,137 @@
+import inspect
+from collections import ChainMap, defaultdict
+from typing import Dict, List
+
+DependencyGraph = Dict[str, List[str]]
+
+
+class DependencyInjector(object):
```
norraxx: imho, this code should be in separate ticket.

norraxx:
And maybe check something, that is already done: https://pypi.org/project/dependency-injector/

server/core/dependency_injector.py
Askaholic: Yea I guess that's reasonable. I don't think the dependency-injector package really does what I want. As far as I can tell with that package you still have to pass arguments in manually after you've defined the classes. Their code is all written in cython though and it's really hard to read.

server/matchmaker/matchmaker_queue.py
```diff
@@ -48,9 +48,12 @@ def __init__(
         self._is_running = True
 
         self.timer = PopTimer(self.queue_name)
-        asyncio.ensure_future(self.queue_pop_timer())
         self._logger.debug("MatchmakerQueue initialized for %s", queue_name)
 
+    async def initialize(self):
+        loop = asyncio.get_running_loop()
+        loop.create_task(self.queue_pop_timer())
```
Askaholic:
```suggestion
        asyncio.create_task(self.queue_pop_timer())
```

cleborys: Then you probably also don't want the `loop` from the previous line?

server/core/service.py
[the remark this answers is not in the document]
cleborys: Maybe also cleanup / shutdown, depending on how many services have one. (I only know of the ladder service.)

tests/unit_tests/test_lobbyconnection.py
cleborys: I think this is also a bit of a general structure refactoring now :)

server/stats/achievement_service.py
```diff
@@ -61,7 +62,7 @@
 
 
 @with_logger
-class AchievementService:
+class AchievementService(Service):
```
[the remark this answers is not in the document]
Askaholic: Yea it doesn't have any effect now that I'm moving the auto-service creation to a different PR. But with those changes it will mean that the class is instantiated and set up automatically. For now it's kindof just a marker I guess.

PR #529, opened 2020-02-03
Two player FFA

cleborys:
The `GameRater` would refuse to rate two-player games if one player had chosen a (numbered) team, while the other player remained teamless (the FFA team "-").
This should fix it.

Includes some further unit tests on various distributions of players on FFA and actual teams.

### DIFF
--- a/server/games/game_rater.py
+++ b/server/games/game_rater.py
@@ -44,6 +44,10 @@ class GameRater(object):
         example output: [ {p1: Rating, p2: Rating}, {p3: Rating, p4: Rating} ]
         """
         if FFA_TEAM in self._players_by_team:
+            number_of_players = sum([
+                len(self._players_by_team[team])
+                for team in self._players_by_team
+            ])
             if (
                 len(self._players_by_team[FFA_TEAM]) == 2
                 and len(self._players_by_team) == 1
@@ -51,11 +55,11 @@ class GameRater(object):
                 return [{
                     player: Rating(*player.ratings[self._rating_type])
                 } for player in self._players_by_team[FFA_TEAM]]
-            else:
+            elif number_of_players != 2:
                 raise GameRatingError(
                     f"Attempted to rate FFA game with other than two players: {{team: players}} = {self._players_by_team}"
                 )
-        elif len(self._players_by_team) == 2:
+        if len(self._players_by_team) == 2:
             return [{
                 player: Rating(*player.ratings[self._rating_type])
                 for player in team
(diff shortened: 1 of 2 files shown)


server/games/game_rater.py
```diff
@@ -44,18 +44,22 @@ def _get_rating_groups(self) -> List[Dict[Player, Rating]]:
         example output: [ {p1: Rating, p2: Rating}, {p3: Rating, p4: Rating} ]
         """
         if FFA_TEAM in self._players_by_team:
+            number_of_players = sum([
+                len(self._players_by_team[team])
```
Askaholic: How about
```python3
number_of_players = sum(len(players) for players in self._players_by_team.values())
```

I think using `.values` makes it a bit clearer since you only need to write `_players_by_team` once. Also by using a generator expression we can eliminate the need for creating an intermediate list that immediately gets discarded again!

server/games/game_rater.py
```diff
@@ -44,18 +44,22 @@ def _get_rating_groups(self) -> List[Dict[Player, Rating]]:
...
                 return [{
                     player: Rating(*player.ratings[self._rating_type])
                 } for player in self._players_by_team[FFA_TEAM]]
-            else:
+            elif number_of_players != 2:
                 raise GameRatingError(
                     f"Attempted to rate FFA game with other than two players: {{team: players}} = {self._players_by_team}"
                 )
-        elif len(self._players_by_team) == 2:
+        if len(self._players_by_team) == 2:
```
Askaholic: I might add a newline above, to make it clear that this is a new `if` block now.

tests/unit_tests/test_game_rater.py
```diff
@@ -129,3 +129,60 @@ def test_compute_rating():
     for team in result:
         for player, new_rating in team.items():
             assert new_rating != Rating(*player.ratings[RatingType.GLOBAL])
+
+
+def test_compute_rating_of_two_player_ffa_match_if_one_chose_a_team():
+    FFA_TEAM = 1
+    p1, p2 = MockPlayer(), MockPlayer()
+    players_by_team = {FFA_TEAM: [p1], 2: [p2]}
```
Askaholic: Awesome!

server/games/game_rater.py
cleborys: Nice, of course.
I wanted the intermediate list, because I thought that `sum` of an empty generator would throw an Error, while `sum([])` is just zero, but a quick test shows that that is also the case for "empty" generators! Not sure why I ever believed that...

tests/unit_tests/test_game_rater.py
cleborys: :)

PR #530, opened 2020-02-04
Issue/#527 fix game_launch options

### LINKED ISSUE #527: Fix game_launch expected_players
`expected_players` should be the number of players that the auto lobby waits for before launching the game. So for ladder this should always be 2. Currently it sends the number of connected players which is going to be 0 or 1.

Askaholic:
Add options to `launch_game` that allow setting parameters which can't be read from the Game class.

Closes #527

### DIFF
--- a/server/ladder_service.py
+++ b/server/ladder_service.py
@@ -16,6 +16,7 @@ from .game_service import GameService
 from .matchmaker import MatchmakerQueue, Search
 from .players import Player, PlayerState
 from .protocol import DisconnectedError
+from .types import GameLaunchOptions
 
 MapDescription = NamedTuple('Map', [("id", int), ("name", str), ("path", str)])
 
@@ -188,7 +189,13 @@ class LadderService:
             # FIXME: Database filenames contain the maps/ prefix and .zip suffix.
             # Really in the future, just send a better description
             self._logger.debug("Starting ladder game: %s", game)
-            await host.lobby_connection.launch_game(game, is_host=True, use_map=mapname)
+            options = GameLaunchOptions(
+                mapname=mapname,
+                expected_players=2
+            )
+            await host.lobby_connection.launch_game(
+                game, is_host=True, options=options
+            )
             try:
                 hosted = await game.await_hosted()
                 if not hosted:
@@ -204,7 +211,7 @@ class LadderService:
 
                 # TODO: Graceful handling of NoneType errors due to disconnect
                 await guest.lobby_connection.launch_game(
-                    game, is_host=False, use_map=mapname
+                    game, is_host=False, options=options
                 )
             self._logger.debug("Ladder game launched successfully")
         except Exception:
--- a/server/ladder_service.py
+++ b/server/ladder_service.py
@@ -189,29 +189,39 @@ class LadderService:
             # FIXME: Database filenames contain the maps/ prefix and .zip suffix.
             # Really in the future, just send a better description
             self._logger.debug("Starting ladder game: %s", game)
+            # Options shared by guest and host
             options = GameLaunchOptions(
                 mapname=mapname,
-                expected_players=2
+                team=1,
+                expected_players=2,
             )
             await host.lobby_connection.launch_game(
-                game, is_host=True, options=options
+                game,
+                is_host=True,
+                options=options._replace(
+                    faction=host.faction,
+                    map_position=1
+                )
             )
             try:
                 hosted = await game.await_hosted()
                 if not hosted:
                     raise TimeoutError("Host left lobby")
             finally:
-                self._logger.debug("Ladder game failed to launch due to a timeout")
-                # TODO: Uncomment this return once the client supports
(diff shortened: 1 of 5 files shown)


server/types.py
```diff
@@ -1,5 +1,24 @@
...
+    host: str
+    port: int
+
+    @classmethod
+    def from_string(cls, s: str) -> "Address":
+        host, port = s.split(":")
+        return cls(host, int(port))
+
+
+# TODO: Find an appropriate place for this
```
cleborys: As long as there is a `server.types` module, I think this is a good place to put it.

server/lobbyconnection.py
```diff
@@ -840,18 +845,14 @@ def on_connection_made(self, protocol: QDataStreamProtocol, peername: Address):
             "args": ["/numgames", self.player.game_count[RatingType.GLOBAL]],
             "uid": game.id,
             "mod": game.game_mode,
-            "mapname": use_map,
             # Following parameters are not used by the client yet. They are
```
cleborys: This comment block is now outdated and should probably be moved to the definition of `GameLaunchOptions`?

server/ladder_service.py
```diff
@@ -161,7 +162,13 @@ def _cancel_existing_searches(self, initiator: Player) -> List[Search]:
         # FIXME: Database filenames contain the maps/ prefix and .zip suffix.
         # Really in the future, just send a better description
         self._logger.debug("Starting ladder game: %s", game)
-        await host.lobby_connection.launch_game(game, is_host=True, use_map=mapname)
+        options = GameLaunchOptions(
+            mapname=mapname,
+            expected_players=2
+        )
```
cleborys: The _unused_ options "team", "faction", and "map_position" are now no longer set,
before they were coming from `game.get_player_options`.
Since they are unused I assume that is no problem, I just wanted to make sure this is as intended.

If they are ever used, is the correct place to include them here?
If _not_, then I would move them back into the lobbyconnection.
I guess @norraxx knows what he would want, since he is working on tmm?

server/lobbyconnection.py
Askaholic: I don't think so. The comment is still accurate. `name`, `init_mode` and any of the game launch options are not used by the client (and probably wont be until long after this code is deployed)

server/ladder_service.py
```diff
@@ -188,23 +189,39 @@ def _cancel_existing_searches(self, initiator: Player) -> List[Search]:
...
+            # Options shared by guest and host
+            options = GameLaunchOptions(
+                mapname=mapname,
+                team=1,
+                expected_players=2,
+            )
+            await host.lobby_connection.launch_game(
+                game,
+                is_host=True,
+                options=options._replace(
```
norraxx: imho `_replace` should be public. ;-)

server/types.py
```diff
@@ -1,5 +1,23 @@
...
+
+class Address(NamedTuple):
+    """A peer IP address"""
+
+    host: str
+    port: int
+
+    @classmethod
+    def from_string(cls, s: str) -> "Address":
+        host, port = s.split(":")
```
norraxx: please, don't name `s`, name it as `string_address`

norraxx: I would use `rsplit` with limitation on 2 peaces (checkou docs, imho it's something like `rsplit(":", 1)`

server/types.py
```diff
@@ -1,5 +1,23 @@
...
 
-# Represents a peer IP address
-Address = NamedTuple('Address', [('host', str), ('port', int)])
-Address.from_string = lambda s: Address(*(s.split(':')))
+
+class Address(NamedTuple):
+    """A peer IP address"""
+
+    host: str
+    port: int
```
norraxx: When i see this, i was thinking, when we will be on 3.7 we might use here dataclasses. https://docs.python.org/3/library/dataclasses.html
Do you have any experience with them? Are they better/worse than named tuples?

norraxx:
This coveralls are crazy. Like you're trying to fix something or add functionality, and to pass it, you can't go lower than before... crazy...

cleborys:
Coveralls has an option for that which we can set to something more reasonable, for example such that it only complains when coverage decreases by more than one percent.
Should be something like "Repo Settings > Coverage Decrease Threshold For Failure".

Askaholic:
The coveralls check doesn't block merging, it's just there so we can see how the coverage changed.

server/types.py
Askaholic: No I haven't used them before.

cleborys:
It's just sometimes a bit annoying to have that little red X next to PRs that are ready for merging, because it suggests that things are still being worked on and likely to change.

Askaholic:
Yea but I see it as more of a good thing because it makes me actually click the "Details" button and go look at why the coverage decreased. Then I can make an informed decision as to whether I should go back and add more test, or whether it's just an artifact of coveralls.

PR #531, opened 2020-02-06
Improve matchmaking performance

Askaholic:
Turns out that matchmaking performance is dominated by the creation of the `MatchingGraph`. Most of this comes from calls to `trueskill` for determining the edge weights but there were a few other instances of unnecessary work being performed.

### Improvements made:
- Precompute quality against yourself. Easily eliminates 10's of calls to trueskill per search.
- Cache calls to `Search.quality_with` for the duration of `MatchingGraph` creation. This significantly reduces the amount of trueskill calls made.
- Only compute quality in one direction since quality is symmetric (`self.quality_with(other) == other.quality_with(self)`)

### My benchmarks:
For a MatchingGraph on 30 searches, these improvements bring the time of `make_matches` from 2.294 per call down to 0.148 per call. Easily a 10x speedup.

Before:
```
         35261689 function calls (34610527 primitive calls) in 22.939 seconds

   Ordered by: cumulative time
   List reduced from 168 to 35 due to restriction <'fa-server/server'>

   ncalls  tottime  percall  cumtime  percall filename:lineno(function)
       10    0.000    0.000   22.937    2.294 /home/dope/Documents/fa-server/server/matchmaker/algorithm.py:20(make_matches)
       10    0.000    0.000   22.937    2.294 /home/dope/Documents/fa-server/server/matchmaker/algorithm.py:138(find)
       10    0.002    0.000   22.937    2.294 /home/dope/Documents/fa-server/server/matchmaker/algorithm.py:44(find)
       10    0.000    0.000   22.542    2.254 /home/dope/Documents/fa-server/server/matchmaker/algorithm.py:161(build_sparse)
       10    0.001    0.000   22.542    2.254 /home/dope/Documents/fa-server/server/matchmaker/algorithm.py:169(<dictcomp>)
      300    0.001    0.000   22.113    0.074 /home/dope/Documents/fa-server/server/matchmaker/algorithm.py:179(_get_top_matches)
    72350    0.232    0.000   22.084    0.000 /home/dope/Documents/fa-server/server/matchmaker/search.py:148(quality_with)
     8700    0.142    0.000   21.194    0.002 /home/dope/Documents/fa-server/server/matchmaker/algorithm.py:181(is_possible_match)
    47060    0.109    0.000   14.612    0.000 /home/dope/Documents/fa-server/server/matchmaker/search.py:134(match_threshold)
     8700    0.028    0.000    7.184    0.001 /home/dope/Documents/fa-server/server/matchmaker/search.py:168(matches_with)
    17400    0.031    0.000    5.790    0.000 /home/dope/Documents/fa-server/server/matchmaker/search.py:214(__str__)
```

After:
```
         2448879 function calls (2404767 primitive calls) in 1.575 seconds

   Ordered by: cumulative time
   List reduced from 161 to 34 due to restriction <'fa-server/server'>

   ncalls  tottime  percall  cumtime  percall filename:lineno(function)
     4900    0.015    0.000    1.488    0.000 /home/dope/Documents/fa-server/server/matchmaker/search.py:150(quality_with)
       10    0.000    0.000    1.482    0.148 /home/dope/Documents/fa-server/server/matchmaker/algorithm.py:9(make_matches)
       10    0.000    0.000    1.482    0.148 /home/dope/Documents/fa-server/server/matchmaker/algorithm.py:131(find)
       10    0.002    0.000    1.481    0.148 /home/dope/Documents/fa-server/server/matchmaker/algorithm.py:33(find)
       10    0.012    0.001    1.398    0.140 /home/dope/Documents/fa-server/server/matchmaker/algorithm.py:154(build_weighted)
      300    0.002    0.000    0.093    0.000 /home/dope/Documents/fa-server/server/matchmaker/search.py:20(__init__)
     9800    0.023    0.000    0.088    0.000 /home/dope/Documents/fa-server/server/matchmaker/search.py:77(ratings)
      390    0.001    0.000    0.079    0.000 /home/dope/Documents/fa-server/server/matchmaker/algorithm.py:75(_propose)
     4350    0.018    0.000    0.062    0.000 /home/dope/Documents/fa-server/server/matchmaker/algorithm.py:181(is_possible_match)
    14980    0.014    0.000    0.038    0.000 /home/dope/Documents/fa-server/server/matchmaker/search.py:137(match_threshold)
     9800    0.008    0.000    0.038    0.000 /home/dope/Documents/fa-server/server/matchmaker/search.py:87(raw_ratings)
```

### DIFF
--- a/server/matchmaker/algorithm.py
+++ b/server/matchmaker/algorithm.py
@@ -126,20 +126,20 @@ class Matchmaker(object):
 
     def find(self) -> List[Match]:
         self._logger.debug("Matching with stable marriage...")
-        searches = [s for s in self.searches if s is not None]
+        searches = list(self.searches)
         if len(searches) < 30:
             ranks = _MatchingGraph.build_full(searches)
         else:
             ranks = _MatchingGraph.build_fast(searches)
         self.matches.update(StableMarriage().find(ranks))
 
-        self._register_unmatched_searches()
-
         remaining_searches = [
             search for search in self.searches if search not in self.matches
         ]
         self._logger.debug("Matching randomly for remaining newbies...")
         self.matches.update(RandomlyMatchNewbies().find(remaining_searches))
+
+        self._register_unmatched_searches()
 
         return self._remove_duplicates()
 
@@ -148,11 +148,10 @@ class Matchmaker(object):
         Tells all unmatched searches that they went through a failed matching
         attempt.
         """
-        unmatched_searches = filter(
-            lambda search: search not in self.matches,
-            self.searches
-        )
-        for search in unmatched_searches:
+        for search in self.searches:
+            if search in self.matches:
+                continue
+
             search.register_failed_matching_attempt()
             self._logger.debug(
                 "Search %s remained unmatched at threshold %f in attempt number %i",
@@ -171,7 +170,7 @@ class Matchmaker(object):
 @with_logger
 class _MatchingGraph:
     @staticmethod
-    def build_full(searches: Iterable[Search]) -> WeightedGraph:
+    def build_full(searches: List[Search]) -> WeightedGraph:
         """ A graph in adjacency list representation, whose nodes are the searches
         and whose edges are the possible matchings for each node. Checks every
         possible edge for inclusion in the graph.
@@ -200,7 +199,7 @@ class _MatchingGraph:
         return adj_list
 
     @staticmethod
-    def build_fast(searches: Iterable[Search]) -> WeightedGraph:
+    def build_fast(searches: List[Search]) -> WeightedGraph:
         """ Builds approximately the same graph as `build_full`, but does not
         check every possible edge.
 
@@ -255,22 +254,22 @@ class _MatchingGraph:
 
         Note: This assumes that edges are undirected. Calling this on directed
         graphs will produce incorrect results. """
-        for k, v in list(graph.items()):
-            if not v:
-                del graph[k]
+        for search, neighbors in list(graph.items()):
+            if not neighbors:
+                del graph[search]
 
 
 def subset_pairs(l: list):
-    """ Generates all possible 2 subsets of `l` as tuples. Each pair of items will
-    show up in only one order and the elements in the pair will be distinct.
-    Note that the number of items generated is `len(l) choose 2`.
+    """ Generates all possible 2 subsets of `l` as tuples. Each pair of items
+    will show up in only one order and the elements in the pair will be
+    distinct. Note that the number of items generated is `len(l) choose 2`.
 
     # Example
+    `subset_pairs([1, 2, 3])`
+        yields pairs (1, 2), (1, 3), (2, 3)
 
-    subset_pairs([1, 2, 3]) yields pairs (1, 2), (1, 3), (2, 3)
-
-    Note that this has O(n^2) time complexity.
+    Time complexity: O(n^2)
     """
     for i, a in enumerate(l[:-1]):
         for b in l[i+1:]:
-            yield a, b
+            yield (a, b)
--- a/server/matchmaker/algorithm.py
+++ b/server/matchmaker/algorithm.py
@@ -12,8 +12,7 @@ def make_matches(searches: Iterable[Search]) -> List[Match]:
 
 @with_logger
 class MatchmakingPolicy(object):
-    def __init__(self, searches: Iterable[Search]):
-        self.searches = searches
+    def __init__(self):
         self.matches: Dict[Search, Search] = {}
 
     def _match(self, s1: Search, s2: Search):
@@ -30,17 +29,13 @@ class MatchmakingPolicy(object):
 
 
 class StableMarriage(MatchmakingPolicy):
-    def find(self) -> Dict[Search, Search]:
+    def find(self, ranks: WeightedGraph) -> Dict[Search, Search]:
         """ Perform the stable matching algorithm until a maximal stable matching
-        is found. Assumes that _MatchingGraph.build_weighted() only returns edges
-        whose matches are acceptable to both parties.
+        is found.
         """
-        ranks = _MatchingGraph.build_weighted(self.searches)
-        _MatchingGraph.remove_isolated(ranks)
-        max_degree = max((len(edges) for edges in ranks.values()), default=0)
-
         self.matches.clear()
 
+        max_degree = max((len(edges) for edges in ranks.values()), default=0)
         for i in range(max_degree):
             self._logger.debug(
                 "Round %i of stable marriage, currently %i matches", i,
@@ -111,12 +106,12 @@ class StableMarriage(MatchmakingPolicy):
 
 
 class RandomlyMatchNewbies(MatchmakingPolicy):
-    def find(self) -> Dict[Search, Search]:
+    def find(self, searches: Iterable[Search]) -> Dict[Search, Search]:
         self.matches.clear()
 
         unmatched_newbies = [
-            search for search in self.searches
-            if search.is_single_ladder_newbie() and search not in self.matches
+            search for search in searches
+            if search.is_single_ladder_newbie()
         ]
 
         while len(unmatched_newbies) >= 2:
@@ -130,7 +125,7 @@ class RandomlyMatchNewbies(MatchmakingPolicy):
             default_if_no_available_opponent = None
 
             opponent = next((
-                search for search in self.searches
+                search for search in searches
                 if search != newbie and search not in self.matches
                 and search.is_single_party() and search.has_no_top_player()
             ), default_if_no_available_opponent)
--- a/server/matchmaker/algorithm.py
+++ b/server/matchmaker/algorithm.py
@@ -1,4 +1,5 @@
 import math
+from statistics import mean
 from typing import Dict, Iterable, List, Set, Tuple
 
 from ..decorators import with_logger
@@ -206,9 +207,11 @@ class _MatchingGraph:
         Time complexity: O(n*log(n))
         """
         adj_list = {search: [] for search in searches}
-        # TODO: Support searches with larger parties
-        # Sort all searches by their average rating
-        searches = sorted(searches, key=lambda s: s.ratings[0][0])
+        # Sort all searches by players average trueskill mean
+        searches = sorted(
+            searches,
+            key=lambda s: mean(map(lambda r: r[0], s.ratings))
+        )
         # Now compute quality with `num_to_check` nearby searches on either side
         num_to_check = int(math.log(max(16, len(searches)), 2)) // 2
         for i, search in enumerate(searches):
--- a/server/matchmaker/algorithm.py
+++ b/server/matchmaker/algorithm.py
@@ -3,7 +3,7 @@ from typing import Dict, Iterable, List, Set, Tuple
 from ..decorators import with_logger
 from .search import Match, Search
 
-WeighedGraph = Dict[Search, List[Tuple[Search, float]]]
+WeightedGraph = Dict[Search, List[Tuple[Search, float]]]
 
 
 def make_matches(searches: Iterable[Search]) -> List[Match]:
@@ -142,7 +142,7 @@ class RandomlyMatchNewbies(MatchmakingPolicy):
 
 @with_logger
 class Matchmaker(object):
-    def __init__(self, searches: List[Search]):
+    def __init__(self, searches: Iterable[Search]):
         self.searches = searches
         self.matches: Dict[Search, Search] = {}
 
@@ -170,7 +170,7 @@ class Matchmaker(object):
 @with_logger
 class _MatchingGraph:
     @staticmethod
-    def build_weighted(searches: Iterable[Search]) -> WeighedGraph:
+    def build_weighted(searches: Iterable[Search]) -> WeightedGraph:
         """ A graph in adjacency list representation, whose nodes are the searches
         and whose edges are the possible matchings for each node.
 
@@ -217,7 +217,7 @@ class _MatchingGraph:
             return False
 
     @staticmethod
-    def remove_isolated(graph: WeighedGraph):
+    def remove_isolated(graph: WeightedGraph):
         """ Remove any searches that have no possible matchings.
 
         Note: This assumes that edges are undirected. Calling this on directed
--- a/server/matchmaker/algorithm.py
+++ b/server/matchmaker/algorithm.py
@@ -70,8 +70,10 @@ class StableMarriage(MatchmakingPolicy):
 
                 self._logger.debug(
                     "Quality between %s and %s: %f thresholds: [%f, %f]",
-                    search, preferred, search.quality_with(preferred),
-                    search.match_threshold, preferred.match_threshold
+                    search, preferred,
+                    search.quality_with(preferred),
+                    search.match_threshold,
+                    preferred.match_threshold
                 )
 
                 self._propose(search, preferred)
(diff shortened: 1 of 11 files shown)


server/matchmaker/algorithm.py
```diff
@@ -1,29 +1,18 @@
...
-from typing import Dict, Iterable, List, Set
+from typing import Dict, Iterable, List, Set, Tuple
 
 from ..decorators import with_logger
 from .search import Match, Search
 
-################################################################################
-#                           Constants and parameters                           #
-################################################################################
+WeighedGraph = Dict[Search, List[Tuple[Search, float]]]
```
cleborys: `WeightedGraph`? Unless my English is off :)
(The type is consistently without a t, but there is one in `build_weighted`)

server/matchmaker/algorithm.py
```diff
@@ -42,39 +31,44 @@ def _unmatch(self, s1: Search):
...
-        for i in range(SM_NUM_TO_RANK):
+        for i in range(max_degree):
             self._logger.debug(
                 "Round %i of stable marriage, currently %i matches", i,
                 len(self.matches) // 2
             )
             # Do one round of proposals
-            if len(self.matches) == len(self.searches):
+            if len(self.matches) == len(ranks):
                 # Everyone found a match so we are done
```
cleborys: Depending on your style choices, maybe you can avoid the comment like that:
```python
everyone_found_a_match = (len(self.matches) == len(ranks))
if everyone_found_a_match:
  break
```
(the brackets are unnecessary but I feel they improve readability)

server/matchmaker/search.py
```diff
@@ -173,8 +175,20 @@ def matches_with(self, other: 'Search'):
...
+    def _match_quality_acceptable(self, other: 'Search', quality: float) -> bool:
+        """
+        Determine if the given match quality is acceptable.
+
+        This gets it's own function so we can call it from the Matchmaker using
+        a cached `quality` value.
+        """
+        # NOTE: We are assuming for optimization purposes that quality is
+        # symmetric. If this ever changes, update here
+        if quality >= self.match_threshold and quality >= other.match_threshold:
```
cleborys: Why not simply return the condition of the `if` statement?
Or if it is more readable:
```python
is_acceptable = quality >= self.match_threshold and quality >= other.match_threshold
return is_acceptable
```

server/matchmaker/algorithm.py
```diff
@@ -42,39 +31,44 @@ def _unmatch(self, s1: Search):
...
                 break
 
-            for search in self.searches:
+            for search in ranks:
                 if search in self.matches:
                     continue
 
                 if not ranks[search]:
                     # Preference list exhausted
```
cleborys: as above maybe
```python
preference_list_exhausted = len(ranks[search]) == 0
if preference_list_exhausted:
  continue
```

server/matchmaker/algorithm.py
Askaholic: Yea, that was a typo

server/matchmaker/search.py
Askaholic: Good point :P

server/matchmaker/algorithm.py
```diff
@@ -41,64 +30,45 @@ def _unmatch(self, s1: Search):
...
-        ranks = _MatchingGraph.build_sparse(self.searches)
+    def find(self, ranks: WeightedGraph) -> Dict[Search, Search]:
+        """ Perform the stable matching algorithm until a maximal stable matching
+        is found.
+        """
         self.matches.clear()
 
-        for i in range(SM_NUM_TO_RANK):
+        max_degree = max((len(edges) for edges in ranks.values()), default=0)
+        for i in range(max_degree):
```
norraxx: If variable i is not used, in python we use underscore.

server/matchmaker/algorithm.py
norraxx: If this calculations are not in loop, i don't see the reason for that.

server/matchmaker/algorithm.py
norraxx: More lines, more code. I disagree

server/matchmaker/algorithm.py
```diff
@@ -151,22 +120,45 @@ def find(self) -> Dict[Search, Search]:
...
 class Matchmaker(object):
-    def __init__(self, searches: List[Search]):
+    def __init__(self, searches: Iterable[Search]):
         self.searches = searches
         self.matches: Dict[Search, Search] = {}
 
     def find(self) -> List[Match]:
         self._logger.debug("Matching with stable marriage...")
-        self.matches.update(StableMarriage(self.searches).find())
+        searches = [s for s in self.searches if s is not None]
```
norraxx: How is it possible to have `None` values in that searches. Like list witho bjects and nones is bad. If this is necessary, then typing of self search in init should be iterable optional search

server/matchmaker/algorithm.py
```diff
@@ -151,22 +120,45 @@ def find(self) -> Dict[Search, Search]:
...
 
+    def _register_unmatched_searches(self):
+        """
+        Tells all unmatched searches that they went through a failed matching
+        attempt.
+        """
+        unmatched_searches = filter(
+            lambda search: search not in self.matches,
+            self.searches
+        )
```
norraxx: Imho, we should use one of this: list comprehension or this map functions.
But this is more on discussion. In fact there is no reason to not use both.

server/matchmaker/algorithm.py
```diff
@@ -151,22 +120,45 @@ def find(self) -> Dict[Search, Search]:
...
+        for search in unmatched_searches:
```
norraxx: Btw, you could filter with one more if + indent in that loop. Less code :-)

server/matchmaker/algorithm.py
Askaholic: TBH I have no idea how that got in there. Somehow I thought it was in the code before and I just left it alone, but looking at the diff I don't see it.

server/matchmaker/algorithm.py
Askaholic: It is used for logging

server/matchmaker/algorithm.py
Askaholic: Well, the reason would be convey the information from the comment directly in the code. This way you can avoid having a stale comment. Because if someone ever does change the code, then they will remember to rename the variable if needed, but they might not remember to update the comment. 

Personally, I don't really like it though. I think that solution is pretty ugly and I'd prefer the comment. Usually I try to avoid comments, but I think if you're going to put a comment anywhere in your code, it should be in the part with the complicated algorithm. I see the intent, but I prefer the comment in this situation.

server/matchmaker/algorithm.py
Askaholic: Good point

server/matchmaker/algorithm.py
```diff
@@ -179,44 +170,105 @@ def _remove_duplicates(self) -> List[Match]:
...
-                    f"{quality_log_string} Will be discarded for stable marriage."
-                )
-                return False
+        Time complexity: O(n*log(n))
+        """
+        adj_list = {search: [] for search in searches}
+        # TODO: Support searches with larger parties
+        # Sort all searches by their average rating
+        searches = sorted(
+            searches, key=lambda s: s.ratings[0][0] - 3 * s.ratings[0][1]
```
cleborys: You are using the "displayed" rating of `mean - 3 * deviation`?
I would have expected just the mean here, I think that will be better for trueskill 
(match quality is more likely to be high if the means are close, not if the adjusted values are close )

I think of the `mean - 3 * deviation` as the "displayed" rating, because I think it was meant as a hack to a) give new players who aren't accurately rated yet the positive reinforcement of a growing rating and b) make sure inaccurately rated players aren't overly confident/arrogant.
If I'm not mistaken there is no "mathematical" reasoning behind the value.

cleborys: You are using the "displayed" rating of `mean - 3 * deviation`?
I would have expected just the mean here, I think that will be better for trueskill 
(match quality is more likely to be high if the means are close, not if the adjusted values are close )

I think of the `mean - 3 * deviation` as the "displayed" rating, because I think it was meant as a hack to a) give new players who aren't accurately rated yet the positive reinforcement of a growing rating at the start and b) make sure inaccurately rated players aren't overly confident/arrogant.
If I'm not mistaken there is no "mathematical" reasoning behind the value.

server/matchmaker/algorithm.py
```diff
@@ -179,44 +170,105 @@ def _remove_duplicates(self) -> List[Match]:
...
+        )
+        # Now compute quality with `num_to_check` nearby searches on either side
+        num_to_check = max((4, int(math.log(len(searches), 2)))) // 2
```
cleborys: You chose that so that it checks the highest number of edges while still staying `O(n log n)`? Nice :)

tests/conftest.py
```diff
@@ -141,29 +142,54 @@ def make_game(database, uid, players):
...
+):
+    ratings = {k: v for k, v in {
+     RatingType.GLOBAL: global_rating,
+     RatingType.LADDER_1V1: ladder_rating,
+    }.items() if v is not None}
+
+    games = {k: v for k, v in {
+     RatingType.GLOBAL: numGames,
+     RatingType.LADDER_1V1: ladder_games
+    }.items() if v is not None}
```
cleborys: This `is not None` check feels slightly weird, since the default value for these is zero, not `None`.

tests/unit_tests/conftest.py
```diff
@@ -134,3 +135,34 @@ def add(gameobj: Game, n: int, team: int=None):
         return players
 
     return add
+
+
+class Benchmark(object):
```
cleborys: Awesome!

tests/unit_tests/test_matchmaker_queue_algorithm.py
```diff
@@ -7,32 +9,112 @@
...
+    s2 = Search([p(1500, 63, ladder_games=20)])
+    s3 = Search([p(1600, 75, ladder_games=50)])
+    searches = [s1, s2, s3]
+
+    ranks = algorithm._MatchingGraph.build_fast(searches)
+
+    import pprint
+
+    pp = pprint.PrettyPrinter(indent=1)
+    pp.pprint(ranks)
```
cleborys: I assume all the printing is meant to be removed now? :)

tests/unit_tests/test_matchmaker_queue_algorithm.py
```diff
@@ -7,32 +9,112 @@
...
         s2: [s3, s1],
         s3: [s1, s2]
+    })
+
+
+def test_remove_isolated(p):
+    s1 = Search([p(1500, 64, ladder_games=20)])
+    s2 = Search([p(1500, 63, ladder_games=20)])
+    s3 = Search([p(1600, 75, ladder_games=50)])
+    ranks = add_graph_edge_weights({
```
cleborys: This example is a bit weird because it is not undirected (or not symmetric, if you want).
i.e. there is an edge s1 -> s2 but not s2 -> s1.
In case we ever expect to have directed graphs I would expect `remove_isolated` to only remove nodes which have neither in- nor outgoing edges, so I would rather document that expectation.
That is,
```python
  s1: [s3],
  s2: [],
  s3: [s1],
```

tests/unit_tests/test_matchmaker_queue_algorithm.py
```diff
@@ -7,32 +9,112 @@
...
+        s1: [s3, s2],
+        s2: [],
+        s3: [s1, s2]
+    })
+
+    algorithm._MatchingGraph.remove_isolated(ranks)
+
+    assert ranks == add_graph_edge_weights({
+        s1: [s3, s2],
```
cleborys: This is now even weirder, since `s2` doesn't exist anymore (in the way I would interpret it).

tests/unit_tests/test_matchmaker_queue_algorithm.py
```diff
@@ -7,32 +9,112 @@
...
+        s2: [],
+        s3: []
     }
 
+    algorithm._MatchingGraph.remove_isolated(ranks)
+
+    assert ranks == {}
+
 
 def test_match_graph_will_not_include_matches_below_threshold_quality(p):
```
cleborys: This test should probably cover both `build_full` and `build_fast`? (Or there should be an equivalent test for `build_fast`.)

cleborys: That might also go for some tests below, your choice.

server/matchmaker/algorithm.py
```diff
@@ -111,7 +81,6 @@ def _propose(self, search: Search, preferred: Search):
 
         current_match = self.matches[preferred]
         current_quality = preferred.quality_with(current_match)
```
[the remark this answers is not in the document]
Askaholic: Yes I considered it, but from the benchmarks, this is pretty inconsequential. This code contributes only a tiny fraction of total run time so I didn't bother to optimize it.

server/matchmaker/algorithm.py
Askaholic: Yea I figured we might as well maximize the number of opponents checked while still keeping the time complexity the same.

tests/unit_tests/test_matchmaker_queue_algorithm.py
Askaholic: Right.

tests/unit_tests/test_matchmaker_queue_algorithm.py
Askaholic: Mmmm. Good catch. Maybe I'll rename the function or add to the docstring that it assumes edges are bidirectional. Because such a graph should never be created so there's no point in adding extra logic to handle that case if we don't need it.

tests/unit_tests/test_matchmaker_queue_algorithm.py
Askaholic: I could probably handle this with a `pytest.mark.parametrize`.

server/matchmaker/algorithm.py
Askaholic: Yea that's true. That will also make extending to parties easier too since we would just use the average rating.

tests/conftest.py
Askaholic: Yea that is odd. I don't see them set to `None` anywhere in the tests.

tests/unit_tests/test_matchmaker_queue_algorithm.py
Askaholic: Decided to just add a description to the docstring since I don't really expect our matchmaking to ever need directed graphs. I was actually thinking this function might not be necessary at all. I doubt that it makes a performance impact to remove isolated vertices before running the matching algorithm since they will just be skipped anyways.

tests/unit_tests/test_matchmaker_queue_algorithm.py
```diff
@@ -254,7 +387,32 @@ def test_matchmaker(p):
...
+
+def test_matchmaker_performance(p, bench, caplog):
+    # Disable debug logging for performance
+    caplog.set_level(logging.INFO)
+    NUM_SEARCHES = 200
+
+    searches = [Search([p(1500, 500, ladder_games=1)]) for _ in range(NUM_SEARCHES)]
+
+    with bench:
+        _ = algorithm.make_matches(searches)
```
cleborys:
```suggestion
        algorithm.make_matches(searches)
```

tests/unit_tests/conftest.py
```diff
@@ -135,3 +137,35 @@ def add(gameobj: Game, n: int, team: int=None):
...
+    ```
+    with Benchmark() as b:
+        time.sleep(1)
+
+    b.elapsed()
+    ```"""
+
+    def __init__(self):
+        self.start = None
+        self.end = None
```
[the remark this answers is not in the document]
Askaholic: I guess it's just habit, initializing members to `None` in the constructor. I didn't think about errors at all when I wrote this.

tests/unit_tests/test_matchmaker_queue_algorithm.py
```diff
@@ -1,63 +1,190 @@
...
+    for search, neighbors in graph.items():
+        for other, quality in neighbors:
+            assert (search, quality) in graph[other]
+
+
+@pytest.mark.parametrize("build_func", (
+    algorithm._MatchingGraph.build_full,
+    algorithm._MatchingGraph.build_fast
+))
+@given(searches=st_searches_list(max_players=2))
```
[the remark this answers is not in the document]
Askaholic: Yes exactly, then you would use `@example()`.

PR #532, opened 2020-02-09
Asyncio drain fix 2

Askaholic:
Catching connection errors in `QDatastreamProtocol` and transforming them into a consistent error type. This error can then be handled by the caller if desired, and the connection will be cleaned up by `ServerContext`.

The key to fixing `socket.send() raised exception` was to check `writer.transport.is_closing()` to see if the transport was still connected.

### Backpressure Handling
Using a coroutine based interface for `send_message` allows us to easily handle backpressure appropriately. For more information on the topic I can recommend https://vorpus.org/blog/some-thoughts-on-asynchronous-api-design-in-a-post-asyncawait-world/#bug-1-backpressure. 

However, this isn't quite sufficient for us as the server doesn't use a pure `request/response` architecture. Sometimes we need to broadcast state updates to players without them asking for it. In this case we need to check if the client is still actively reading our messages to avoid broadcasted messages from filling up our write buffers indefinitely.

In this PR, the server will kill the connection of a client if the client does not read a broadcast within 10 seconds and the `asyncio` buffer has filled up to more than 4096 bytes. These numbers can be adjusted via config parameters.


tests/integration_tests/test_load.py
```diff
@@ -0,0 +1,180 @@
...
+@pytest.mark.slow
+@fast_forward(300)
+async def test_game_info_broadcast_on_connection_error(
+    event_loop, lobby_server, tmp_user, ladder_service, game_service, caplog
+):
+    """
+    Causes connection errors in `do_report_dirties` which in turn will cause
+    closed games not to be cleaned up if the errors aren't handled properly.
+    """
+    # This test causes way to much logging output otherwise
```
cleborys:
```suggestion
    # This test causes way too much logging output otherwise
```

tests/integration_tests/test_login.py
```diff
@@ -0,0 +1,103 @@
+import pytest
+
+from .conftest import (
+    connect_and_sign_in, connect_client, perform_login, read_until_command
+)
+
+pytestmark = pytest.mark.asyncio
```
cleborys: Is this used anywhere?

tests/integration_tests/test_matchmaker.py
```diff
@@ -6,8 +6,20 @@
 pytestmark = pytest.mark.asyncio
```
cleborys: Is this used anywhere?

tests/integration_tests/test_servercontext.py
```diff
@@ -1,11 +1,12 @@
...
 from unittest import mock
 
 import pytest
 from asynctest import CoroutineMock, exhaust_callbacks
 from server import ServerContext, fake_statsd
 from server.lobbyconnection import LobbyConnection
-from server.protocol import QDataStreamProtocol
+from server.protocol import DisconnectedError, QDataStreamProtocol
 
 pytestmark = pytest.mark.asyncio
```
cleborys: I must be missing something. Does this magically apply to every test?

server/gameconnection.py
```diff
@@ -100,38 +112,49 @@ def player(self, val: Player):
...
+            if self._state is GameConnectionState.ENDED:
+                # We aborted while trying to connect
+                return
+
+            self._state = GameConnectionState.CONNECTED_TO_HOST
+
+            try:
                 self.game.add_game_connection(self)
+            except GameError as e:
+                await self.abort(f"GameError while joining %s: %s", self.game.id, e)
```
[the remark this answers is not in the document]
Askaholic: Oh yea, that looks like a mistake

server/gameconnection.py
```diff
@@ -143,16 +166,20 @@ def player(self, val: Player):
...
-            offer=False
-        )
+        if peer:
+            await self.send_ConnectToPeer(
+                player_name=peer.player.login,
+                player_uid=peer.player.id,
+                offer=True
+            )
+
+        if peer:
```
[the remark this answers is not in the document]
Askaholic: Hmm, now that I think about it I'm actually not sure how passing weak references to functions interact. 

The intent is that we need to check the existence of the reference before using it, otherwise for example `peer.player` could throw a `NoneType` error. And it is also the case that whenever you `await`, you need to re-check all of your weak references again since they could have been deleted and cleaned up by the GC since your first check. Now the only thing I'm not sure is how this interaction works when your weak reference is passed as a function argument. I would assume that the function argument its self would still be a weak reference and you'd have to check it every time you `await`.

server/gameconnection.py
```diff
@@ -168,12 +195,11 @@ def player(self, val: Player):
                 "Unrecognized command %s: %s from player %s",
                 command, args, self.player
             )
-        except (TypeError, ValueError) as e:
-            self._logger.exception("Bad command arguments: %s", e)
+        except (TypeError, ValueError):
+            self._logger.exception("Bad command arguments: %s")
```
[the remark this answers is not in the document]
Askaholic: No, I just need to delete the `%s` part. The `exception` method already logs the stack trace so it's unnecessary to add the exception to the message string.

server/gameconnection.py
```diff
@@ -424,15 +458,11 @@ def player(self, val: Player):
             await self._handle_idle_state()
 
         elif state == 'Lobby':
-            # The game is initialized and awaiting commands
-            # At this point, it is listening locally on the
-            # port we told it to (self.player.game_port)
-            # We schedule an async task to determine their connectivity
-            # and respond appropriately
+            # TODO: Do we still need to schedule with `ensure_future`?
```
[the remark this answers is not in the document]
Askaholic: I would think it's been unneeded since we switched to ICE, but IDK the original intent.

server/lobbyconnection.py
```diff
@@ -324,7 +326,9 @@ def on_connection_made(self, protocol: QDataStreamProtocol, peername: Address):
                                 raise ClientError('Your ban attempt upset the database: {}'.format(e))
                     else:
                         self._logger.warning('Administrative action: %s closed client for %s', self.player, player)
-                    await player.lobby_connection.kick()
+                    if player.lobby_connection:
```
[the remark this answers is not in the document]
Askaholic: Yea, I guess I should add that explicitly since I'm trying to follow the example in the weakref docs.
https://docs.python.org/3.5/library/weakref.html#weak-reference-objects

tests/integration_tests/test_login.py
Askaholic: Yes, see the `Markers` section: https://pypi.org/project/pytest-asyncio

Askaholic:
> I am lacking the bigger picture, so all my comments are very "local", but this is the best I feel I can do :)

That's perfect. Understanding the big picture of these changes is kindof impossible anyways since they were mostly a result of a lot of ad-hoc testing done on the test server to try and replicate difficult to produce bugs.

tests/integration_tests/test_load.py
```diff
@@ -0,0 +1,162 @@
...
+    # cleaned up.
+    assert len(game_service.all_games) == 0
+
+
+@fast_forward(30)
+async def test_backpressure_handling(lobby_server, caplog):
+    # TRACE will be spammed with thousands of messages
+    caplog.set_level(logging.DEBUG)
+
+    _, _, proto = await connect_and_sign_in(
```
norraxx: what does this `proto` mean?

Askaholic: It's a `QDataStreamProtocol` which is connected to the server and logged in as the user `test`

server/ladder_service.py
```diff
@@ -125,64 +151,77 @@ def _cancel_existing_searches(self, initiator: Player) -> List[Search]:
...
-        self._logger.debug("Ladder game launched successfully")
+            self._logger.debug("Ladder game launched successfully")
+        except Exception:
+            self._logger.exception("Failed to start ladder game!")
+            msg = {"command": "game_launch_cancelled"}
+            with contextlib.suppress(DisconnectedError):
+                await asyncio.gather(
+                    host.send_message(msg),
+                    guest.send_message(msg)
+                )
```
norraxx: I'm thinking about that big 60 lines of code inner content, that is in try. It's big and it's not readable much, i know, that there is a lot of small operations. Maybe we could extract it to separate function, to see the final logic: 
```python3
self._logger.debug("Starting ladder game: %s", game)
            await host.lobby_connection.launch_game(game, is_host=True, use_map=mapname)
            try:
                hosted = await game.await_hosted()
                if not hosted:
                    raise TimeoutError("Host left lobby")
            except TimeoutError:
                msg = {"command": "game_launch_cancelled"}
                with contextlib.suppress(DisconnectedError):
                    # Sending to host might not be needed, but we do it anyways
                    await asyncio.gather(
                        host.send_message(msg),
                        guest.send_message(msg)
                    )
                # TODO: Uncomment this line once the client supports `game_launch_cancelled`.
                # Until then, returning here will cause the client to think it is
                # searching for ladder, even though the server has already removed it
                # from the queue.

                # return
                self._logger.debug("Ladder game failed to launch due to a timeout")

            # TODO: Graceful handling of NoneType errors due to disconnect
            await guest.lobby_connection.launch_game(
                game, is_host=False, use_map=mapname
            )
```

Like: `prepare_game`, `connect_users` (`host` + `guest`)
\+ I think, we can get rid of inner `except TimeoutError:`. This is a little bit ridiculous to have variable `hosted` and then raise, catch, and then again... this is useless load for our server. i know, that `python way`, etc.. But generating exception still has to collect whole stack-trace, and gather other information, run raise mechanics etc.. instead of boolean check...

```python3
            hosted = await game.await_hosted()
            if not hosted:
                ....
```

server/ladder_service.py
```diff
@@ -125,64 +151,77 @@ def _cancel_existing_searches(self, initiator: Player) -> List[Search]:
...
+                # Until then, returning here will cause the client to think it is
+                # searching for ladder, even though the server has already removed it
+                # from the queue.
+
+                # return
+                self._logger.debug("Ladder game failed to launch due to a timeout")
+
+            # TODO: Graceful handling of NoneType errors due to disconnect
+            await guest.lobby_connection.launch_game(
+                game, is_host=False, use_map=mapname
```
norraxx: Do we need to launch game for `guest` if game is not hosted?

server/lobbyconnection.py
```diff
@@ -586,11 +590,18 @@ def on_connection_made(self, protocol: QDataStreamProtocol, peername: Address):
         old_player = self.player_service.get_player(self.player.id)
         if old_player:
             self._logger.debug("player {} already signed in: {}".format(self.player.id, old_player))
-            if old_player.lobby_connection:
-                await old_player.lobby_connection.send_warning("You have been signed out because you signed in elsewhere.", fatal=True)
-                old_player.lobby_connection.game_connection = None
-                old_player.lobby_connection.player = None
-                self._logger.debug("Removing previous game_connection and player reference of player {} in hope on_connection_lost() wouldn't drop her out of the game".format(self.player.id))
+            if old_player.lobby_connection is not None:
```
norraxx: Can be `old_player.lobby_connection` as `0` or `False`? If not, i'd use `if old_player.lobby_connection`, no need to check for None explicit, it's easier to read

server/matchmaker/matchmaker_queue.py
```diff
@@ -93,7 +93,7 @@ def __init__(
     def find_matches(self) -> None:
         self._logger.info("Searching for matches: %s", self.queue_name)
 
-        # Call self.match on all matches and filter out the ones that were canceled
+        # Call self.match on all matches and filter out the ones that were cancelled
```
norraxx: WOW! Never thought, that I will review american vs british grammar :-), ofc British ftw :-D :+1:

server/players.py
```diff
@@ -125,6 +127,17 @@ def game_connection(self, value):
     def game_connection(self):
         self._game_connection = lambda: None
 
+    async def send_message(self, message):
```
norraxx: :+1:

server/protocol/qdatastreamprotocol.py
```diff
@@ -24,11 +30,17 @@ def __init__(self, reader: StreamReader, writer: StreamWriter):
         """
         self.reader = reader
         self.writer = writer
+        self.writer.transport.set_write_buffer_limits(high=0)
 
         # drain() cannot be called concurrently by multiple coroutines:
         # http://bugs.python.org/issue29930.
         self._drain_lock = asyncio.Lock()
 
+    @property
```
norraxx: I wish we won't use `@property` at all. Reason - property is hiding some "behavior" which is not cool. Property are used to have the way for backward code compatibility, when you have some variable, but in fact you need logic and you don't want to change api.

server/protocol/qdatastreamprotocol.py
```diff
@@ -52,19 +64,6 @@ def read_qstring(buffer: bytes, pos: int=0) -> Tuple[int, str]:
...
-        Read a serialized 32-bit integer from the given buffer at given position
-
-        :return (int, int): (buffer_pos, int)
-        """
-        chunk = buffer[pos:pos + 4]
-        assert len(chunk) == 4
-
-        (num, ) = struct.unpack('!i', chunk)
-        return pos + 4, num
-
```
norraxx: :+1:

server/protocol/qdatastreamprotocol.py
```diff
@@ -24,11 +30,17 @@ def __init__(self, reader: StreamReader, writer: StreamWriter):
         """
         self.reader = reader
         self.writer = writer
+        self.writer.transport.set_write_buffer_limits(high=0)
```
norraxx: please, add here in comment reason why do you set such value

server/protocol/qdatastreamprotocol.py
```diff
@@ -24,11 +30,17 @@ def __init__(self, reader: StreamReader, writer: StreamWriter):
...
         # http://bugs.python.org/issue29930.
         self._drain_lock = asyncio.Lock()
 
+    @property
+    def connected(self):
+        # TODO: In python 3.7 and above call writer.is_closing() directly
+        return not self.writer.transport.is_closing()
+
     @staticmethod
     def read_qstring(buffer: bytes, pos: int=0) -> Tuple[int, str]:
```
norraxx: Formatting `int = 0`

server/protocol/qdatastreamprotocol.py
```diff
@@ -127,31 +126,60 @@ def pack_message(*args: str) -> bytes:
...
         :return:
         """
         self.writer.close()
 
+    def abort(self):
+        """
+        Close writer stream immediately discarding the buffer contents
+        :return:
+        """
+        self.writer.transport.abort()
```
norraxx: Do I understand right, that HERE we don't care about anything, we want just close the writer?

But I don't understand why? I see the catch on timeout, but why do we have to call internal transport function on writer to force abort the connection?

server/protocol/qdatastreamprotocol.py
```diff
@@ -127,31 +126,60 @@ def pack_message(*args: str) -> bytes:
...
+
+
```
norraxx: remove one line

server/protocol/qdatastreamprotocol.py
```diff
@@ -127,31 +126,60 @@ def pack_message(*args: str) -> bytes:
...
+        Await the write buffer to empty.
+        See StreamWriter.drain()
+
+        :raises: DisconnectedError if the client disconnects while waiting for
+        the write buffer to empty.
+        """
+        # NOTE: This sleep is needed in python versions <= 3.6
+        # https://github.com/aio-libs/aioftp/issues/7
+        await asyncio.sleep(0)
+        async with self._drain_lock:
```
norraxx: When I see this two lines, I must say, that i'm not happy from `asyncio`... :disappointed: 
(or we're doing something wrong :see_no_evil: )

server/servercontext.py
```diff
@@ -53,24 +54,59 @@ def close(self):
...
 
+    async def broadcast(self, message, validate_fn=lambda a: True):
+        await self._do_broadcast(
+            validate_fn,
+            QDataStreamProtocol.send_message,
+            message
+        )
+        self._logger.log(TRACE, "]]: %s", message)
+
     async def broadcast_raw(self, message, validate_fn=lambda a: True):
```
norraxx: do we even need that function? It's used only for `ping` every 45 secs

server/servercontext.py
```diff
@@ -53,24 +54,59 @@ def close(self):
...
+        await self._do_broadcast(
+            validate_fn,
+            QDataStreamProtocol.send_raw,
+            message
+        )
+
+    async def _do_broadcast(self, validate_fn, send_fn, message):
         server.stats.incr('server.broadcasts')
+
+        async def broadcast_with_stall_handling(proto):
```
norraxx: Typing + we can move that inner function out of the scope of `_do_broadcast`.
Inner functions are bad practice + hard to test, etc...

server/servercontext.py
```diff
@@ -53,24 +54,59 @@ def close(self):
...
+            try:
+                await asyncio.wait_for(
+                    send_fn(proto, message),
+                    timeout=CLIENT_STALL_TIME
+                )
+            except asyncio.TimeoutError:
+                buffer_size = len(proto.writer.transport._buffer)
```
norraxx: we touch private stuff of the transport `._buffer`, one day it will could have problems. Don't we want `get_write_buffer_size()` ?

server/ladder_service.py
Askaholic: Yes, let me paste my comment about it here:
```
# TODO: Uncomment this line once the client supports `game_launch_cancelled`.
# Until then, returning here will cause the client to think it is
# searching for ladder, even though the server has already removed it
# from the queue.
```

This is how the client expects messages to happen (and how they would have been handled before). The other option is of course to purposefully introduce a bug to force the client guys to fix it properly, but that's probably not the best idea.

server/lobbyconnection.py
Askaholic: I'm simply following the example in the `weakref` docs. They check for `None` explicitly, so that's what I'm doing too.

https://docs.python.org/3/library/weakref.html#weak-reference-objects

server/protocol/qdatastreamprotocol.py
Askaholic: Yes so this is called when we've detected unresponsive clients. We don't want to call `close` because that will still wait for the buffer to drain (which we expect isn't going to happen anytime soon in this case). Sadly the stream API does not expose `abort` directly, so we have to call it on the transport object manually.

server/protocol/qdatastreamprotocol.py
Askaholic: It's an asyncio issue. `drain` doesn't support multiple waiters so we have to use a lock. This is actually the solution from the `websockets` library.

server/servercontext.py
Askaholic: Yes, somehow I missed that function.

server/ladder_service.py
```diff
@@ -55,17 +71,26 @@ def __init__(self, database: FAFDatabase, games_service: GameService):
 
         asyncio.ensure_future(self.queues[queue_name].search(search))
```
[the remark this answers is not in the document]
Askaholic: As far as I can tell `MatchmakerQueue::search()` parameters and docs are all correct. We need to use `ensure_future` because `search` will block until a match is found or the search is cancelled, and we need to keep processing messages while this happens.

server/ladder_service.py
Askaholic: Problem is that `await_hosted()` can raise `TimeoutError` too. This way folds the error handling neatly into one case, instead of having to write the code twice.

server/servercontext.py
Askaholic: Yes, we need it for precisely this reason. This is also why I refactored the main logic into `_do_broadcast` so that I this function can be a single function call.

norraxx:
Imho, we should squash

Askaholic:
I will squash when I merge

Issue #482, opened 2019-07-28
Revise protocol documentation


Askaholic:
We really need a dedicated server protocol document where we write down what messages are sent by the client, and what responses come from the server. That way the client developers can have something to reference without needing to read the server code. This would also push us toward defined protocol versions where we can explicitly say "version 1.0 has exactly these commands".

Ideally we would refactor the server so that much of this documentation could be generated automatically from doc strings.

KaukaHan:
how about implementing the hole protocol into a webassembly module, witch can be shared between the server and client ?

KaukaHan:
https://webassembly.org/
https://wasmer.io/
tldr: webassembly and wasmer would allow to compile into a platform indepenent, efficient, sandboxed, binary format witch can be embed into other software. the biggest problem i think is, you need a compiled language to compile into wasm, like c/c++ or rust. might be overkill therefor.

PR #533, opened 2020-02-10
[Postponed until after TMM] Protocol Documentation

cleborys:
Makes explicit what messages the server expects from clients and generates documentation accordingly, see Issue #482.

Parsing logic has been moved from `LobbyConnection.on_message_received`, partially to `clientmessages.MessageParser` and partially to the factory methods of Messages associated with a given command.

To generate documentation, install `pdoc3` and run it from a pipenv shell in the server root:
```
$ pipenv shell
$ pdoc3 --html server
```
The documentation target should be the `server` folder, even if you are only interested in `server/lobbyconnection.py`, to get all imports right.
Pdoc will produce a folder named `html`, to see the protocol documentation open `html/server/lobbyconnection.html`, where every displayed method of `LobbyConnection` reflects one message format (usually uniquely identified by its "command" entry).

Information about what the server does in reaction to a method is stored in the docstring of the handling message (e.g. `LobbyConnection.command_ping`), information about the expected format is stored in the message class doing the parsing (e.g. `clientmessages.PingMessage`). Both are concatenated, along with information on whether the client needs to be authenticated, and overwrite the pdoc docstring by creating an appropriate entry in the `__pdoc__` dict of `lobbymessages`.

The original intent was to document he individual message classes, but since `lobbyconnection` depends on `clientmessages` and the `handler` routing happens in `lobbyconnection`, it is easier to pull the `clientmessages` docstrings into `lobbyconnection` than the other way around.
Let me know if you have thoughts or considerations on generating the documentation differently.

As far as running different protocol versions in parallel goes, I would hope that the `clientmessages` module is modular enough, as long as the signatures of the handling `LobbyConnection` methods don't change. ("Signature" here meaning what fields are expected in the parsed message.)

While `clientmessages` has its own unit tests to document the expected parsing behavior, I am not yet confident that `lobbyconnection` is sufficiently covered, especially regarding error handling.

### DIFF
--- a/server/clientmessages.py
+++ b/server/clientmessages.py
@@ -79,7 +79,7 @@ class LobbyTargetMessage(MessageFromClient):
     """
 
     @staticmethod
-    def build(lobbyconnection, message):
+    def parse(message):
         command = MessageParser.parse_command(message)
         return COMMAND_TO_CLASS[command].build(message)
 
@@ -92,36 +92,34 @@ class ConnectivityTargetMessage(MessageFromClient):
     """
 
     @staticmethod
-    async def __init__(self, lobbyconnection=None, message=None):
-        lobbyconnection.register_connectivity_test()
-
-    async def handle(self):
-        raise ClientError(
-            f"Your client version is no longer supported."
-            f"Please update to the newest version: https://faforever.com"
-        )
+    def build(message):
+        raise NotImplementedError("TODO")
 
 
-class GameTargetMessage(MessageFromClient):
+
+class GameTargetMessage(NamedTuple, MessageFromClient):
     """
     Common base class for messages with `target` entry `'game'`.
 
     To be handled by a `GameConnection`.
     """
+    command: str
+    args: list
 
-    @staticmethod
-    async def build(lobbyconnection=None, message=None):
-        # need to make a distinction on command, arguments
-        # should fail if passed lobbyconnection does not have a game_connection
-        # message entry 'args' needs to be passed along
-        # previous code:
-        #   if target == 'game':
-        #       if not self.game_connection:
-        #           return
-        #
-        #       await self.game_connection.handle_action(cmd, message.get('args', []))
-        #       return
-        raise NotImplementedError("TODO")
+    @classmethod
+    def build(cls, message):
+
+        if "command" not in message:
+            raise MessageParsingError(
+                f"Message with target 'game' must contain field 'command'. "
+                f"Offending message: {message}."
+            )
+
+        command = message["command"]
+        arguments = message.get("args", [])
+
+
+        return cls(command, arguments)
 
 
 class PingMessage(LobbyTargetMessage):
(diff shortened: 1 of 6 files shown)


Geosearchef:
I really like this, I'm gonna go crazy rebasing the party system though.

server/lobbyconnection.py
```diff
@@ -36,6 +37,34 @@
...
+
+
+__pdoc__ = {}
+COMMAND_HANDLERS = {}
+
+def handler(handling_class):
+    def decorator(function: Callable) -> Callable:
+        COMMAND_HANDLERS[handling_class] = function
+
+        func_string = function.__doc__ or "[Method description missing]"
```
Askaholic: This is also visually unpleasing. Maybe adding italics could make it more clear.

```suggestion
        func_string = function.__doc__ or "*[Method description missing]*"
```

server/lobbyconnection.py
```diff
@@ -161,30 +208,54 @@ def on_connection_made(self, protocol: QDataStreamProtocol, peername: Address):
...
             await self.abort("Error processing command")
 
-    async def command_ping(self, msg):
+    @handler(GameTargetMessage)
+    async def handle_game_target_message(self, parsed_message):
+        if not self.game_connection:
+            return
+
+        await self.game_connection.handle_action(
+                parsed_message.command,
```
Askaholic: extra indent

server/clientmessages.py
```diff
@@ -0,0 +1,660 @@
...
+    def build(cls, message):
+        if "friend" in message:
+            adding_a_friend = True
+            id_to_add = message["friend"]
+        elif "foe" in message:
+            adding_a_friend = False
+            id_to_add = message["foe"]
+        else:
+            raise MessageParsingError(
+                f"Message to add friend or foe "
```
Askaholic: btw you don't need the `f`'s here

server/clientmessages.py
```diff
@@ -0,0 +1,660 @@
...
+class ActionField(Enum):
+    """
+    Valid entries for the `'action'` field in a message from the client.
+    """
+
+    list_avatar = 1
+    select = 2
+
+
+class CommandField(Enum):
```
Askaholic: Need to figure out how we can get rid of this class and generate the string to class name mapping automatically. Maybe some metaclass magic would work so that you only need to give the `command` attribute a default value.

Brutus5000:
As I'm working on a separate client protocol implementation, one thing I'd like to add: A description of messages won't help another dev at all, if the message flow is undocumented.
Like how to perform a login. What message is a response to which? Which messages are sent in a fixed interval?

The optional and required fields are nice, but less important imho.

Geosearchef:
This PR is awesome, but:

TMM is currently split over multiple branches and modifies large parts of ladder code as well as 
changing parts of the protocol.
If this gets merged it's going to kill TMM. (We're gonna be in rebase hell)

server/clientmessages.py
cleborys: Not sure how that survived :)

cleborys:
@Geosearchef Well, I obviously don't want to kill TMM, so I'll wait until that is done.

@Brutus5000 That sounds reasonable, but I am unsure where to put that documentation in the code. The nice thing about signature and behavior is of course, that they can be documented right where they are implemented and that is why we are pulling the docstrings from the parser and the handlers. We can of course add our current expectations of message flow at one place or another, but the documentation is less likely to be updated the expectation it ever changes. Unless you have a better suggestion, I will see what I can piece together and add that after rebasing onto a (hopefully finished) TMM at some point in the future :)

server/clientmessages.py
cleborys: I think that would be straightforward, I went for this because I wanted `pdoc` to also generate a list of all possible commands. That probably also works with "metaclass magic" :)

server/lobbyconnection.py
```diff
@@ -88,6 +117,36 @@ def __init__(
 
         self._logger.debug("LobbyConnection initialized")
 
+    class Decorators:
```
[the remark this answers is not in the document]
cleborys: It cannot be an instance method, since it is called without an instance, but it cannot be a static method either (see below). I didn't want to declare it outside the class, since it knows too much about `LobbyConnection` (although it's maybe unlikely that anything will ever inherit from `LobbyConnection`?).
An inner class is one of the workarounds recommended here:
https://medium.com/@vadimpushtaev/decorator-inside-python-class-1e74d23107f6

cleborys:
Closing, since I think there has been enough refactoring that this PR isn't worth a rebase. It will still be linked to the issue so this discussion won't be lost.

PR #538, opened 2020-03-01
Prometheus metrics

cleborys:
Replaces `aiomeasures` with the python prometheus client,
see issue #474. 
Now called "metrics", since the previous name "stats" is also used for the achievements module.

I did not yet include all changes suggested there, but appreciate feedback on the general setup.
I can provide the accompanying grafana dashboard.

Items on the issue:
- [x] Use a single "client connected" metric
- [x] Tag it with user_agent version as in: https://github.com/FAForever/server/pull/260
- [x] Remove `connections.aborted`. It doesn't seem to provide any useful info
- [x] Figure out a better data format for ladder match quality
- [x] Add a guage or counter for the number of active games
- [x] Player count is only updated when someone logs in, not when someone disconnects

Things to be checked:
- [x] negative user agent numbers and reset to zero after server reboot.
- [x] Do Gauges reset on server restart? (e.g. logged in users, user_agents)


server/protocol/qdatastreamprotocol.py
```diff
@@ -151,7 +152,7 @@ def close(self):
             await self.writer.drain()
 
     async def send_raw(self, data):
-        server.stats.incr('server.sent_messages')
+        metrics.sent_messages.labels("raw").inc()
```
Askaholic: I don't think there's a reason to distinguish between these two. Especially since in my asyncio fix PR I'm refactoring `send_message` to call `send_raw`.

server/matchmaker/matchmaker_queue.py
```diff
@@ -4,15 +4,33 @@
...
 import server
+import server.metrics as metrics
 
 from ..decorators import with_logger
 from .algorithm import make_matches
 from .pop_timer import PopTimer
 from .search import Match, Search
 
 
+def timed_async_search(func):
```
Askaholic: Why does this need to be a decorator?

We also might want to know whether the search ended because it found a match or because it was cancelled. A nice way I can think of for doing this would be with a context manager, especially since context managers can see if an exception was thrown. https://docs.python.org/3/reference/datamodel.html#object.__exit__

Then the search code might look something like:
```python
try:
    with metrics.matchmaker_search_timer():
        # Wait for search to complete
except CancelledError:
    pass
finally:
    # Do the cleanup
```

and the context manager would be able to detect if `CancelledError` was thrown or not and label the observation appropriately.

server/matchmaker/matchmaker_queue.py
```diff
@@ -4,15 +4,33 @@
...
+    @functools.wraps(func)
+    async def wrapper(*args, **kwargs):
+        queue_name = args[0].queue_name
+        metric = metrics.matchmaker_searches.labels(queue_name)
+        start_time = time.perf_counter()
```
Askaholic: Not sure if `perf_counter` is the right function to use here, it's meant for "short" time durations whatever that means. Probably `time` or maybe even `monotonic` would be better since it's technically possible for players to be queued for hours: https://docs.python.org/3/library/time.html#time.monotonic

server/lobbyconnection.py
```diff
@@ -427,8 +428,9 @@ def on_connection_made(self, protocol: QDataStreamProtocol, peername: Address):
 
         self.user_agent = message.get('user_agent')
         version = message.get('version')
-        server.stats.gauge('user.agents.None', -1, delta=True)
-        server.stats.gauge('user.agents.{}'.format(self.user_agent), 1, delta=True)
+        metrics.user_connections.labels("None").dec()
```
Askaholic: BTW I'm not sure how correctly this code works. Right now these numbers are always wrong, especially because the gauges don't seem to reset to 0 correctly when the server is stopped. So if you can think of a better way to track user agents, then go for it.

cleborys: Previously, they seem to have been set to zero on server start manually here:
https://github.com/FAForever/server/blob/afacbce94e4303d1f6c2e148652985c4769aafdb/server/servercontext.py#L28-L33

I assume that was meant to counteract them not being set to zero on server shutdown? However, it seems weird that they don't start out at zero in a new python process, anyways.
We should check again after testing `prometheus_client` on the test sever.

server/matchmaker/matchmaker_queue.py
```diff
@@ -52,20 +70,26 @@ def __init__(
...
-            server.stats.gauge(f"matchmaker.queue.{self.queue_name}.matches", len(self._matches))
-            if self._matches:
-                server.stats.gauge(
-                    f"matchmaker.queue.{self.queue_name}.quality",
-                    mean(map(lambda m: m[0].quality_with(m[1]), self._matches))
+
+            number_of_matches = len(self._matches)
+            metrics.matches.labels(self.queue_name).set(number_of_matches)
+
+            #TODO: Move this into algorithm, then don't need to recalculate quality_with?
```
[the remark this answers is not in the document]
cleborys: I will bookmark that for a later PR, I'd prefer not to touch the matchmaker implementation in this one if not strictly necessary.

server/lobbyconnection.py
```diff
@@ -531,8 +533,8 @@ def on_connection_made(self, protocol: QDataStreamProtocol, peername: Address):
 
         async with self._db.acquire() as conn:
             player_id, login, steamid = await self.check_user_login(conn, login, password)
-            server.stats.incr('user.logins', tags={'status': 'success'})
-            server.stats.gauge('users.online', len(self.player_service))
+            metrics.user_logins.labels("success").inc()
+            metrics.users_online.set(len(self.player_service))
```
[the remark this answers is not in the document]
cleborys: Ah very good. I had been thinking about putting it into the `on_connection_lost` and `abort` methods, which is clearly inferior.

server/lobbyconnection.py
cleborys: Looking at this again, it seems that `user_agents.None` is decreased, every time the client sends the `ask_session` command, and I assume we cannot guarantee that this is only sent once and I can reproduce this locally.

Then I know what to do :)

server/matchmaker/matchmaker_queue.py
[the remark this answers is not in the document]
Askaholic: Yea definitely. I'm just posing it as a potential solution.

server/lobbyconnection.py
Askaholic: I just recently added the setting to 0 stuff, but it didn't seem to work at all when we deployed it. I didn't really test it, I just threw it in there hoping it would solve the problem.

server/matchmaker/matchmaker_queue.py
```diff
@@ -52,14 +75,19 @@ def __init__(
...
+            # Probably not a major bottleneck though.
+            for match in self._matches:
```
Askaholic: This could be even slicker with some unpacking (parenthesis optional)
```suggestion
            for (s1, s2) in self._matches:
```

server/metrics.py
```diff
@@ -0,0 +1,75 @@
...
+
+games_hosted = Counter(
+    "server_game_hosted_games_total", "Total number of games hosted", ["game_mode"]
+)
+
+live_games = Gauge(
+    "server_game_active_games_total", "Number of currently active games", ["game_mode"]
+)
+
+games_in_service = Gauge(
```
Askaholic: Maybe call it `all_games` or something like that. At least lets keep it consistent with the others. So we'd have `live_games`, `all_games` and `hosted_games`, or the other way around.

server/game_service.py
```diff
@@ -155,9 +157,22 @@ def create_game(
 
     @property
     def live_games(self) -> List[Game]:
-        return [game for game in self.games.values()
+        live_games =  [game for game in self._games.values() 
                 if game.state == GameState.LIVE]
 
+        #FIXME in case the live_games gauge ever gets out of sync with the
```
Askaholic: Actually as far as I can tell, this function is never called (and neither is `pending_games`). Although usually I think it would be safer to update the metric with a `set` rather than an `inc/dec` since it won't be able to go out of sync that way.

server/metrics.py
Askaholic: On second though, maybe they should all be one Gauge with different labels. Having a Counter for hosted games is kind of dumb anyways.

server/matchmaker/matchmaker_queue.py
cleborys: Yes, that is indeed nicer.

server/game_service.py
cleborys: I will rework it such that it only ever calls `set`, at the expense of scanning through the list of active games a lot.
Since it is typically short (<400?) I assume that is acceptably fast.
Counting 1000 items with 25 distinct values and then setting a prometheus gauge to the 25 counts takes 220 microseconds on my laptop. We would be doing that whenever a game is added, removed, or changes its `GameState`.

Askaholic: Yea that's the downside. I'm not sure what the right call is. Game lists are short, but games change pretty often. Maybe we could update the stat only in `__init__.py::do_report_dirties`. This triggers once per second which is plenty of resolution.

cleborys: Moving it to `do_report_dirties` sounds great to me!

server/game_service.py
```diff
@@ -145,18 +148,32 @@ def create_game(
...
         game.password = password
 
         self.mark_dirty(game)
         return game
 
+    def update_active_game_metrics(self):
+        # reset the count of every key to zero, 
+        # without removing it from Counter.elements()
+        for key in self._metrics_counter:
+            self._metrics_counter[key] = 0
```
Askaholic: Kindof annoying that it has to be done this way, but it makes sense. If there were a way to do this without the member variable that would be a lot better. Especially since I'm not sure what will happen on server restarts. The game list will be cleared, but will the guage be reset properly?

cleborys: The `Gauge` object maintains all labels in a private dictionary `gauge._metrics`, see here:
https://github.com/prometheus/client_python/blob/2d85f4ae9ba6f16c87e40ce28e0b7111c9258c5a/prometheus_client/metrics.py#L110-L113

So if you are OK with depending on internals of the prometheus_client module, we can use that to get all labels used so far, instead of the `Counter`.

Another alternative is, of course, to only keep track of a fixed number of `game_mode`s (e.g. "faf", "ladder1v1", and "other").
The latter would also help with properly resetting at server restart.

Askaholic: I guess we could use `GameService.featured_mods` which is the list of available featured mods from the database. This would be the same list used by the clients.

cleborys:
I am not sure what to do about `Pipfile.lock`. I added the new package, but  by installing it, I also updated the lockfile to new versions for a ton of other packages. Do we want that, or do we want to stick with the older versions? In the latter case I would fix version numbers in the `Pipfile` and then make the switches to newer versions consciously whenever we see fit.

Askaholic:
I think updating the lock file is fine. I don't see a reason why we need to continue to use outdated packages if new ones are available. 

For a bigger project, maybe it would be useful to update the lock file only in it's own PR to avoid annoying conflicts, but I don't think we have enough churn on this repo for that to be a problem.

server/metrics.py
```diff
@@ -0,0 +1,98 @@
...
+
+
+# =====
+# Games
+# =====
+active_games = Gauge(
+    "server_game_active_games_total",
+    "Number of currently active games."
+    "Includes games in lobby, games currently running, and games that ended"
+    "but are still in the game_service.",
```
Askaholic: This doesn't really matter since these mostly just act as comments, but for future reference when you're doing multiline strings like this you need to remember not to ignore spaces at the edge of your strings, otherwise the words before and after a line break will be mashed together.

```suggestion
    "Number of currently active games. "
    "Includes games in lobby, games currently running, and games that ended "
    "but are still in the game_service.",
```

server/protocol/qdatastreamprotocol.py
```diff
@@ -157,11 +158,13 @@ def abort(self):
                 raise DisconnectedError("Protocol connection lost!") from e
 
     async def send_message(self, message: dict):
+        metrics.sent_messages.inc()
```
Askaholic: `send_raw` already increments the metric, so we don't need to do it here

server/servercontext.py
```diff
@@ -70,8 +67,7 @@ def __contains__(self, connection):
         )
 
     async def _do_broadcast(self, validate_fn, send_fn, message):
-        server.stats.incr('server.broadcasts')
-
+        metrics.server_broadcasts.labels("raw").inc()
```
Askaholic: And now there is both a "raw" and a "regular" broadcast function, so if we're tagging the "raw" ones we should also be tagging the "regular" ones.

cleborys:
Let me know when you can put this on the test server, I should be able to test this on my own and make sure the grafana dashboard works :)

server/config.py
```diff
@@ -15,8 +15,8 @@
 # Optimum values for ladder here, using them for global as well.
 trueskill.setup(mu=1500, sigma=500, beta=240, tau=10, draw_probability=0.10)
 
-STATSD_SERVER = os.getenv('STATSD_SERVER', '127.0.0.1:8125')
-ENABLE_STATSD = os.getenv('ENABLE_STATSD', 'false').lower() == 'true'
+METRICS_PORT = int(os.getenv('METRICS_PORT', 8125))
```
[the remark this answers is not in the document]
Askaholic: Also we should delete `fake_statsd.py`

Askaholic:
Closes #260

PR #540, opened 2020-03-07
Ice unit tests

Askaholic:
Adding some unit tests for ICE (which @Geosearchef promised me he would write, but never did).


Geosearchef:
I did? Don't remember that, sry.

Askaholic:
Maybe it was Louvegarde, but someone promised me they would do it and of course never did.

PR #542, opened 2020-03-17
Improve broadcast performance

Askaholic:
Refactor `ServerContext.broadcast` so that it only performs json and QDataStream encoding once per message (as opposed to once per connection per message).

### DIFF
--- a/server/config.py
+++ b/server/config.py
@@ -44,7 +44,7 @@ FORCE_STEAM_LINK = os.getenv('FORCE_STEAM_LINK', 'false').lower() == 'true'
 CLIENT_STALL_TIME = int(os.getenv('CLIENT_STALL_TIME', 10))
 # Maximum number of bytes we will allow a stalled connection to get behind
 # before we terminate their connection.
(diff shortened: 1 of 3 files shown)


server/config.py
```diff
@@ -44,7 +44,7 @@
 CLIENT_STALL_TIME = int(os.getenv('CLIENT_STALL_TIME', 10))
 # Maximum number of bytes we will allow a stalled connection to get behind
 # before we terminate their connection.
-CLIENT_MAX_WRITE_BUFFER_SIZE = int(os.getenv('CLIENT_MAX_WRITE_BUFFER_SIZE', 2**16))
+CLIENT_MAX_WRITE_BUFFER_SIZE = int(os.getenv('CLIENT_MAX_WRITE_BUFFER_SIZE', 2**17))
```
cleborys: How was this related / why was it necessary?

Askaholic: It was terminating more connections than I wanted it to and I realized that I probably should set the max buffer size higher than the default for the high watermark.

PR #543, opened 2020-03-17
Write GameOutcome to db in Game.persist_results.

cleborys:
Now `Game.persist_results` writes the `GameOutcome` for every single player to the new `result` column of `game_player_stats`, according to `Game.outcome` at the time of the call.

Also introduces new `GameOutcome.CONFLICTING`, which is treated as `UNKNOWN` for rating purposes, so this shouldn't change any rating decisions.

### DIFF
--- a/server/games/game_results.py
+++ b/server/games/game_results.py
@@ -91,8 +91,8 @@ class GameResults(Mapping):
             return outcomes.pop()
         elif len(outcomes) > 1:
             self._logger.info(
-                "Multiple outcomes for game %s army %s: %s", self._game_id,
-                army, list(outcomes)
+                "Multiple outcomes for game %s army %s: %s",
+                self._game_id, army, list(outcomes)
             )
             return GameOutcome.CONFLICTING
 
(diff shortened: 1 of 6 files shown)


server/games/game_results.py
```diff
@@ -142,6 +144,11 @@ def victory_only_score(self, army: int):
                 startspot, score = row[0], row[1]
                 # FIXME: Assertion about startspot == army
                 # FIXME: Reporter not retained in database
-                result = GameResult(0, startspot, GameOutcome.UNKNOWN, score)
+                # TODO: I don't understand the comment above
```
Askaholic: Yea I'm not sure what that comment refers to either. As for the `startspot == army` I think that's because the lua code always assumes that to be true, and I guess therefore the server does too, but it isn't checked for or enforced.

tests/unit_tests/test_game.py
```diff
@@ -706,17 +706,18 @@ def custom_game(event_loop, database, game_service, game_stats_service):
...
+async def test_game_get_player_outcome_ignores_unknown_results(
+    game, game_add_players
+):
     game.state = GameState.LOBBY
     players = game_add_players(game, 2)
 
     await game.add_result(0, 0, 'defeat', 0)
     await game.add_result(0, 0, 'score', 0)
-    assert game.get_army_result(players[0]) == GameOutcome.DEFEAT
+    assert game.get_player_outcome(players[0]) == GameOutcome.DEFEAT
```
Askaholic:
```suggestion
    assert game.get_player_outcome(players[0]) is GameOutcome.DEFEAT
```

PR #544, opened 2020-03-17
Bump faf stack version in Travis to 20.2.7

cleborys:
Move all `test-data.sql` into `tests/data/test-data.sql` and use faf-stack version `20.2.7` in Travis.

Necessary for #543

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -14,7 +14,7 @@ before_install:
   - git clone https://github.com/FAForever/faf-stack.git faf-stack
       && pushd faf-stack
       && git checkout v20.2.7
-      && source .env.template
+      && export $(egrep -v '^#' .env.template | xargs)
       && cp -r config.template config
       && ./scripts/init-db.sh
       && popd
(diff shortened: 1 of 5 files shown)


.travis.yml
```diff
@@ -13,12 +13,11 @@ before_install:
   - sudo /etc/init.d/mysql stop
   - git clone https://github.com/FAForever/faf-stack.git faf-stack
       && pushd faf-stack
-      && git checkout v19.8.12
+      && git checkout v20.2.7
+      && export $(egrep -v '^#' .env.template | xargs)
```
Askaholic: What does this line do? Set environment variables?

cleborys: Exactly. It seems to be necessary for `scripts/init-db.sh` now but apparently wasn't before.

Askaholic: I see. Yea it seems I didn't have a `.env` file before, but now I was unable to rebuild the database until I copied it from `.env.template`.

PR #548, opened 2020-03-19
Performance: Don't await broadcasts

Askaholic:
Broadcasting is by far the most expensive single operation that the server does. There is also no good way of handling backpressure in the broadcasts since we can't just stop broadcasting updates to certain players. 

It seems that calling drain for each connection and each message adds significant overhead, so by writing synchronously and not waiting for the messages to be sent we save ourselves a lot of calls to asyncio machinery which ultimately doesn't gain us anything anyways. 

This is how everything was done in v1.2.

### DIFF
--- a/server/__init__.py
+++ b/server/__init__.py
@@ -10,7 +10,6 @@ import logging
 from typing import Optional
 
 from prometheus_client import start_http_server
-import asyncio
 
 from server.db import FAFDatabase
 from . import config as config
@@ -50,14 +49,13 @@ __all__ = (
 
 DIRTY_REPORT_INTERVAL = 1  # Seconds
 stats = None
+logger = logging.getLogger("server")
 
 if config.ENABLE_METRICS:
     start_http_server(config.METRICS_PORT)
 
 
-def encode_message(message: str):
-    # Crazy evil encoding scheme
-    return QDataStreamProtocol.pack_message(message)
+PING_MSG = QDataStreamProtocol.pack_message('PING')
 
 
 def run_lobby_server(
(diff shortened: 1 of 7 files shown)


server/__init__.py
```diff
@@ -83,25 +81,20 @@ def run_lobby_server(
...
 
         if dirty_players:
-            tasks.append(
-                ctx.broadcast({
-                        'command': 'player_info',
-                        'players': [player.to_dict() for player in dirty_players]
-                    },
-                    lambda lobby_conn: lobby_conn.authenticated
-                )
+            ctx.write_broadcast({
```
norraxx: put curly bracket on next line, pls

server/__init__.py
```diff
@@ -83,25 +81,20 @@ def run_lobby_server(
...
-                        'command': 'matchmaker_info',
-                        'queues': [queue.to_dict() for queue in dirty_queues]
-                    },
-                    lambda lobby_conn: lobby_conn.authenticated
-                )
+            ctx.write_broadcast({
+                    'command': 'matchmaker_info',
+                    'queues': [queue.to_dict() for queue in dirty_queues]
+                },
+                lambda lobby_conn: lobby_conn.authenticated
```
norraxx: if we talk about performance, you can move that lambda out of the scope, to not to create this function every time

Askaholic: Well you see I'm basing my performance improvements on actual data and I can tell you that refactoring this lambda function will make absolutely zero difference. In fact this is what lambda functions are for.

```
python3 -m timeit "lambda lobby_conn: lobby_conn.authenticated"
10000000 loops, best of 3: 0.0614 usec per loop
```

server/__init__.py
```diff
@@ -125,21 +118,14 @@ def validation_func(lobby_conn):
...
                 message,
                 lambda lobby_conn: lobby_conn.authenticated and validation_func(lobby_conn)
-            ))
-
-        try:
-            await asyncio.gather(*tasks)
-        except Exception as e:
-            logging.getLogger().exception(e)
-
-    ping_msg = encode_message('PING')
```
norraxx: What about exception processing? How they will behave?

server/__init__.py
```diff
@@ -125,21 +118,14 @@ def validation_func(lobby_conn):
                 def validation_func(lobby_conn):
```
norraxx: Dame here, extract functions out of the scope, just define them once

tests/integration_tests/test_servercontext.py
```diff
@@ -70,36 +70,6 @@ def fin():
...
-    writer.close()
-
-    # If connection errors aren't handled, this should fail due to a
-    # ConnectionError
-    for _ in range(20):
-        await ctx.broadcast(["Some message"])
-
-    assert len(ctx.connections) == 0
-
-
```
norraxx: I'm offering to add a little bit performance test, create like 100 clients, one server instance and profile it before and after

Askaholic: How would you determine if the test failed or succeeded? It would also be dependent on the hardware it's running on, and any constraints we would try to put on it would have to be pretty low so it would work on travis. 

Another problem is that the profile data depends largely on the clients and what they are doing. So in order to get a semi accurate profile data we would have to simulate the clients in an accurate way as if they were real clients. And that's another rabbit hole which Brutus and I have tried to go down already, but it would take a lot of work to get something which would simulate even remotely accurate usage data.

server/__init__.py
Askaholic: Hm. `send` doesn't raise errors as much as `drain` but I guess there are still some situations where it could raise an error.

https://github.com/python/cpython/blob/2477aed12d409c78a16ec435e01cee235af53221/Lib/asyncio/selector_events.py#L888

server/__init__.py
Askaholic: No point. These capture the current `game` variable. If you define this function elsewhere you would just have to call it with `functools.partial` which would probably be worse. Although again, defining 1 function per second is going to be completely negligible.

server/__init__.py
Askaholic: Actually, I've already wrapped all calls to `StreamWriter.write` in an exception handler so I think the main issue would be if there is a message encoding error.

server/__init__.py
norraxx: Ok

server/__init__.py
norraxx: Let's see :-)

norraxx:
:+1: good job

PR #550, opened 2020-03-21
Majority vote game outcomes

cleborys:
If multiple game outcomes are reported for a single player, `GameResults.outcome` will now pick the outcome with the most votes (i.e. number of distinct players reporting this outcome) if either of the following holds:
 - The outcome with the most votes has two or more votes and the runner-up has at most one.
 - The outcome with the most votes has at least three more votes than the runner-up.

Otherwise, `GameOutcome.CONFLICTING` will be returned.

Note that the number of distinct reporters is counted on a per-outcome basis. That is, if a single player reports two different outcomes, he will count as one vote towards each of these outcomes.


tests/data/test-data.sql
```diff
@@ -65,6 +65,10 @@ insert into login (id, login, email, password) values (6, 'newbie', 'noob@exampl
 insert into login (id, login, email, password, steamid) values (7, 'steambie', 'steambie@example.com', SHA2('password', 256), 111111);
 -- Test IPv6
 insert into login (id, login, email, password, ip) values (8, 'ipv6', 'ipv6@example.com', SHA2('ipv6', 256), '0000:0000:0000:0000:0000:0000:127.127.127.127');
+insert into login (id, login, email, password, create_time) values (9, 'player9', 'player9@example.com', SHA2('password', 256), '2000-01-01 00:00:00');
+-- player with id 10 is created further down...
+insert into login (id, login, email, password, create_time) values (11, 'player11', 'player11@example.com', SHA2('password', 256), '2000-01-01 00:00:00');
+insert into login (id, login, email, password, create_time) values (12, 'player12', 'player12@example.com', SHA2('password', 256), '2000-01-01 00:00:00');
```
Askaholic: I would put these at the bottom, especially because this is currently inside of a big comment label claiming that it's
`Contents of old FAForever/db/test-data.sql`.

Although are these even used?

server/games/game_results.py
```diff
@@ -75,21 +75,37 @@ def outcome(self, army: int) -> GameOutcome:
...
+        top_votes = len(voters[sorted_outcomes[0]])
+        runnerup_votes = len(voters[sorted_outcomes[1]])
+        if top_votes > 1 >= runnerup_votes or top_votes >= runnerup_votes + 3:
+            decision = sorted_outcomes[0]
+        else:
+            decision = GameOutcome.CONFLICTING
 
-        return GameOutcome.UNKNOWN
+        self._logger.info(
+            f"Multiple outcomes for game {self._game_id} army {army} "
```
Askaholic: Not that it matters here, but its better to use `%s` and let the logger do string formatting lazily. 

If you do it that way all the time, then you will never accidentally shoot yourself in the foot when the performance actually matters :P

tests/data/test-data.sql
cleborys: Good point.
They are necessary for the 12 player game, because they are used as foreign keys when the game writes stats to the database.
Of course these writes aren't really necessary for the test so they could also be mocked out instead...

server/games/game_results.py
cleborys: Ah, I wasn't really aware. I just went with the more readable one.

cleborys:
If you have more interesting edge cases, I'm happy to add them :)

tests/data/test-data.sql
Askaholic: Yea let's just mock it

tests/unit_tests/test_game.py
```diff
@@ -1074,6 +1074,158 @@ def custom_game(event_loop, database, game_service, game_stats_service):
...
+                (Rating(1500, 250), 0, 3),
+                (Rating(1500, 250), 0, 3),
+            ], 1)]
+    add_connected_players(game, [player for player, _, _ in players])
+    for player, _, team in players:
+        game.set_player_option(player.id, 'Team', team)
+        game.set_player_option(player.id, 'Army', player.id - 1)
+    await game.launch()
+
+    await game.add_result(players[0][0].id, 1, "defeat", 0)
```
Askaholic: Here's an idea. How about we extract this to a json file that we can put into `tests/data`. We could even write a script to create such json files from parsing actual server log snippets. Then we could easily add lots of real test cases without manually entering the results like this.

Maybe some simple format like:
```
{
    "teams": {
        2: [1,2,3,4,5,6],
        3: [7,8,9,10,11,12]
    },
    "results": [
        [3, 1, "defeat", 0],
        [9, 1, "defeat", 0],
        ...
    ],
    "winning_team": 2
}
```
We would have to add the `winning_team` field manually to each file.

tests/unit_tests/test_game.py
```diff
@@ -1074,6 +1075,49 @@ def custom_game(event_loop, database, game_service, game_stats_service):
...
+        player_id: player_factory(
+            login=f"{player_id}",
+            player_id=player_id,
+            global_rating=Rating(old_rating, 250),
+            with_lobby_connection=False,
+        )
+        for team in log_dict["teams"].values() for player_id in team
+    }
+
+    add_connected_players(game, [player for player in players.values()])
```
Askaholic: Why not this?
```suggestion
    add_connected_players(game, list(players.values()))
```

server/games/game_results.py
```diff
@@ -75,21 +75,38 @@ def outcome(self, army: int) -> GameOutcome:
...
+            lambda r: r.outcome is not GameOutcome.UNKNOWN, self[army]
+        ):
+            voters[report.outcome].add(report.reporter)
+
+        if len(voters) == 0:
+            return GameOutcome.UNKNOWN
+
+        if len(voters) == 1:
+            unique_outcome = next(iter(voters.keys()))
+            return unique_outcome
```
norraxx: you don't need this `unique_outcome`, just return first key from map

server/games/game_results.py
```diff
@@ -75,21 +75,38 @@ def outcome(self, army: int) -> GameOutcome:
...
+        for report in filter(
+            lambda r: r.outcome is not GameOutcome.UNKNOWN, self[army]
+        ):
+            voters[report.outcome].add(report.reporter)
+
+        if len(voters) == 0:
+            return GameOutcome.UNKNOWN
+
+        if len(voters) == 1:
+            unique_outcome = next(iter(voters.keys()))
```
norraxx: `return voters.keys()[0]` is better

server/games/game_results.py
```diff
@@ -75,21 +75,38 @@ def outcome(self, army: int) -> GameOutcome:
...
-            r.outcome for r in self[army]
-            if r.outcome is not GameOutcome.UNKNOWN
+        voters = defaultdict(set)
+        for report in filter(
+            lambda r: r.outcome is not GameOutcome.UNKNOWN, self[army]
+        ):
+            voters[report.outcome].add(report.reporter)
+
+        if len(voters) == 0:
+            return GameOutcome.UNKNOWN
```
norraxx: in docstring there is: `Reports unknown outcome if players disagree.`, do they disagree here?

server/games/game_results.py
```diff
@@ -75,21 +75,38 @@ def outcome(self, army: int) -> GameOutcome:
...
+
+        if len(voters) == 1:
+            unique_outcome = next(iter(voters.keys()))
+            return unique_outcome
+
+        sorted_outcomes = sorted(
+            voters.keys(),
+            reverse=True,
+            key=lambda x: (len(voters[x]), x.value)
```
norraxx: `x` in math is unknown to compute, the result, that we want to get. What `x` mean here? What's inside the `x`?

server/games/game_results.py
```diff
@@ -75,21 +75,38 @@ def outcome(self, army: int) -> GameOutcome:
...
-        if len(outcomes) == 1:
-            return outcomes.pop()
-        elif len(outcomes) > 1:
-            self._logger.info(
-                "Multiple outcomes for game %s army %s: %s",
-                self._game_id, army, list(outcomes)
-            )
-            return GameOutcome.CONFLICTING
+        top_votes = len(voters[sorted_outcomes[0]])
+        runnerup_votes = len(voters[sorted_outcomes[1]])
```
norraxx: `runner_up_votes`

server/games/game_results.py
cleborys: That gives me
```
TypeError: 'dict_keys' object does not support indexing
```

server/games/game_results.py
```diff
@@ -75,21 +75,38 @@ def outcome(self, army: int) -> GameOutcome:
         if army not in self:
             return GameOutcome.UNKNOWN
 
-        outcomes = set(
-            r.outcome for r in self[army]
-            if r.outcome is not GameOutcome.UNKNOWN
+        voters = defaultdict(set)
```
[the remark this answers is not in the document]
cleborys: Hm, I see your point, but then you would need
```python
voters[outcome] = voters.get(outcome, set()).add(reporter)
```
instead of 
```python
voters[outcome].add(reporter)
```
which I think is a bit less slick?

server/games/game_results.py
cleborys: Right, its unnecessary. I only added it to document why we are returning this seemingly random key.

server/games/game_results.py
cleborys: Oh, I forgot to update many things in that docstring, good catch!

server/games/game_results.py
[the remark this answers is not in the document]
norraxx: Let's see if it is sick.
First of all you know what's inside,, because near default dict you don't have proper typing.
Second, you don't add unnecessary magic, everything is explicit.

server/games/game_results.py
norraxx: Heh, in python 2 it worked... Sorry my bad.

server/games/game_results.py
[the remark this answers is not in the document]
Askaholic: Also @norraxx solution will create extra sets for ever call whereas `defaultdict` only creates them when they're needed. IMO `defaultdict` is the way to go. This is exactly what `defaultdict` was created for.

server/games/game_results.py
Askaholic: Yea, `.keys()`, `.items()` and `.values()` return iterables because it's faster than allocating a list when you don't need it.

Askaholic: In this case you could actually use `popitem` although this modifies the dictionary.
```suggestion
            unique_outcome = voters.popitem()[0]
```

PR #551, opened 2020-03-23
Only show players if they've managed to connect

### LINKED ISSUE #225: Players in lobby are not always detected properly
From time to time clients can show that a lobby has, say, 8/8 players even though in reality the lobby has only 7 players in it. I assume one or more players have left the lobby and it hasn't been properly detected.

This is quite an annoying bug and can cause problems filling a game, when clients show lobby is full even though it isn't.

Askaholic:
Might fix #225.

When doing testing of the game result issue and struggling with getting the ICE adapter to connect two machines on the same network I noticed what might be causing the display of extra players in lobby. 

The server would count a player as "in-lobby" as soon as they sent `GameState: Lobby` meaning as soon as their game made it to the `connecting to host` screen. This means if someone was trying to join a lobby, but could not connect to the host, it would still falsely bump the `num_players` which might cause the lobby to appear full when it was actually one player short. 

It seems like the only way to detect if a player has successfully connected to the host is by the host sending `PlayerOption` commands for them. So we can use the presence of player options as an indicator for successful connection.

I'm not sure if this is the only reason why this bug occurs so I'll leave the issue open for now.

### DIFF
--- a/server/games/game.py
+++ b/server/games/game.py
@@ -205,14 +205,17 @@ class Game:
         :return: frozenset
         """
         if self.state is GameState.LOBBY:
-            return frozenset(self._connections.keys())
+            return frozenset(
+                player for player in self._connections.keys()
+                if player and player.id in self._player_options
+            )
         else:
-            return frozenset({
+            return frozenset(
                 player
                 for player in self._players
                 if self.get_player_option(player.id, 'Army') is not None
                 and self.get_player_option(player.id, 'Army') >= 0
-            })
+            )
 
     @property
     def connections(self):
@@ -391,16 +394,25 @@ class Game:
         """
         if game_connection not in self._connections.values():
             return
-        del self._connections[game_connection.player]
-        if game_connection.player:
-            del game_connection.player.game
+
+        player = game_connection.player
+        del self._connections[player]
+
+        if player:
+            del player.game
+
+            if (
+                self.state is GameState.LOBBY and
+                player.id in self._player_options
+            ):
+                del self._player_options[player.id]
+
         await self.check_sim_end()
 
         self._logger.info("Removed game connection %s", game_connection)
 
         def host_left_lobby() -> bool:
-            return (game_connection.player == self.host and
-                    self.state is not GameState.LIVE)
+            return player == self.host and self.state is not GameState.LIVE
 
         if self.state is not GameState.ENDED and (
             len(self._connections) == 0 or host_left_lobby()
--- a/server/games/game.py
+++ b/server/games/game.py
@@ -207,7 +207,7 @@ class Game:
         if self.state is GameState.LOBBY:
             return frozenset(
                 player for player in self._connections.keys()
-                if player and player.id in self._player_options
+                if player.id in self._player_options
             )
         else:
             return frozenset(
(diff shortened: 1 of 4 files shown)


server/games/game.py
```diff
@@ -391,16 +394,25 @@ def add_game_connection(self, game_connection):
...
+                self.state is GameState.LOBBY and
+                player.id in self._player_options
+            ):
+                del self._player_options[player.id]
+
         await self.check_sim_end()
 
         self._logger.info("Removed game connection %s", game_connection)
 
         def host_left_lobby() -> bool:
```
norraxx: Do we need that function?
1. like i don't see the reason for inner function
2. this is just boolean result for condition, maybe we could just move body to condition

server/games/game.py
```diff
@@ -391,16 +394,25 @@ def add_game_connection(self, game_connection):
         """
         if game_connection not in self._connections.values():
             return
-        del self._connections[game_connection.player]
-        if game_connection.player:
-            del game_connection.player.game
+
+        player = game_connection.player
+        del self._connections[player]
```
norraxx: `if not player` there will be `KeyError`

And there is possibility that there could be no player, see the line below

server/games/game.py
```diff
@@ -391,16 +394,25 @@ def add_game_connection(self, game_connection):
...
+
+        if player:
```
norraxx: `if not player` - when that situation will happen?

server/games/game.py
cleborys: Instead of moving the body into the condition directly, I'd make it a variable.
```python
host_left_lobby = (game_connection.player == self.host and self.state is not GameState.LIVE)

if self.state is not GameState.ENDED and (
    len(self._connections) == 0 or host_left_lobby
):
```

tests/unit_tests/test_game.py
```diff
@@ -220,9 +220,39 @@ def custom_game(event_loop, database, game_service, game_stats_service):
     mock_game_connection.player = players.hosting
     mock_game_connection.state = GameConnectionState.CONNECTED_TO_HOST
     game.add_game_connection(mock_game_connection)
+    # Players should not be considered as 'in lobby' until the host has sent
+    # "PlayerOption" configuration for them
+    assert game.to_dict()["num_players"] == 0
```
cleborys: Maybe also add
```
assert len(game.players) == 0
```
Even if that's called by `to_dict` now it might not at some point in the future.

tests/unit_tests/test_game.py
```diff
@@ -220,9 +220,39 @@ def custom_game(event_loop, database, game_service, game_stats_service):
...
+    game.set_player_option(players.hosting.id, 'Team', 1)
+    assert game.players == {players.hosting}
+    # Join a new player
+    join_conn = add_connected_player(game, players.joining)
+    assert game.players == {players.hosting, players.joining}
+    # Player leaves
+    await game.remove_game_connection(join_conn)
+    assert game.players == {players.hosting}
+    # Player joins again
+    game.add_game_connection(join_conn)
```
cleborys: Maybe also add the previous assert here.
 ```python
assert game.players == {players.hosting}
```

server/games/game.py
Askaholic: Yea I'll do that

tests/unit_tests/test_game.py
Askaholic: I was trying to avoid making it too implementation specific, but maybe it's a good idea.

server/games/game.py
Askaholic: There won't be a `KeyError` there was already a test checking this.

server/games/game.py
Askaholic: I don't think it can actually.

PR #552, opened 2020-03-25
Unranking Games with Ambiguous Outcome

cleborys:
Completely reworks the decision flow in `GameRater`, the most prominent change being that `DEFEAT` vs `UNKNOWN` is no longer resolved to `DEFEAT` vs `VICTORY` but rather unranked.

Decision flow is now looks at the outcomes of all players on both teams (instead of resolving player outcomes to a single outcome per team first) and proceeds as follows:
1. Did someone claim `VICTORY`?
    - If no, proceed to 2. 
    - If players on both teams did, unrank.
    - Otherwise award victory.
2. Did someone claim `DRAW` or `MUTUAL_DRAW`?
    - If no, proceed to 3.
    - If players on both teams did, resolve to draw.
    - If only players on one team did, unrank.
3. Does someone have `UNKNOWN` or `CONFLICTING` outcome?
    - If yes, then we can't decide the outcome, unrank.
    - If no, then every player has `DEFEAT`, resolve to draw.

Note that `DRAW` and `MUTUAL_DRAW`, as well as `UNKNOWN` and `CONFLICTING` are treated the same from the rater's perspective.

This PR relies on and includes #550 so the diffs will be difficult to read until that has been merged. Changes in this PR are only to `game_rater.py` and `test_game_rater.py`, see commit `a9bde0b`.

There's a unit test for documenting the judgment of all 21 possible outcomes for 1v1 games, let me know if one of these isn't as you would expect.

As a side-note: The core method of `GameRater` is now a `staticmethod`, in particular I feel that it was a bad decision to give `GameRater` any state at all in the first place. I address this in the upcoming rating service refactor.

### DIFF
--- a/server/games/game_rater.py
+++ b/server/games/game_rater.py
@@ -1,4 +1,4 @@
-from typing import Dict, Iterable, List
+from typing import Dict, List, Set
 
 import trueskill
 from server.config import FFA_TEAM
@@ -20,7 +20,7 @@ class GameRater(object):
         self,
         players_by_team: Dict[int, List[Player]],
         outcome_by_player: Dict[Player, GameOutcome],
-        rating_type=RatingType.GLOBAL
+        rating_type=RatingType.GLOBAL,
     ):
         self._rating_type = rating_type
         self._outcome_by_player = outcome_by_player
@@ -29,7 +29,8 @@ class GameRater(object):
     def compute_rating(self) -> Dict[Player, Rating]:
         rating_groups = self._get_rating_groups()
         team_outcomes = [
-            self._get_team_outcome(team.keys()) for team in rating_groups
+            set(self._outcome_by_player[player] for player in team)
+            for team in rating_groups
         ]
         ranks = self._ranks_from_team_outcomes(team_outcomes)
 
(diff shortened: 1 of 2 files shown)


norraxx:
Why unrank? Is it approved by others?

cleborys:
I assume you are referring to both teams reporting `VICTORY` bein unranked?
As far as i understand, the lua code should never report victory for both teams, unless something got seriously out of sync or someone is faking reports, and in either case I would prefer to unrank.

If this is not as intended and should rather be resolved to `DRAW`, let me know.

server/games/game_results.py
```diff
@@ -75,21 +75,38 @@ def outcome(self, army: int) -> GameOutcome:
...
-            r.outcome for r in self[army]
-            if r.outcome is not GameOutcome.UNKNOWN
+        voters = defaultdict(set)
+        for report in filter(
+            lambda r: r.outcome is not GameOutcome.UNKNOWN, self[army]
+        ):
+            voters[report.outcome].add(report.reporter)
+
+        if len(voters) == 0:
+            return GameOutcome.UNKNOWN
```
Askaholic: I almost feel like we should have different types for `PlayerOutcome` and `GameOutcome`.

Issue #546, opened 2020-03-19
My rating isn't updating anymore


SvartfaR:
If you search for the 1v1 Leaderboards for "SvartfaR" it will show a rating of 411 points. This rating seems not to update since 4 further ranked games. Use the search of https://www.faforever.com/competitive/leaderboards/1v1 and take a look at the graph. The graph says it should be a rating of 705 points.

Maybe the ranked with replay id 11241961 caused the error. After this match I didn't got any notfication and the 1v1-Tab content didn't update as usually. (Thought my oppenent didn't leave the game.)

This issue occurs in the client and the website, so i think it is a server issue.
![Anmerkung 2020-03-19 170501](https://user-images.githubusercontent.com/51929253/77087864-eb19bc80-6a03-11ea-8e6b-faf7ca5b3116.png)
![Anmerkung 2020-03-19 170524](https://user-images.githubusercontent.com/51929253/77087866-ebb25300-6a03-11ea-91cf-4d978540a346.png)
![website](https://user-images.githubusercontent.com/51929253/77087868-ebb25300-6a03-11ea-9684-2a425f87e99d.png)

Askaholic:
Hi, thanks for making the issue. I see that the results for both armies in the game you mentioned was `UNKNOWN` which explains why it wasn't rated. 

https://api.faforever.com/data/game?filter=(id==11241961)&include=playerStats

I need to get more recent logs before I can investigate why, but my guess is there was probably a conflict in the reported results. We've been trying to make the game rating more accurate, but it appears that in reality there are a rather large number of games that report inconsistent results and therefore get rejected (which would have been rated anyways by the old code).

If you could give a brief explanation of the outcome of the game here that would be useful. Include stuff like who won and how they won (was their ACU killed, did they ctrl-k, or did they leave the game?)

SvartfaR:
Explanation of the outcome of the game: Horvat won because I ctrl-k my ACU and left the match via score screen.

> I see that the results for both armies in the game you mentioned was UNKNOWN which explains why it wasn't rated.

Do you mean it wasn't rated by the new code? In the graph you can see that I dropped from 461 to 412 points. (412 is next to 411, so I thought it's a rounding issue.)

Displays the graph the rating calculated by old code and the table displays the rating calculated of new code (which is stuck for me)?

Askaholic:
Actually, I misread it. The `result` field isn't implemented yet so it will always show `UNKNOWN`. I see that there are 
```
afterDeviation | 149.63
[...]

Issue #534, opened 2020-02-11
Implement new leaderboard tables


Askaholic:
We added some new tables for storing people's ratings in a more sensible manner. These need to be implemented on the server (and also API, but the server will be the only application writing to the tables).

### Checklist
- [x] Use new `leaderboard` and `leaderboard_rating` tables (#553)
- Migrate old tables into new ones
    - [ ] Recompute `wonGames`
    - [ ] Populate `leaderboard_rating_journal` with rating change data from `game_player_stats`
- [x] Load `RatingType` values from database instead of having a hardcoded enum (#428)

### How it works
A `leaderboard` table defines a certain type of rating i.e. global, ladder1v1 or ladder2v2. A `leaderboard_rating` table stores players current ratings for each leaderboard. This includes the total number of games played for that leaderboard. After a game finishes, the server computes players new ratings and writes the proposed change into a `leaderboard_rating_journal` table. This can also include verifying that current player rating data is consistent with the last recorded journal entry. Once the journal entry is written, the `leaderboard_rating` table is updated to the new computed rating.

#### List of steps for writing new ratings
1. Check the game type for an associated `leaderboard` entry.
2. Check that the last known `leaderboard_rating_journal` entry for each player is consistent with the current ratings. 
    2.1. If the ratings are inconsistent, try to repair them. That is, if the current `leaderboard_rating` value corresponds to the last entry for `mean_before` and `deviation_before`, then write the values of `mean_after` and `deviation_after` to the `leaderboard_rating` table. ^
    2.2 If the ratings are inconsistent, but do no match up with the last journal entry, then add a journal entry and signal that something bad happened to the database (it would be good to monitor when this happens in grafana).
3. Compute the new ratings.
4. Write the rating change to the `leaderboard_rating_journal`.
5. Write the new rating into the `leaderboard_rating` table.

See the migrations here:
https://github.com/FAForever/db/blob/develop/migrations/V73__leaderboards_and_map_pools.sql

#### Other thoughts
- ^ Need to be able to tell from the database if a player won their game in order to correctly repair `leaderboard_rating` from the journal.
- We need think carefully about what happens when a player is in two games which are rated at the same time. This is actually possible because of flawed logic in the game end calculations. That means we should think about how to set up our database transactions so that computing a new rating and writing it to the rating tables is more or less atomic.

Brutus5000:
This is long done.

PR #553, opened 2020-03-25
Rating Service

### LINKED ISSUE #265: rating is always calculated based on player rating when game started
join a game -> die in game -> join another game -> score of game that finished first doesn't count

probably needs to be fixed by updating the player rating from db before calculating new rating

cleborys:
Moves all rating functionality from methods on individual games to a `RatingService` with an input queue and a worker loop that constantly receives game summaries and rates them in a linearizable order.

The main benefit is to avoid race conditions like what happened in issue #546 or is mentioned in #534. Closes #265. Would hopefully eliminate the need to ever repair the rating journal as mentioned in #534. 

To test before merging:
 - [x] Rating a game with a player with old rating entries
 - [x] Rating a game with a player with new rating entries
 - [x] Rating a game with a newly signed up player without rating entries

Items for future PRs:
 - (Re)calculate `won_games` for ladder and global
 - Record previous games from `game_player_stats` in `leaderboard_journal`
 - Replace `RatingType` Enum with rating types from the database (almost) everywhere
 - Remove `GameOutcome.MUTUAL_DRAW` (?)


server/games/game.py
```diff
@@ -245,55 +244,49 @@ def has_ai(self) -> bool:
...
+        Special cases:
+         - Returns True if there are zero teams.
+         - Returns False if there is a single team.
+        """
+        # FIXME: explicitly allowing a team to be set to None here.
+        # All players with team `None` are considered on one team.
+        # This is what code in production does and allows to launch games where
+        # players are missing the "Team" option.
+        # (Rating code will later raise an Error if players are missing that option.)
+        # Is this intentional?
```
cleborys: Some unit tests were relying on games being able to launch without having the "Team" option set for every player. Is that intentional?

server/rating_service/game_rater.py
```diff
@@ -0,0 +1,126 @@
...
+    @staticmethod
+    def _check_rating_groups(rating_groups: RatingGroups):
+        if len(rating_groups) != 2:
+            raise GameRatingError(
+                "Attempted to rate game with other than two parties. "
+                f"Rating groups: {rating_groups}"
+            )
+
+    @staticmethod
+    def _get_team_outcome(team_rating_data: Dict[PlayerID, RatingData]) -> GameOutcome:
```
cleborys: This is outdated and removed in PR #552

server/rating_service/game_rater.py
```diff
@@ -0,0 +1,126 @@
...
+        if len(outcomes) > 1:
+            raise GameRatingError(
+                "Attempted to rate game where one team  has inconsistent outcome. "
+                f"Offending team: {team_rating_data}"
+            )
+        else:
+            return outcomes.pop()
+
+    @staticmethod
+    def _ranks_from_team_outcomes(team_outcomes: List[GameOutcome]) -> List[int]:
```
cleborys: This is outdated and replaced in PR #552

server/rating_service/rating_service.py
```diff
@@ -0,0 +1,254 @@
...
+                self._logger.warning("Missing rating entry to rate game %s", summary)
+            else:
+                self._logger.debug("Done rating request.")
+
+            self._queue.task_done()
+            rating_service_backlog.set(self._queue.qsize())
+
+            if self._queue.empty() and not self._accept_input:
+                self._logger.info("RatingService shutting down.")
+                break
```
cleborys: This isn't actually necessary, since `shutdown` will kill the task once the queue is empty. Might be cleaner if these lines here are removed.

server/rating_service/rating_service.py
```diff
@@ -0,0 +1,254 @@
...
+            self._logger.debug("Now rating request %s", summary)
+
+            try:
+                await self._rate(summary)
+            except GameRatingError:
+                self._logger.warning("Error rating game %s", summary)
+            except RatingNotFoundError:
+                self._logger.warning("Missing rating entry to rate game %s", summary)
+            else:
+                self._logger.debug("Done rating request.")
```
cleborys: If `self._rate` throws any other error that isn't caught, then the loop ends, the task exits, and we won't rate any more games while the input queue still keeps accepting input.

Maybe this is the place to add a bare except, log the error, and move on. Is that really the cleanest solution?

tests/conftest.py
```diff
@@ -22,10 +22,13 @@
 from server.lobbyconnection import LobbyConnection
 from server.matchmaker import MatchmakerQueue
 from server.player_service import PlayerService
+from server.rating_service.rating_service import RatingService
 from server.players import Player, PlayerState
 from server.rating import RatingType
 from tests.utils import MockDatabase
 
+pytestmark = pytest.mark.asyncio
```
Askaholic: Conftest doesn't have any tests in it so you shouldn't need this line

tests/unit_tests/test_game_rater.py
```diff
@@ -9,211 +11,228 @@ class MockPlayer:
...
-        p3: GameOutcome.DEFEAT,
-        p4: GameOutcome.UNKNOWN
+    rating = Rating(1500, 500)
+    victory_team = {
+        1: RatingData(GameOutcome.VICTORY, rating),
+        2: RatingData(GameOutcome.UNKNOWN, rating),
+    }
+    defeat_team = {
+        1: RatingData(GameOutcome.DEFEAT, rating),
+        2: RatingData(GameOutcome.UNKNOWN, rating),
```
Askaholic: Doesn't matter for the test, but in practice these player id's would be different from the ones on `victory_team` right?

tests/unit_tests/test_rating_service.py
```diff
@@ -0,0 +1,196 @@
...
+from trueskill import Rating
+from server.rating_service.typedefs import GameRatingSummary, RatingData
+from server.games.game_results import GameOutcome
+
+
+pytestmark = pytest.mark.asyncio
+
+
+@pytest.fixture()
+async def initialized_service(database, player_service):
```
Askaholic: Could just call it `rating_service`

tests/unit_tests/test_rating_service.py
```diff
@@ -0,0 +1,196 @@
...
+    # second game: results have been saved.
+    service._persist_rating_changes.assert_called_once()
+
+
+async def test_nonexisting_rating_error_handled(
+    initialized_service, game_rating_summary
+):
+    service = initialized_service
+    service._persist_rating_changes = CoroutineMock()
+    service._logger = mock.MagicMock()
```
Askaholic: Pretty sure a regular `Mock` will be fine here

tests/unit_tests/test_game_rating.py
```diff
@@ -0,0 +1,695 @@
...
+
+def get_persisted_results(mock_service):
+    PersistedResults = namedtuple(
+        "PersistedResults", ["rating_type", "ratings", "outcomes"]
+    )
+    args = mock_service._persist_rating_changes.await_args
+    if args is None:
+        return PersistedResults(None, {}, {})
+
+    # TODO why does this need to be args[0]?
```
Askaholic: Because mock `args` are a list of args for each time the mock was called. So `args[0]` gives you the args for the first call of the mock.

tests/unit_tests/test_game_rating.py
```diff
@@ -0,0 +1,695 @@
...
+    custom_game.state = GameState.LOBBY
+    add_connected_players(custom_game, [players.hosting, players.joining])
+    custom_game.set_player_option(players.hosting.id, "Team", 2)
+    custom_game.set_player_option(players.joining.id, "Team", 3)
+
+    await custom_game.launch()
+    await custom_game.add_result(players.hosting.id, 0, "victory", 1)
+    await custom_game.add_result(players.joining.id, 1, "defeat", 0)
+
+    # will refuse to rate short custom games
```
Askaholic: Or just call `enforce_rating`

server/rating_service/rating_service.py
cleborys: Ah, of course, thanks!

server/rating_service/rating_service.py
```diff
@@ -0,0 +1,254 @@
...
+        """
+        Finish rating all remaining games, then exit.
+        """
+        self._accept_input = False
+        self._logger.debug(
+            "Shutdown initiated. Waiting on current queue: %s", self._queue
+        )
+        await self._queue.join()
+        self._logger.debug("Queue emptied: %s", self._queue)
+        await self.kill()
```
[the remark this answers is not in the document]
cleborys: I wasn't sure if there's a benefit to calling `self._task.cancel()`.
I think I wanted to safeguard against having multiple worker threads in case you call `shutdown` and then `initialize` again, but actually I think that safeguarding should rather happen in `initialize`, I'll rework that...

server/game_service.py
```diff
@@ -10,19 +10,28 @@
 from server.games.game import Game, GameState, VisibilityState
 from server.matchmaker import MatchmakerQueue
 from server.players import Player
+from server.rating_service.rating_service import RatingService
+from server.rating_service.typedefs import GameRatingSummary
```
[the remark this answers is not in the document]
cleborys: I couldn't get that to work without introducing reference cycles, since `game.py` needs `GameRatingSummary` and is needed by e.g. `PlayerService`, which is needed by `RatingService`.
Some of these are only imported for typing though. Would you then fall back to providing the names as string instead of the class for typing?

tests/conftest.py
cleborys: True! Took me some time to figure out how to make everything use the same event loop, at some point I thought this was necessary...

tests/unit_tests/test_game_rater.py
cleborys: Right!

server/rating_service/rating_service.py
```diff
@@ -0,0 +1,382 @@
...
+from server.rating import RatingType
+from trueskill import Rating
+from .game_rater import GameRater, GameRatingError
+
+from sqlalchemy import select, and_
+from server.db.models import legacy_ladder1v1_rating as legacy_ladder1v1_table
+from server.db.models import legacy_global_rating as legacy_global_table
+from server.db.models import leaderboard_rating as rating_table
+from server.db.models import leaderboard as rating_type_table
+from server.db.models import game_player_stats as gps_table
```
Askaholic, line 31: I feel like these renames make the queries more confusing since you have to go back and check what table they're actually accessing.

server/rating_service/rating_service.py
```diff
@@ -0,0 +1,382 @@
...
+            if not row:
+                new_rating = await self._create_default_rating(
+                    conn, player_id, rating_type
+                )
+                return new_rating
+
+            if rating_type is RatingType.GLOBAL:
+                # The old `global_rating` table does not have a `winGames` column.
+                # This should be a decent approximation though.
+                won_games = row[table.c.numGames] // 2
```
Askaholic, line 193: We could probably query for this, but it would be pretty annoying. I think we'd have to look at the game_player_stats before/after rating values to see if they went up or down.

server/rating_service/rating_service.py
```diff
@@ -0,0 +1,382 @@
...
+                row[table.c.mean],
+                row[table.c.deviation],
+                row[table.c.numGames],
+                won_games,
+                self._rating_type_ids[rating_type.value],
+            )
+
+            return Rating(row[table.c.mean], row[table.c.deviation])
+
+    async def _create_default_rating(
```
Askaholic, line 211: The only thing I'm worried about with doing the creation here is that the matchmaker for example won't have a rating entry to use for matchmaking until AFTER someone plays a game. I think it makes more sense to do this on first login in `PlayerService`.

server/rating_service/rating_service.py
```diff
@@ -0,0 +1,382 @@
...
+                        new_rating.sigma,
+                        old_rating.mu,
+                        old_rating.sigma,
+                        game_id,
+                        player_id,
+                    ),
+                )
+
+                gps_rows = await conn.execute(
+                    select([gps_table.c.id]).where(
```
Askaholic, line 271: I think the previous query will return this value, or there is a way to make it return the value. This is a pretty common thing you want so it's been optimized to be doable with only one query AFAIK

server/rating_service/rating_service.py
```diff
@@ -0,0 +1,382 @@
...
+        player_id: PlayerID,
+        new_rating: Rating,
+        old_rating: Rating,
+        outcome: GameOutcome,
+    ) -> None:
+
+        is_victory = outcome is GameOutcome.VICTORY
+        rating_type_id = self._rating_type_ids[rating_type.value]
+
+        await conn.execute(
```
Askaholic, line 312: Insert expressions are actually really slick in SQLAlchemy: https://docs.sqlalchemy.org/en/13/core/tutorial.html#insert-expressions

server/rating_service/rating_service.py
cleborys: The `Player` objects will still provide a default rating of `(1500, 500)` that is hardcoded in their initialization (unless I'm missing something) but not written to the database until it changes.
I'll at least make sure that that default value is now chosen according to the config variable.

cleborys: I think I like that all writes to the rating tables are in the same place (the rating service), even if it means there is this ambiguity about default ratings...
Note that the default value written here only exists so that the rating change will be an `UPDATE` instead of an `INSERT` and can therefore also be left out if there is a slick way to "`UPSERT`".

Let me know if you disagree :)

server/rating_service/rating_service.py
cleborys: Hm, using `returning` like in the link below (for postgres) tells my that "RETURNING is not supported by this dialect's statement compiler"...
https://docs.sqlalchemy.org/en/13/dialects/postgresql.html#insert-update-returning

server/games/game_results.py
```diff
@@ -60,9 +62,7 @@ def is_mutually_agreed_draw(self, player_armies):
         for army in player_armies:
             if army not in self:
                 continue
-            if any(
-                r.outcome is not GameOutcome.MUTUAL_DRAW for r in self[army]
-            ):
+            if any(r.outcome is not GameOutcome.MUTUAL_DRAW for r in self[army]):
```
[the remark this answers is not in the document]
cleborys: As in we should remove all mention of `MUTUAL_DRAW` from the server code?

server/rating_service/rating_service.py
cleborys: We could probably do this for writing rating changes:
https://docs.sqlalchemy.org/en/13/dialects/mysql.html#insert-on-duplicate-key-update-upsert

server/rating_service/rating_service.py
cleborys: The best thing I though of so far is to find the row id first and then update the row in the second query, rather than the other way around. That way there is at least only one `WHERE` search that's not on the primary key.

Askaholic: This is what I had in my memory but I guess it doesn't work for update statements. https://docs.sqlalchemy.org/en/13/core/connections.html#sqlalchemy.engine.ResultProxy.inserted_primary_key.

I think in this case we can update both table in a single expression:
https://www.mysqltutorial.org/mysql-update-join/

server/game_service.py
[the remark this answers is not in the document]
cleborys: Moving all the `Enum`s defined in `game.py` gets rid of the circular imports :)

server/rating_service/rating_service.py
cleborys: Hm, but one is an update (to `game_player_stats`) and one is an insert (into `leaderboard_rating_journal`) which references the id of the updated row.

server/rating_service/rating_service.py
cleborys: Ah no that doesn't work, since the table doesn't list (`login_id`, `leaderboard_id`) pairs as unique I think.

server/rating_service/rating_service.py
cleborys: Currently I am doing three queries and one search:
1. Select the `game_player_stats` row id based on the game data
2. Update that row (knowing it's id already)
3. Insert into `leaderboard_rating_journal`, referencing the id.

The following would have two queries and two searches:
1.Update the `game_player_stats` row (search for it based on game data)
2. Insert into  `leaderboard_rating_journal`, setting the id based on a `SELECT`.

Neither of them is as efficient as returning the updated id on the first query and I can't say whether the connection or the selection overhead is worse without experimenting on the real data.

Here's what the second query in the latter case would look like:
```python
journal_insert_sql = leaderboard_rating_journal.insert().values(
    leaderboard_id=rating_type_id,
    rating_mean_before=old_rating.mu,
    rating_deviation_before=old_rating.sigma,
    rating_mean_after=new_rating.mu,
    rating_deviation_after=new_rating.sigma,
    game_player_stats_id=select([game_player_stats.c.id]).where(
        and_(
            game_player_stats.c.playerId == player_id,
            game_player_stats.c.gameId == game_id,
        )
    )
)
```

Askaholic: Yea that second one is what I was going to suggest next. Just making the `id` query a subquery of the insert statement.

Askaholic: There is also the option of tracking and passing around game_player_stats ids (from game to rating service) but that's a pretty ugly solution just to save 1 query.

cleborys: I'll go for the subquery then.

Askaholic:
Rebase please

server/rating_service/rating_service.py
```diff
@@ -0,0 +1,365 @@
...
+                    conn, player_id, rating_type
+                )
+                return new_rating
+
+            if rating_type is RatingType.GLOBAL:
+                # The following query counts entries in game_player_stats,
+                # where the rating of the given player improved.
+                # If the concerned game was never rated, after_mean is NULL
+                # and the row is not counted.
+                count_wins_query = (
```
Askaholic, line 197: Cool. Unfortunately getting accurate win counts is going to be quite a bit more complicated than this. First off we will need to check the game mode. Global games are only games played on the `faf` game mode AFAIK, and it gets worse:

Way back when ladder was first introduced, there was no `ladder1v1` game mode, so all ladder games were also played on the `faf` game mode. Granted, this only affects the few players who have been playing FAF for a very long time, but these are also the players who will be most upset if their ladder history is messed up. I ran into this issue when trying to fix the ladder win ratio for the first time, so I've already done some work to figure out how we can distinguish these old ladder games from custom games with few false positives. You might want to take a look at:
https://github.com/FAForever/db/blob/develop/migrations/V67__add-old-ladder-games-to-winratio.sql and
https://github.com/FAForever/db/blob/develop/migrations/V65__fix-ladder-win-ratio.sql
Given that the ladder win ratio's still aren't working correctly, I think we should also do this win counting for ladder too.

Another thing we might need to keep an eye on is performance. As I recall, those migrations for updating every player's game count both took several hours to complete. Doing things lazily on a per-player basis should help a lot, but maybe we should track the execution time of this query, and put some extra thought into making sure it does only the minimum amount of work necessary.

server/rating_service/rating_service.py
Askaholic: Ok, as long as the player object has the default rating too then I guess that's fine.

server/rating_service/rating_service.py
```diff
@@ -0,0 +1,365 @@
...
+                        and_(
+                            game_player_stats.c.playerId == player_id,
+                            game_player_stats.c.gameId == game_id,
+                        )
+                    )
+                    .values(
+                        after_mean=new_rating.mu,
+                        after_deviation=new_rating.sigma,
+                        mean=old_rating.mu,
+                        deviation=old_rating.sigma,
```
Askaholic, line 279: This now becomes duplicated across both `game_player_stats` and `leaderboard_rating_journal`. I think in a perfect world these fields would be removed from `game_player_stats` and instead rating change history would only be recorded in the journal. But I'd like to know what @Brutus5000 thinks about this.

If we drop those columns from `game_player_stats` we'll have to back-fill `leaderboard_rating_journal` first in order to preserve the history, which again is going to be a lot of work. Plus apps that display your rating history (like the client) will need to be changed to use the journal table, although this should be pretty trivial (simply calling a different API endpoint).

server/rating_service/rating_service.py
cleborys: Hm,... I still think it might be a good place to cut corners and just set the global game wins to half of the total global games, especially if the performance is a problem and keep the current ladder game win count. For players with a *lot* of global games that is expected anyways and hard to take offense at, while for players with *very few* global games that care about it (aka that will play more games) it should converge somewhat quickly anyways.

Of course that means it wouldn't fix the ladder win count. But maybe that's something that's better done in a background job like the ones you described?
Since the counts are only kept in the `leaderboard_rating` table but not in the journal, we wouldn't make fixing these counts any harder I think, since we wouldn't need to rewrite the journal history.

cleborys: As for making these fixes a bit easier, maybe we could ask the community / the high level players, whether they would mind if their game counts would only reflect games from 2015 on (so that the mod would correctly distinguish between ladder and global).

server/rating_service/rating_service.py
cleborys: Back-filling `leaderboard_rating_journal` would make the global/ladder game count and win ratio queries discussed above significantly simpler, I think.

Brutus5000: Dropping these fields from `game_player_stats` would be the plan, but that is a backward-compatible change and will break all clients touching these.
We might need an API workaround to still populate them :thinking:

server/db/models.py
```diff
@@ -157,7 +161,7 @@
     Column('last_login',    TIMESTAMP)
 )
 
-ladder1v1_rating = Table(
+legacy_ladder1v1_rating = Table(
```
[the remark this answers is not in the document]
cleborys: I thought it would be good to mark them as deprecated in the server code. Will change.

server/rating_service/rating_service.py
```diff
@@ -0,0 +1,365 @@
...
+        if self._task is not None:
+            self._logger.error("Service already runnning or not properly shut down.")
+            return
+
+        await self._load_rating_type_ids()
+        self._accept_input = True
+        self._logger.debug("RatingService starting...")
+        self._task = asyncio.create_task(self._handle_rating_queue())
+
+    async def _load_rating_type_ids(self):
```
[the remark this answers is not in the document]
cleborys: ... Actually `PlayerService` wasn't using it yet (but will have to if we want to get rid of the `RatingType` enum), so I removed it for now.

server/rating_service/rating_service.py
```diff
@@ -0,0 +1,349 @@
...
+                "Tried to fetch player data before initializing service."
+            )
+            raise ServiceNotReadyError("RatingService not yet initialized.")
+
+        rating_type_id = self._rating_type_ids.get(rating_type.value)
+        if rating_type_id is None:
+            raise ValueError(f"Unknown rating type {rating_type}.")
+
+        async with self._db.acquire() as conn:
+            sql = select([leaderboard_rating]).where(
```
Askaholic, line 149: This is selecting 6 columns, but you only need 2 of them

server/player_service.py
```diff
@@ -111,6 +96,80 @@ async def fetch_player_data(self, player):
...
+            if rating_type not in retrieved_ratings
+        ]
+        await self._fetch_player_legacy_rating(player, types_not_found, conn)
+
+    async def _fetch_player_legacy_rating(self, player, rating_types, conn):
+        if not rating_types:
+            return
+
+        sql = select(
+            [global_rating, ladder1v1_rating], use_labels=True
```
Askaholic, line 137: Selecting 13 columns but you only need 6.

server/games/game.py
cleborys: [bumping this]

server/games/enums.py
```diff
@@ -0,0 +1,88 @@
...
+
+@unique
+class Victory(Enum):
+    DEMORALIZATION = 0
+    DOMINATION = 1
+    ERADICATION = 2
+    SANDBOX = 3
+
+    @staticmethod
+    def from_gpgnet_string(value: str) -> Optional["Victory"]:
```
norraxx, line 21: Imho, such functions, shouldn't be the part of the enum classes.

PR #564, opened 2020-04-04
Add ability for proper integration testing

Askaholic:
Experimenting with some real integration tests, tests that actually run on the test server. Not sure if we want to keep them in this repository, or give them their own dedicated one, but for now this will do.


integration_tests/fafclient.py
```diff
@@ -0,0 +1,155 @@
...
+
+from server.protocol import QDataStreamProtocol
+from tests.integration_tests.conftest import read_until, read_until_command
+
+
+class FAFClient(object):
+    """docstring for FAFClient."""
+
+    def __init__(self, user_agent="faf-client", version="1.0.0-dev"):
+        self.proto = None
```
norraxx, line 13: What is proto?

integration_tests/conftest.py
```diff
@@ -0,0 +1,15 @@
...
+@pytest.fixture(scope="session")
+def test_client(request):
+    """Create a new client connected to the test server"""
+    async def connect(username, password="foo"):
+        client = FAFClient()
+        await client.connect("test.faforever.com", 8001)
+        msg = await client.login(username, password)
+        return client, msg
+
+    return connect
```
norraxx, line 15: Imho, you would like yield fixture, that automatically connects to server and disconnects after tests,

integration_tests/conftest.py
```diff
@@ -0,0 +1,15 @@
+import pytest
+
+from .fafclient import FAFClient
+
+
+@pytest.fixture(scope="session")
+def test_client(request):
```
norraxx, line 7: Imho, this should ne async

integration_tests/test_game.py
```diff
@@ -0,0 +1,106 @@
...
+            )
+        )
+
+    await host.send_message({
+        "target": "game",
+        "command": "EnforceRating"
+    })
+
+    for client in all_clients:
+        for result in results:
```
cleborys, line 37: It doesn't matter right now and might not ever, but I guess the more "realistic" order of messages here might be
```python
for result in results:
  for client in all_clients:
    # ...
```
That is, I expect people to die one after the other and messages being sent in response to someone dying.
(Although it's true that individual clients rather send their messages in bursts at the end when there is a lot happening at once.)

integration_tests/test_game.py
```diff
@@ -0,0 +1,106 @@
...
+    await simulate_game(client1, client2, results=[
+        [2, "defeat -10"],
+        [1, "victory 10"]
+    ])
+
+    # Check that the ratings were updated
+    new_ratings = await client1.get_player_ratings("test", "test2")
+
+    assert ratings["test"][0] < new_ratings["test"][0]
+    assert ratings["test2"][0] > new_ratings["test2"][0]
```
cleborys, line 71: I got briefly confused here because I expected to read `new_ratings > old_ratings` rather than `ratings < new_ratings`. (i.e. the new one on the left and maybe an `old_` prefix.)

integration_tests/test_game.py
```diff
@@ -0,0 +1,106 @@
...
+
+    for client in all_clients:
+        for result in results:
+            await client.send_message({
+                "target": "game",
+                "command": "GameResult",
+                "args": result
+            })
+
+    # Report GameEnded
```
cleborys, line 44: Is this what's send when I click "exit to score" or the like?

cleborys: I assume these are meant as cleanup for the test server rather than as part of the test, since it happens after all assertions? It would be nice to put this in some kind of teardown part so that they disconnect and the game on the test server is cleaned up even if the test fails. (Although I'm not completely sure how to have teardown for anything in pytest that's not a fixture right now.)

integration_tests/test_game.py
```diff
@@ -0,0 +1,106 @@
...
+    with pytest.raises(asyncio.TimeoutError):
+        await client1.get_player_ratings("test", "test2", timeout=3)
+
+    # Now disconnect both players
+    for client in (client1, client2):
+        await client.send_message({
+            "target": "game",
+            "command": "GameState",
+            "args": ["Ended"]
+        })
```
cleborys, line 106: Same as above

integration_tests/test_login.py
```diff
@@ -0,0 +1,17 @@
...
+@pytest.mark.parametrize("username", (
+    "test",
+    "test2"
+))
+async def test_user_existence(test_client, username):
+    """Verify that these users exist on the test server"""
+    client, welcome_message = await test_client(username, "foo")
+
+    assert welcome_message["login"] == welcome_message["me"]["login"] == username
+    assert welcome_message["id"] == welcome_message["me"]["id"]
```
cleborys, line 16: That seems a bit of a weird assert but ok.

integration_tests/fafclient.py
```diff
@@ -0,0 +1,155 @@
...
+
+    def get_unique_id(self, session):
+        try:
+            process = subprocess.run(
+                [self.faf_uid_path, str(session)],
+                stdout=subprocess.PIPE
+            )
+            return process.stdout.decode()
+        except OSError:
+            return "some_fake_uid"
```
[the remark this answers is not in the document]
Askaholic: The uniqueness actually doesn't seem to matter since any results returned by the policy server are just ignored. The problem is if you send it data that isn't a valid uid then the policy server will return an http error which does cause the login to fail. So it doesn't matter at all what goes here, if this branch is reached then the login will fail.

integration_tests/test_game.py
Askaholic: These are reported as soon as the simulation ends, so when you see "Game Over" appear on the screen. No need to click on the score button.

integration_tests/test_login.py
Askaholic: It's apparently legacy behaviour. I'm assuming the python client uses it.

integration_tests/fafclient.py
```diff
@@ -0,0 +1,155 @@
...
+        return await read_until_command(self.proto, command, timeout=timeout)
+
+    # Commonly used functionality here
+    async def login(self, username, password):
+        """Perform the entire login sequence including fetching a session
+        and calling `faf-uid`"""
+        await self.proto.send_message({
+            "command": "ask_session",
+            "user_agent": self.user_agent,
+            "version": self.version
```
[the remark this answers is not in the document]
Askaholic: If it's too old then the server won't send a `session` and you won't be able to pass the `uid` check. So you won't be able to log in.

PR #565, opened 2020-04-05
Issue/#558 api authentication

### LINKED ISSUE #558: API Authentication error
Seems like the OAuth authentication with the API is not happening correctly. This is preventing achievements from being saved.
```
  File "/code/server/games/game.py", line 355, in _process_army_stats_for_player
    player, self, self._army_stats
  File "/code/server/stats/game_stats_service.py", line 104, in process_game_stats
    updated_achievements = await self._achievement_service.execute_batch_update(player.id, a_queue)
  File "/code/server/stats/achievement_service.py", line 98, in execute_batch_update
    response, content = await self.api_accessor.update_achievements(queue, player_id)
  File "/code/server/api/api_accessor.py", line 68, in update_achievements
    return await self.api_patch("achievements/update", achievements_data)
  File "/code/server/api/api_accessor.py", line 86, in api_patch
    return await api.request("PATCH", API_BASE_URL + path, headers=headers, json=json_data)
  File "/code/server/api/oauth_session.py", line 67, in request
    raise MissingTokenError()
oauthlib.oauth2.rfc6749.errors.MissingTokenError: (missing_token)
```

Askaholic:
Seems like the API was not returning refresh tokens in the way that was expected by the code, so after 1 hour when the first token expired, achievements would stop working. Now token expiration is checked on request, and if the token has expired it will either use a refresh token if one is available or just fetch a new token using the client id and secret.

Also adds more logging to the achievement/event stuff.

Fixes #558

### DIFF
--- a/server/api/api_accessor.py
+++ b/server/api/api_accessor.py
@@ -1,4 +1,3 @@
-import logging
 from ssl import SSLError
 from typing import Optional
 
@@ -57,6 +56,7 @@ class SessionManager:
         raise ConnectionError
 
 
+@with_logger
 class ApiAccessor:
     def __init__(self):
         self.api_session = SessionManager()
@@ -69,6 +69,8 @@ class ApiAccessor:
             achievement['achievementId'] = achievement.pop('achievement_id')
             achievement['operation'] = achievement.pop('update_type')
 
+        self._logger.debug("Sending achievement data: %s", achievements_data)
+
         return await self.api_patch("achievements/update", achievements_data)
 
     async def update_events(self, events_data, player_id):
@@ -77,6 +79,8 @@ class ApiAccessor:
         for event in events_data:
             event['playerId'] = player_id
             event['eventId'] = event.pop('event_id')
+
+        self._logger.debug("Sending event data: %s", events_data)
 
         return await self.api_patch("events/update", events_data)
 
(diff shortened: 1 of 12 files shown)


server/api/api_accessor.py
```diff
@@ -8,53 +7,56 @@
...
     def __init__(self):
         self.session = None  # Instance of session
-        self.logger = logging.getLogger()
 
     async def get_session(self) -> Optional[OAuth2Session]:
-        if self.session:
-            return self.session
-
-        try:
+        if not self.session:
```
cleborys, line 24: Is there more than just the `is not None` check here?

server/api/oauth_session.py
```diff
@@ -34,28 +46,24 @@ def __init__(
...
-            self.refresh_token = refresh_token
-            expires_in = int(creds['expires_in'])
-            asyncio.create_task(self._schedule_refresh(expires_in - 5))
-
-    async def _schedule_refresh(self, wait_time):
-        await asyncio.sleep(wait_time)
-        await self.refresh_tokens()
+        self.refresh_token = creds.get('refresh_token')
+        expires_in = creds.get('expires_in')
+        if expires_in is not None:
```
cleborys, line 64: When will this be None? Can it happen that the API decides to send tokens without an expiry field and then you will consider all of them to be expired immediately?

server/api/api_accessor.py
Askaholic: No

server/api/oauth_session.py
Askaholic: Well, it's data supplied by an external source so anything can happen. AFAIK it shouldn't happen though

tests/conftest.py
```diff
@@ -241,13 +242,11 @@ def matchmaker_queue(game_service) -> MatchmakerQueue:
 
 @pytest.fixture()
 def api_accessor():
-    class FakeSession:
-        def __init__(self):
-            self.request = CoroutineMock(return_value=(200, 'test'))
-            self.fetch_token = CoroutineMock()
+    session = asynctest.create_autospec(OAuth2Session)
```
norraxx, line 245: :+1:

Issue #432, opened 2019-05-11
Add support for new database schema


Askaholic:
The new schema uses postgresql so we can switch to `aiopg`. With SQLAlchemy queries should be pretty easy to rewrite. 

https://supcomhub.gitlab.io/sh-core-db-migrations/

We'll want to support both schemas at first, so that the new version can be enabled with a config switch.
* Add config variable for choosing the db schema to use
* Add `db/models_2.py` for needed table definitions
* Refactor queries to abstract them away from the business logic.

Brutus5000:
Still on mariadb in 2025 :D

PR #569, opened 2020-04-08
More SQLAlchemy

cleborys:
In my quest to understand more SQL and SQLAlchemy I converted most of the queries that would execute a hardcoded string, into pure sqlalchemy.

There are some cases in which I'm not completely confident in the test coverage of these statements, mostly where concerning the modvault.

### DIFF
--- a/server/db/models.py
+++ b/server/db/models.py
@@ -3,6 +3,7 @@ from sqlalchemy import (
     DateTime, Enum, Float, ForeignKey, Integer,
     MetaData, String, Table, Text
 )
+from sqlalchemy.dialects.mysql import LONGBLOB
 
 from ..games.game import Victory
 
@@ -158,6 +159,15 @@ coop_leaderboard = Table(
     Column('player_count',  Integer)
 )
 
+coop_map = Table(
+    'coop_map', metadata,
+    Column('id',            Integer,    primary_key=True),
+    Column('type',          Integer,    nullable=False),
+    Column('name',          String),
+    Column('description',   String),
+    Column('filename',      String,     index=True)
+)
+
 # This is actually a view into the `ban` table with proper handling of ban
 # expiration and revocation
 lobby_ban = Table(
@@ -202,5 +212,38 @@ teamkills = Table(
     Column('teamkiller',    Integer, ForeignKey('login.id')),
     Column('victim',        Integer, ForeignKey('login.id')),
     Column('game_id',       Integer, ForeignKey('game_stats.id')),
-    Column('gametime',      Integer)
+    Column('gametime',      Integer),
+)
+
+table_mod = Table(
+    'table_mod', metadata,
+    Column('id',            Integer,    nullable=False),
+    Column('uid',            String,     nullable=False),
+    Column('name',          String,     nullable=False),
+    Column('version',       Integer,    nullable=False),
+    Column('author',        String(100), nullable=False),
+    Column('ui',            Integer),
+    Column('date',          TIMESTAMP,  nullable=False),
+    Column('downloads',     Integer),
+    Column('likes',         Float),
+    Column('played',        Integer),
+    Column('description',   String,     nullable=False),
+    Column('filename',      String,     nullable=False),
+    Column('icon',          String),
+    Column('likers',        LONGBLOB),
+)
+
+mod_stats = Table(
+    'mod_stats', metadata,
+    Column('mod_id',    Integer,    ForeignKey('mod.id'), primary_key=True),
+    Column('likes',     Float,      nullable=False),
+    Column('likers',    LONGBLOB,   nullable=False),
+    Column('downloads', Integer,    nullable=False),
+)
+
+mod_version = Table(
+    'mod_version', metadata,
+    Column('id',        Integer,    primary_key=True),
+    Column('mod_id',    Integer,    nullable=False),
+    Column('uid',       String(40), nullable=False,     index=True),
 )
--- a/server/db/models.py
+++ b/server/db/models.py
@@ -1,5 +1,6 @@
 from sqlalchemy import (
-    TIMESTAMP, Boolean, Column, DateTime, Enum, Float, ForeignKey, Integer,
+    TIMESTAMP, TIME, Boolean, Column,
+    DateTime, Enum, Float, ForeignKey, Integer,
     MetaData, String, Table, Text
 )
 
@@ -103,15 +104,15 @@ game_player_stats = Table(
 
 game_stats = Table(
     'game_stats', metadata,
-    Column('id',        Integer,    primary_key=True),
-    Column('startTime', TIMESTAMP,  nullable=False, server_default="CURRENT_TIMESTAMP"),
+    Column('id',        Integer,        primary_key=True),
+    Column('startTime', TIMESTAMP,      nullable=False, server_default="CURRENT_TIMESTAMP"),
     Column('endTime',   TIMESTAMP),
-    Column('gameType',  Enum(Victory), nullable=False),
-    Column('gameMod',   Integer, ForeignKey('game_featuredMods.id'), nullable=False),
-    Column('host',      Integer,    nullable=False),
+    Column('gameType',  Enum(Victory),  nullable=False),
+    Column('gameMod',   Integer,        ForeignKey('game_featuredMods.id'), nullable=False),
+    Column('host',      Integer,        nullable=False),
     Column('mapId',     Integer),
-    Column('gameName',  String,     nullable=False),
-    Column('validity',  Integer,    nullable=False),
+    Column('gameName',  String,         nullable=False),
+    Column('validity',  Integer,        nullable=False),
 )
 
 global_rating = Table(
@@ -147,6 +148,16 @@ ladder1v1_rating = Table(
     Column('is_active',     Boolean,    nullable=False)
 )
 
+coop_leaderboard = Table(
+    'coop_leaderboard', metadata,
+    Column('id',            Integer,    primary_key=True),
+    Column('mission',       Integer),
+    Column('gameuid',       Integer,    ForeignKey('game_stats')),
+    Column('secondary',     Integer),
+    Column('time',          TIME),
+    Column('player_count',  Integer)
+)
+
 # This is actually a view into the `ban` table with proper handling of ban
 # expiration and revocation
 lobby_ban = Table(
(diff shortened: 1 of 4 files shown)


server/db/models.py
```diff
@@ -147,6 +149,25 @@
...
+)
+
+coop_map = Table(
+    'coop_map', metadata,
+    Column('id',            Integer,    primary_key=True),
+    Column('type',          Integer,    nullable=False),
+    Column('name',          String),
+    Column('description',   String),
+    Column('filename',      String,     index=True)
+)
```
norraxx, line 169: Poor pep-8

server/db/models.py
```diff
@@ -158,25 +179,72 @@
...
-   Column('create_time',   TIMESTAMP,  nullable=False),
-   Column('update_time',   TIMESTAMP,  nullable=False)
+   Column('last_moderator',         ForeignKey('login.id'),         index=True),
+   Column('create_time',            TIMESTAMP,                      nullable=False),
+   Column('update_time',            TIMESTAMP,                      nullable=False),
+   Column(
+       'report_status',
+       Enum('AWAITING', 'PROCESSING', 'COMPLETED', 'DISCARDED'),
+       nullable=False
+    )
```
norraxx, line 196: You know, you didn't made it better

server/db/models.py
```diff
@@ -158,25 +179,72 @@
...
+    Column('likers',        LONGBLOB,   nullable=False),
+    Column('downloads',     Integer,    nullable=False),
+    Column('times_played',  Integer),
+)
+
+mod_version = Table(
+    'mod_version', metadata,
+    Column('id',        Integer,    primary_key=True),
+    Column('mod_id',    Integer,    nullable=False),
+    Column('uid',       String(40), nullable=False,     index=True),
```
norraxx, line 249: C'mon...

server/lobbyconnection.py
```diff
@@ -16,12 +16,16 @@
 import server.metrics as metrics
 from server.db import FAFDatabase
 from sqlalchemy import and_, func, select, text
+from sqlalchemy.sql.functions import now as sql_now
```
norraxx, line 19: From sqlalchemy import func, `func.NOW(...)`
Func is proxy, you can all with it any db function.

server/lobbyconnection.py
```diff
@@ -585,7 +609,10 @@ def _set_user_agent_and_version(self, user_agent, version):
             irc_pass = "md5:" + str(m.hexdigest())
 
             try:
-                await conn.execute("UPDATE anope.anope_db_NickCore SET pass = %s WHERE display = %s", (irc_pass, login))
+                await conn.execute(
+                    "UPDATE anope.anope_db_NickCore SET pass = %s WHERE display = %s",
+                    (irc_pass, login)
+                )
```
norraxx, line 615: :see_no_evil:

server/lobbyconnection.py
```diff
@@ -715,8 +747,17 @@ def _set_user_agent_and_version(self, user_agent, version):
...
             async with self._db.acquire() as conn:
                 result = await conn.execute(
-                    "SELECT url, tooltip FROM `avatars` "
-                    "LEFT JOIN `avatars_list` ON `idAvatar` = `avatars_list`.`id` WHERE `idUser` = %s", (self.player.id,))
+                    select([
+                        avatars_list.c.url,
+                        avatars_list.c.tooltip
+                    ]).select_from(
+                        avatars.outerjoin(
+                            avatars_list
```
norraxx, line 755: You better give conditions

server/lobbyconnection.py
```diff
@@ -715,8 +747,17 @@ def _set_user_agent_and_version(self, user_agent, version):
...
+                        )
+                    ).where(
+                        avatars.c.idUser == self.player.id
+                    )
+                )
```
norraxx, line 760: Imho, you made different query here. Please check what print(str(query)) does.

Imho select_from will generate whole table like select something from cartesian product, and then filter it.

server/lobbyconnection.py
```diff
@@ -895,66 +951,121 @@ def _set_user_agent_and_version(self, user_agent, version):
...
-                        "UPDATE mod_stats s "
-                        "JOIN mod_version v ON v.mod_id = s.mod_id "
-                        "SET s.likes = s.likes + 1, likers=%s WHERE v.uid = %s",
-                        json.dumps(likers), uid)
+                        mod_stats.update().values(
+                            likes=mod_stats.c.likes + 1,
+                            likers=json.dumps(likers)
+                        ).where(
+                            mod_version.c.uid == uid
+                        ).where(
```
norraxx, line 1052: Imho you can write multiple and conditions in one where separated with comma

server/lobbyconnection.py
```diff
@@ -895,66 +951,121 @@ def _set_user_agent_and_version(self, user_agent, version):
...
-                    "UPDATE mod_stats s "
-                    "JOIN mod_version v ON v.mod_id = s.mod_id "
-                    "SET downloads=downloads+1 WHERE v.uid = %s", uid)
+                    mod_stats.update().values(
+                        downloads=mod_stats.c.downloads + 1,
+                    ).where(
+                        mod_version.c.uid == uid
+                    ).where(
+                        mod_stats.c.mod_id == mod_version.c.mod_id
+                    )
```
norraxx, line 1067: -//-

server/lobbyconnection.py
Askaholic: No, SQLAlchemy does this automatically. That's the whole point.

server/lobbyconnection.py
Askaholic: You are wrong. https://docs.sqlalchemy.org/en/13/core/tutorial.html#using-joins

server/db/models.py
Askaholic: Flake8 has no problem with this

server/db/models.py
Askaholic: Yea, in this file I'd let lines go as long as they need so we have 1 table column per line

server/db/models.py
Askaholic: What is that supposed to mean?

server/db/models.py
```diff
@@ -147,6 +149,25 @@
     Column('is_active',     Boolean,    nullable=False)
 )
 
+coop_leaderboard = Table(
```
Askaholic, line 152: Please place tables in alphabetical order

server/db/models.py
```diff
@@ -158,25 +179,72 @@
...
+    Column('played',        Integer),
+    Column('description',   String,     nullable=False),
+    Column('filename',      String,     nullable=False),
+    Column('icon',          String),
+    Column('likers',        LONGBLOB),
+)
+
+mod_stats = Table(
+    'mod_stats', metadata,
+    Column('mod_id',        Integer,    ForeignKey('mod.id'), primary_key=True),
```
Askaholic, line 238: Is it supposed to be `table_mod`?
```suggestion
    Column('mod_id',        Integer,    ForeignKey('table_mod.id'), primary_key=True),
```

server/gameconnection.py
```diff
@@ -478,10 +494,14 @@ def player(self, val: Player):
             if len(self.game.mods.keys()) > 0:
                 async with self._db.acquire() as conn:
                     uids = list(self.game.mods.keys())
-                    await conn.execute(text(
-                        """ UPDATE mod_stats s JOIN mod_version v ON v.mod_id = s.mod_id
-                            SET s.times_played = s.times_played + 1 WHERE v.uid in :ids"""),
-                        ids=tuple(uids)
+                    await conn.execute(
```
Askaholic, line 497: Personally I wouldn't touch the modvault stuff. It's legacy code that's only used by the Python client and doesn't have good tests. 

If you were trying to replicate the query exactly I think it would be something like:
```python3
mod_stats.join(mod_version).update().values(
    times_played=mod_stats.c.times_played + 1
).where(mod_version.c.uid.in_uids))
```

SQLAlchemy will take care of the join condition as long as the table objects have been set up correctly.

server/lobbyconnection.py
```diff
@@ -197,7 +201,12 @@ def on_connection_made(self, protocol: QDataStreamProtocol, peername: Address):
     async def command_coop_list(self, message):
         """ Request for coop map list"""
         async with self._db.acquire() as conn:
-            result = await conn.execute("SELECT name, description, filename, type, id FROM `coop_map`")
+            result = await conn.execute(
+                select([
+                    coop_map.c.name, coop_map.c.description, coop_map.c.filename,
+                    coop_map.c.type, coop_map.c.id
+                ])
```
Askaholic, line 208:
```suggestion
                select([coop_map])
```
It will change the order, but that doesn't matter since we're accessing the results by column name anyways.

server/lobbyconnection.py
```diff
@@ -557,12 +579,14 @@ def _set_user_agent_and_version(self, user_agent, version):
...
             await conn.execute(
-                "UPDATE login SET ip = %(ip)s, user_agent = %(user_agent)s, last_login = NOW() WHERE id = %(player_id)s",
-                {
-                    "ip": self.peer_address.host,
-                    "user_agent": self.user_agent,
-                    "player_id": player_id
-                })
+                t_login.update().where(
+                    t_login.c.id == player_id
+                ).values(
```
Askaholic, line 584: Can we put `values` first and `where` at the end?

server/gameconnection.py
Askaholic: Actually, nvm, I just read https://docs.sqlalchemy.org/en/13/core/tutorial.html#multiple-table-updates. I feel like that way would be better, but I have no idea how `update from` compares performance wise

server/lobbyconnection.py
```diff
@@ -895,66 +951,121 @@ def _set_user_agent_and_version(self, user_agent, version):
 
         async with self._db.acquire() as conn:
             if type == "start":
-                result = await conn.execute("SELECT uid, name, version, author, ui, date, downloads, likes, played, description, filename, icon FROM table_mod ORDER BY likes DESC LIMIT 100")
+                result = await conn.execute(
+                    select([
+                        table_mod
```
Askaholic, line 956: This really doesn't need it's own line :P

server/lobbyconnection.py
```diff
@@ -895,66 +951,121 @@ def _set_user_agent_and_version(self, user_agent, version):
...
+                            command="modvault_info",
+                            thumbnail=thumbstr,
+                            link=link,
+                            bugreports=[],
+                            comments=[],
+                            date=int(row["date"].timestamp()),
+                        )
+                        out.update({
+                            key: row[key]
+                            for key in [
```
Askaholic, line 984: Use a tuple instead of a list!

server/lobbyconnection.py
Askaholic: https://docs.sqlalchemy.org/en/13/core/tutorial.html#multiple-table-updates

server/lobbyconnection.py
Askaholic: https://docs.sqlalchemy.org/en/13/core/tutorial.html#multiple-table-updates

tests/integration_tests/test_modvault.py
```diff
@@ -57,3 +57,21 @@
...
+    await read_until_command(proto, 'game_info')
+
+    await proto.send_message({
+        'command': 'modvault',
+        'type': 'download',
+        'uid': 'FFF'
+    })
+
+    # TODO I would like to assert that this increases `mod_stats.c.downloads`,
+    # but that is not reflected in `table_mod`. How are these related?
```
Askaholic, line 77: It would use `mod_version` right?

server/lobbyconnection.py
cleborys: Ah, that's nice, of course!

server/lobbyconnection.py
cleborys: I like this better than the long line, do you not?

server/lobbyconnection.py
cleborys: Here's `print(query)`:
```
SELECT avatars_list.url, avatars_list.tooltip 
FROM avatars LEFT OUTER JOIN avatars_list ON avatars_list.id = avatars."idAvatar" 
WHERE avatars."idUser" = :idUser_1
```

server/lobbyconnection.py
cleborys: That link is indeed what I was looking at.
Using `.where(and_(condition1, condition2))` also works, if you prefer it.

server/db/models.py
cleborys: Ah, that makes sense :P I didn't notice and just tried to group them roughly by "concern" and then add new ones at the end.

server/db/models.py
cleborys: I was looking at this:
http://faforever.github.io/db/tables/mod_stats.html
where the parent is `mod`, not `table_mod` - but I was rather confused why there is both, too.

server/lobbyconnection.py
cleborys: I grudgingly agree :P

server/gameconnection.py
cleborys: I'll revert everything that touches the mod tables I think^^ I've come to rely on the tests a lot.

cleborys:
Will #432 require refactoring the modvault queries anyways?
Even so it might be good to make that a different PR so it's easy to roll back if it ends up breaking the python client.

server/lobbyconnection.py
norraxx: Np :-)

server/db/models.py
norraxx: Well, sure, python could be column based language.... :-)

Askaholic:
That issue is a pretty massive undertaking. I sunk quite a few hours into it this summer, but never ended up "finishing" it. 

My hope is that we can find an opportunity to drop that modvault stuff completely. Maybe after TMM we will see more of a drop in python client users and I'll feel ok doing that.

server/db/models.py
Askaholic: Oh, `table_mod` is a view and `mod` is the actual table. Great choice of names...

PR #570, opened 2020-04-10
Reload Configuration Variables

### LINKED ISSUE #455: Adjust config parameters without server restart
Currently the config parameters are only read on server start (when `config.py` is imported) which means that adjusting them requires a server restart. It would be nice if we could adjust config values without all of the issues of restarting the server.

### Ideas
- Reload the config variables on a fixed time interval
    - Simple to implement
- Trigger config reload via some sort of communication protocol
    - More complicated
    - More security concerns
    - Could be used for triggering other server functionality such as scheduling impending downtime

cleborys:
Moves all configuration variables into an object, which will periodically reload them.
(Either every five minutes or whatever the according config variable is set to.)

I made the worker doing the refreshing into a service, although the only responsibility of that service is literally to start and shutdown that background thread - but I figured it fits nicely into the "start on server startup, clean up on shutdown" framework.

New config variable: `CONFIGURATION_REFRESH_TIME`, interval between refreshes in seconds. default five minutes.

I am 80% confident that config variables used as default arguments in functions will _not_ refresh properly but haven't checked. That would be easy to fix if desired, but might make the syntax a bit bloatier, see comment below.

Closes #455.


cleborys:
RE refreshing config variables that are default arguments in functions:
I think python handles default arguments basically as static variables that, I assume, are initialized when the function is defined. (E.g mutable default arguments retain their state in-between function calls.)

Hence to make something like
```
sign_into_twilio(token=config.TWILIO_TOKEN):
  do_stuff()
```
refresh correctly, I think the way to go would be
```
sign_into_twilio(token=None):
  if token is None:
    token = config.TWILIO_TOKEN
  do_stuff()
```
which I would add in all the places affected.

Another option would of course be to make the arguments mandatory, but that would change interfaces.

server/config.py
```diff
@@ -5,75 +5,105 @@
...
-
-# Team id that players in FFA mode are assigned to
-FFA_TEAM = 1
+
+class ConfigurationStore:
+    def __init__(self):
+        self.refresh()
+
+        self.WWW_URL = "https://www.faforever.com"
+        self.CONTENT_URL = "http://content.faforever.com"
```
Askaholic, line 22: Are these even used anymore?

cleborys: only in `server/lobbyconnection.py`

cleborys:
Now reads from a `yaml` file, with all defaults checked in as `conf.yaml`.  
You can set the `CONFIGURATION_FILE` environment variable to change the file path.

Note that if entries are dropped from the config file between refreshes, they will retain their previous value and that there is no longer any defaults hardcoded in python, besides the config file location.
That means the server will not run (or crash quickly) when given an incomplete config file.

cleborys:
Now reads from a `yaml` file, with all defaults checked in as `conf.yaml`.  
You can set the `CONFIGURATION_FILE` environment variable to change the file path.

Note that if entries are dropped from the config file between refreshes, they will retain their previous value and that there is no longer any defaults hardcoded in python, besides the config file location.
That means the server will not run (or crash quickly) when given an incomplete config file.

server/config.py
```diff
@@ -1,79 +1,31 @@
...
-GEO_IP_DATABASE_MAX_AGE_DAYS = int(os.getenv('GEO_IP_DATABASE_MAX_AGE_DAYS', 22))
+class ConfigurationStore:
+    def __init__(self):
+        self.refresh()
 
-CONTROL_SERVER_PORT = int(os.getenv('CONTROL_SERVER_PORT', 4000))
+    def refresh(self):
+        config_file = os.getenv("CONFIGURATION_FILE", "conf.yaml")
+        with open(config_file, "r") as f:
+            config_dict = yaml.safe_load(f)
```
Askaholic, line 25: Maybe we rename the conf file that's checked in to `default_conf.yaml` and do something like this:
```suggestion
        with open("default_conf.yaml") as f:
            config_dict = yaml.safe_load(f)
        config_file = os.getenv("CONFIGURATION_FILE")
        if config_file:
            with open(config_file) as f:
                config_dict.update(yaml.safe_load(f))
```

That way we'll always have a default value checked in to the repo.

server/config.py
```diff
@@ -1,79 +1,31 @@
...
-GEO_IP_LICENSE_KEY = os.getenv("GEO_IP_LICENSE_KEY")
-GEO_IP_DATABASE_MAX_AGE_DAYS = int(os.getenv('GEO_IP_DATABASE_MAX_AGE_DAYS', 22))
+class ConfigurationStore:
+    def __init__(self):
+        self.refresh()
 
-CONTROL_SERVER_PORT = int(os.getenv('CONTROL_SERVER_PORT', 4000))
+    def refresh(self):
+        config_file = os.getenv("CONFIGURATION_FILE", "conf.yaml")
+        with open(config_file, "r") as f:
```
Askaholic, line 24: Isn't `"r"` the default if you don't specify a mode?

server.py
```diff
@@ -107,6 +104,16 @@ async def run_profiler():
...
+    async def restart_control_server(old_port, new_port):
+        nonlocal ctrl_server
+        nonlocal services
+
+        ctrl_server.shutdown()
+        ctrl_server = await server.run_control_server(
+            services["player_service"],
+            services["game_service"]
+        )
+    config.register_callback("CONTROL_SERVER_PORT", restart_control_server)
```
cleborys, line 116: Alternatively, we could add a `ctrl_server.restart_on_port(new_port)` method.

server/__init__.py
```diff
@@ -51,6 +52,7 @@
 logger = logging.getLogger("server")
 
 if config.ENABLE_METRICS:
+    logger.info("Using prometheus on port: {}".format(config.METRICS_PORT))
```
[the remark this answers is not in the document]
cleborys: Urgh, how do I *keep* missing that?

server/config.py
```diff
@@ -1,79 +1,72 @@
...
-QUEUE_POP_DESIRED_PLAYERS = int(os.getenv('QUEUE_POP_DESIRED_PLAYERS', 8))
-# How many previous queue sizes to consider
-QUEUE_POP_TIME_MOVING_AVG_SIZE = int(os.getenv('QUEUE_POP_TIME_MOVING_AVG_SIZE', 5))
-
-# Team id that players in FFA mode are assigned to
-FFA_TEAM = 1
+
+class ConfigurationStore:
+    def __init__(self):
+        self._callbacks: Dict[str, Callable[[Any, Any], None]] = {}
```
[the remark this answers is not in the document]
cleborys: As you can see I also used to pass the old and new value as `[Any, Any]` and forgot to update that...
Maybe I'll just put `Callable`.

server/games/game.py
```diff
@@ -237,7 +237,7 @@ def is_ffa(self) -> bool:
         if len(self.players) < 3:
             return False
 
-        return FFA_TEAM in self.teams
+        return config.FFA_TEAM in self.teams
```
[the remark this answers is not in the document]
cleborys: Ah, now I got you.

server/geoip_service.py
```diff
@@ -34,6 +34,11 @@ def __init__(self):
 
         self.check_geoip_db_file_updated()
 
+        def refresh_file_path():
+            nonlocal self
```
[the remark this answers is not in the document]
cleborys: For some reason I was convinced that it would lose the `self` pointer if passing `self.refresh_file_path` as a callback, but of course that's nicer.

server/profiler.py
```diff
@@ -0,0 +1,93 @@
...
+        self._running = True
+        if self._task is None:
+            self._task = asyncio.create_task(Profiler._next_run(self))
+
+    async def _next_run(self):
+        await asyncio.sleep(self.interval)
+
+        if not self._running:
+            return
+        self.current_count += 1
```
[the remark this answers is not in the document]
cleborys: Right, I did that intentionally thinking that if the server is under high load for a long time, you don't want the profiling coroutine to keep running until the load drops (if it ever does - right now the player count is rarely dropping below 500 I think, not below 750 on weekends).
So this is rather the count of "attempts" than "runs".
We could reflect that in the variable names?

server/profiler.py
```diff
@@ -0,0 +1,93 @@
...
+            return
+
+        self._logger.info("Starting profiler")
+        self.profiler.enable()
+        await asyncio.sleep(self.duration)
+        self.profiler.disable()
+
+        self._logger.info("Done profiling %i/%i", self.current_count, self.max_count)
+        if self._outfile is not None:
+            self.profiler.dump_stats(self._outfile)
```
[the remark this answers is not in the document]
cleborys: I left that to garbage collection (since the instance of the wrapping `Profiler` class isn't referenced anywhere and only keeps scheduling itself it should go out of scope when it's done), but if it's that large it's maybe a good idea to do explicitly...

server/ice_servers/coturn.py
```diff
@@ -12,15 +12,23 @@
...
+from server.config import config
 
 
 class CoturnHMAC:
-    def __init__(self, coturn_hosts=COTURN_HOSTS, coturn_keys=COTURN_KEYS):
+    def __init__(self, coturn_hosts=None, coturn_keys=None):
+        if coturn_hosts is None:
+            coturn_hosts = config.COTURN_HOSTS
+        if coturn_keys is None:
+            coturn_keys = config.COTURN_KEYS
```
cleborys, line 23: Syntax would be nicer with e.g. `self.coturn_hosts = coturn_hosts or config.COTURN_HOSTS`, but I don't like how that would handle e.g. `coturn_hosts = []`.
`self.coturn_hosts = coturn_hosts if coturn_hosts is not None else config.COTURN_HOSTS` on the other hand is very mouthy.

Askaholic: Probably better to default to an empty list and pass in the variables form config explicitly when created.

server/profiler.py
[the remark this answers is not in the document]
Askaholic: The coroutine running isn't going to cause much load, calling `profile.enable()` might. Although I think it's actually not that bad in python since the profiler is implemented in C and python is already so slow. Might be worth bumping up the "high load" threshold since even at 2000 players we're only at about 20% cpu usage. I had it lower before because when I was awaiting broadcasts it was maxing out at about 1500.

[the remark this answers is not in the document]
Askaholic: I think only increment if the profiler actually runs. When I get the profile file I want it to contain the full amount of data regardless of when the profiler was started.

server/config.py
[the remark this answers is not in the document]
Askaholic: So are you going to change it here too?

tests/unit_tests/test_api_accessor.py
```diff
@@ -38,7 +38,7 @@ async def test_api_patch(api_accessor):
     result = await api_accessor.api_patch('test', data)
     api_accessor.api_session.session.request.assert_called_once_with(
         "PATCH",
-        API_BASE_URL+'test',
+        config.API_BASE_URL+'test',
```
Askaholic, line 41:
```suggestion
        config.API_BASE_URL + 'test',
```

tests/data/test_conf.yaml
```diff
@@ -0,0 +1,55 @@
...
+GEO_IP_DATABASE_MAX_AGE_DAYS: 22
+
+LADDER_ANTI_REPETITION_LIMIT: 3
+LADDER_SEARCH_EXPANSION_MAX: 0.25
+LADDER_SEARCH_EXPANSION_STEP: 0.05
+# The maximum amount of time (in seconds) to wait between pops.
+QUEUE_POP_TIME_MAX: 180
+# The number of players we would like to have in the queue when it pops. The
+# queue pop time will be adjusted based on the current rate of players queuing
+# to try and hit this number.
```
Askaholic, line 52: Do we need to duplicate the comments in the test config file?

server/config.py
```diff
@@ -1,79 +1,122 @@
...
-
-# Team id that players in FFA mode are assigned to
-FFA_TEAM = 1
+
+class ConfigurationStore:
+    def __init__(self):
+        """
+        Change default values here.
+        """
+        self.CONFIGURATION_REFRESH_TIME = 300
```
Askaholic, line 29: Maybe you want to put spaces in between related options like before

server.py
```diff
@@ -17,14 +17,12 @@
...
+from server.config import config
 from docopt import docopt
 from server.api.api_accessor import ApiAccessor
-from server.config import (
-    DB_LOGIN, DB_NAME, DB_PASSWORD, DB_PORT, DB_SERVER, TWILIO_ACCOUNT_SID
-)
 from server.core import create_services
 from server.ice_servers.nts import TwilioNTS
+from server.profiler import get_profiler_factory
 from server.timing import at_interval
```
Askaholic, line 26: Unused import

server.py
```diff
@@ -81,37 +74,24 @@ def signal_handler(sig: int, _frame):
...
-            pr.dump_stats("profile.txt")
+    make_profiler = get_profiler_factory(services["player_service"])
+    await make_profiler()
+    config.register_callback("PROFILING_INTERVAL", make_profiler)
 
     ctrl_server = await server.run_control_server(
         services["player_service"],
         services["game_service"]
     )
+    async def restart_control_server():
```
Askaholic, line 85: Should be a blank line before function definition

server.py
```diff
@@ -81,37 +74,24 @@ def signal_handler(sig: int, _frame):
...
+        nonlocal ctrl_server
+        nonlocal services
+
+        ctrl_server.shutdown()
```
Askaholic, line 89: shutdown is a coroutine
```suggestion
        await ctrl_server.shutdown()
```

server/configuration_service.py
```diff
@@ -0,0 +1,29 @@
...
+        self._store = config
+        self._task = None
+
+    async def initialize(self) -> None:
+        self._task = asyncio.create_task(self._worker_loop())
+        self._logger.info("Configuration service started.")
+
+    async def _worker_loop(self) -> None:
+        while True:
+            self._logger.info("Refreshing configuration variables")
```
Askaholic, line 21: I think I would put this on `debug` log level and instead log when changes are detected. So for instance if I change `QUEUE_POP_TIME_MAX` it would log something like:
```
New value for QUEUE_POP_TIME_MAX: 180 -> 200
```

Askaholic:
Some things when testing:

- If I set the config file path when starting but don't set the file I get a file not found error.
- If I set the file path and the file is empty I get nonetype error:
```
    new_values.update(yaml.safe_load(f))
TypeError: 'NoneType' object is not iterable
```
It would be nice if either of those situations were treated the same as no file being set, since maybe you want to run the server with default params, but still be able to change them later by creating the config file where you said it would be.

And it looks like you forgot to remove the `await` on the profiler factory.
```
  File "server.py", line 143, in <module>
    asyncio.run(main())
  File "/usr/local/lib/python3.7/asyncio/runners.py", line 43, in run
    return loop.run_until_complete(main)
  File "/usr/local/lib/python3.7/asyncio/base_events.py", line 583, in run_until_complete
    return future.result()
  File "server.py", line 78, in main
    await make_profiler()
TypeError: object NoneType can't be used in 'await' expression
```

- Once a profiler is started, changing the profiling interval back to `-1` does not stop it again.

server.py
```diff
@@ -81,38 +73,26 @@ def signal_handler(sig: int, _frame):
...
-            pr.enable()
-            await asyncio.sleep(2)
-            pr.disable()
-            profiled_count += 1
-
-            logging.info("Done profiling %i/%i", profiled_count, max_count)
-            pr.dump_stats("profile.txt")
+    profiler = Profiler(services["player_service"])
+    profiler.refresh()
+    config.register_callback("PROFILING_INTERVAL", profiler.refresh)
```
Askaholic, line 78: Maybe also want to refresh on `PROFILING_COUNT` and `PROFILING_DURATION`?

PR #574, opened 2020-04-14
Refactor game options

Askaholic:
- Store all game options in the game options dictionary.
- Remove warning for unused game options.
- Remove workaround for `AIReplacement` on/off state.
- Make sure only the host can configure the game.
- Update dependencies.

### DIFF
--- a/server/gameconnection.py
+++ b/server/gameconnection.py
@@ -74,6 +74,15 @@ class GameConnection(GpgNetServerProtocol):
     def player(self, val: Player):
         self._player = val
 
+    def is_host(self) -> bool:
+        if not self.game or not self.player:
+            return False
+
+        return (
+            self.player.state == PlayerState.HOSTING and
+            self.player == self.game.host
+        )
+
     async def send(self, message):
         """
         Send a game message to the client.
@@ -210,6 +219,9 @@ class GameConnection(GpgNetServerProtocol):
         self.game.desyncs += 1
 
     async def handle_game_option(self, key, value):
+        if not self.is_host():
+            return
+
         if key == "Victory":
             self.game.gameOptions["Victory"] = Victory.from_gpgnet_string(value)
         else:
@@ -230,6 +242,9 @@ class GameConnection(GpgNetServerProtocol):
         self._mark_dirty()
 
     async def handle_game_mods(self, mode, args):
+        if not self.is_host():
+            return
+
         if mode == "activated":
             # In this case args is the number of mods
             if int(args) == 0:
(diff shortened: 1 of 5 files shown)


server/gameconnection.py
```diff
@@ -74,6 +74,15 @@ def player(self) -> Player:
     def player(self, val: Player):
         self._player = val
 
+    def is_host(self) -> bool:
+        if not self.game or not self.player:
```
cleborys, line 78: Maybe 
```python
if self.game is None or self.player is None:
```

server/gameconnection.py
```diff
@@ -210,41 +219,32 @@ async def handle_desync(self, *_args):  # pragma: no cover
...
-            that makes "On" and "Off" be "AIReplacementOn" and "AIReplacementOff". The code
-            below removes that extra statement to make it a simple "On" "Off".
-            This block can be removed as soon as the game sends "On" and "Off" instead of
-            "AIReplacementOn" and "AIReplacementOff" to the server as game options.
-            https://github.com/FAForever/fa/issues/2610
-            """
-            if key == "AIReplacement":
-                value = value.replace("AIReplacement", "")
+        if not self.is_host():
+            return
```
cleborys, line 223: All of these `if not self.is_host()` checks might work well in  a `@host_only` decorator.
(Although it's slightly cumbersome to give them access to `self`, see [here](https://medium.com/@vadimpushtaev/decorator-inside-python-class-1e74d23107f6).)

tests/unit_tests/test_gameconnection.py
```diff
@@ -317,6 +359,19 @@ async def test_handle_action_GameOption(game: Game, game_connection: GameConnect
     assert game.map_file_path == "maps/some_map.zip"
     await game_connection.handle_action('GameOption', ['Title', 'All welcome'])
     assert game.name == game.sanitize_name('All welcome')
+    await game_connection.handle_action('GameOption', ['ArbitraryKey', 'ArbitraryValue'])
+    assert game.gameOptions['ArbitraryKey'] == 'ArbitraryValue'
+
+
+async def test_handle_action_GameOption_not_hots(
```
cleborys, line 366:
```suggestion
async def test_handle_action_GameOption_not_host(
```

cleborys: [ instert random "got the hots" pun ]

server/gameconnection.py
```diff
@@ -210,41 +219,32 @@ async def handle_desync(self, *_args):  # pragma: no cover
...
+        if key == "Victory":
+            self.game.gameOptions["Victory"] = Victory.from_gpgnet_string(value)
+        else:
             self.game.gameOptions[key] = value
 
         if key == "Slots":
             self.game.max_players = int(value)
-        elif key == 'ScenarioFile':
+        elif key == "ScenarioFile":
             raw = "%r" % value
```
cleborys, line 233: I find the following easier to read:
```suggestion
            raw = repr(value)
```
I am also not sure why `repr` instead of `str` - is `value` ever anything else but a string or a number?

server/gameconnection.py
Askaholic: I decided not to do a decorator because the method signatures of the handlers are actually used for command validation. Adding a decorator would change the function signature of every handler to the same thing `func(self, *args)` which I don't want.

server/gameconnection.py
Askaholic: I have no idea. Possibly a dict?

PR #585, opened 2020-05-17
Update dependencies and adjust code

### LINKED ISSUE #584: Update dependencies
Will need to fix one of the database fixtures as it always breaks on the last test with the newest version of `pytest-asyncio`.

Askaholic:
Adjusting a few fixtures to get around the event loop being closed problem. I also ignore some deprecation warnings in the tests because they do not cause us any issues.

Closes #584.


tests/conftest.py
```diff
@@ -46,55 +45,68 @@ def pytest_configure(config):
...
 @pytest.fixture(scope='session', autouse=True)
-def global_database(request):
-    return _database(request, asyncio.get_event_loop(), False)
+async def test_data(request):
+    db = await global_database(request)
+    with open('tests/data/test-data.sql') as f:
+        async with db.acquire() as conn:
+            await conn.execute(f.read())
 
+    await db.close()
```
norraxx, line 60: Well, I wish, we work with database without dirtying it.
Like it's here.
https://github.com/norraxx/faf-aio-replayserver/blob/master/tests/fixtures/db_fixtures.py#L7

tests/conftest.py
```diff
@@ -46,55 +45,68 @@ def pytest_configure(config):
...
-            db=db
-        )
+    host, user, pw, db, port = (
+        opt('--mysql_host'),
+        opt('--mysql_username'),
+        opt('--mysql_password'),
+        opt('--mysql_database'),
+        opt('--mysql_port')
+    )
+    fdb = FAFDatabase(asyncio.get_running_loop())
```
norraxx, line 73: name  variable properly, please

tests/conftest.py
norraxx: And ofc IF we have `.commit()` in code, we should mock it.

tests/conftest.py
```diff
@@ -46,55 +45,68 @@ def pytest_configure(config):
...
+    host, user, pw, db, port = (
+        opt('--mysql_host'),
+        opt('--mysql_username'),
+        opt('--mysql_password'),
+        opt('--mysql_database'),
+        opt('--mysql_port')
+    )
+    fdb = MockDatabase(event_loop)
+
+    await fdb.connect(
```
norraxx, line 99: -//- same naming.

tests/unit_tests/test_matchmaker_queue_algorithm.py
```diff
@@ -152,6 +152,7 @@ def test_match_graph_will_not_include_matches_below_threshold_quality(player_fac
     }
 
 
+@pytest.mark.filterwarnings("ignore:.*'caplog' fixture")
```
norraxx, line 155: Maybe, you could make ticket To solve this warnings.

tests/conftest.py
Askaholic: The database connection used during tests gets rolled back. This one is just for loading the test-data file once at the start of the session. And for some reason with the newest version of pytest-asyncio having an asyc finalizer on this session scoped fixture would throw an error due to closed event loop.

tests/conftest.py
Askaholic: ...

tests/unit_tests/test_matchmaker_queue_algorithm.py
Askaholic: There isn't anything to fix. We just ignore the warning because we know calling `set_log_level` multiple times is fine. https://github.com/HypothesisWorks/hypothesis/issues/377

tests/conftest.py
norraxx: Do we also rollback these changes?

tests/unit_tests/test_matchmaker_queue_algorithm.py
norraxx: Ok, np. I'm thinking if we should leave the link in a comment on this line? `# https://github.com/HypothesisWorks/hypothesis/issues/377`?

Askaholic: Yea that’s a good idea

tests/conftest.py
Askaholic: No, the test data is not rolled back

norraxx: `=T_T=`

Issue #579, opened 2020-04-28
Travis set up database from faf-db directly


Askaholic:
Might make specifying the database version more convenient. Instead of pointing to a FAF stack version we can point to a db release directly.

See here: https://github.com/FAForever/faf-java-api/commit/2a29afbe2649c748bcf35574cea5be6a8626e207

PR #586, opened 2020-05-17
Broadcast Game Results to RabbitMQ

### LINKED ISSUE #582: Broadcast game results via RabbitMQ
If a game ends, please broadcast a message into a RabbitMQ exchange.
Interested clients can subscribe their own queue to it and listen for it.

Who needs it?
- Replay server to know that it should write a non-final version of the replay stream to disk
- Galactic War for parsing the outcome
- Potential rating services outside the lobby server itself
- Idea: Potential ticker on the website showing a live log of ladder or something

cleborys:
Sends info about game results to a RabbitMQ Exchange.
The message format is specified in `server/games/typedefs` in the `to_dict` method of `EndedGameInfo`, currently as follows (in python-ish pseudo code):
```python
{
  "game_id": int,
  "rating_type": str = self.rating_type.name if self.rating_type is not None else "None",
  "map_id": int,
  "featured_mod": str,
  "sim_mod_ids": List[int],
  "commander_kills": Dict[str, int],
  "validity": str = self.validity.name,
  "teams": [
    {
      "outcome": str = team_summary.outcome.name,
      "player_ids": List[int] = list(team_summary.player_ids),
     }
     for team_summary in self.team_summaries
  ],
}
```

To run the tests locally, you should now also start the RabbitMQ container with the script `.ci/init-rabbitmq.sh`.

Closes #582 

ToDo list:
 - [x] Publisher confirms
 - add ACU kills from stats (only have number of kills)
 - [x] non-featured mods
 - [x] handle connection failure / incorrect credentials
 - [x] figure out a better way to start RabbitMQ for testing


server/games/typedefs.py
```diff
@@ -0,0 +1,177 @@
...
+
+@unique
+class Victory(Enum):
+    DEMORALIZATION = 0
+    DOMINATION = 1
+    ERADICATION = 2
+    SANDBOX = 3
+
+    @staticmethod
+    def from_gpgnet_string(value: str) -> Optional["Victory"]:
```
Askaholic, line 25: I learned recently that this can be done more easily with `Victory[value.upper()]` (maybe need to catch `KeyError` depending on what the caller bahavior is)

server/__init__.py
```diff
@@ -21,10 +21,13 @@
 from .protocol import QDataStreamProtocol
 from .servercontext import ServerContext
 from .configuration_service import ConfigurationService  # noqa: F401
+from .message_queue_service import MessageQueueService  # noqa: F401
 from .geoip_service import GeoIpService
 from .player_service import PlayerService
 from .game_service import GameService
 from .ladder_service import LadderService
+from .rating_service.rating_service import RatingService
+from .message_queue_service import MessageQueueService
```
Askaholic, line 30: Double import?

server/__init__.py
```diff
@@ -68,6 +71,8 @@ async def run_lobby_server(
     nts_client: Optional[TwilioNTS],
     geoip_service: GeoIpService,
     ladder_service: LadderService,
+    rating_service: RatingService,
+    message_queue_service: MessageQueueService,
```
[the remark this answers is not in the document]
cleborys: I thought this was necessary for the `Service` metaclass to know about them so they can be used in the dependency injector?

Askaholic:
I see. Importing them should be enough. Don't need to pass them to the run function.

‐‐‐‐‐‐‐ Original Message ‐‐‐‐‐‐‐

cleborys:
Yes, when replied I though that comment was an the imports - I'm not sure
how it happened that I was passing them, maybe I got confused during the
rebase.

cleborys:
I decided that transactions for publishing are easier than "publisher confirms" (https://www.rabbitmq.com/confirms.html#publisher-confirms). They have way lower throughput (apparently up to 250x) but I highly doubt that publishing to RabbitMQ will anywhere near our bottleneck and this way we're guaranteed not to lose any messages. If it becomes an issue, it's easy to switch to fire-and-forget publishing and then we can still think about implementing "publisher confirms" if needed.

In the sample stats data that I could find, the best I can see is a count of ACU kills for every player, so I included that (which necessitates moving the JSON parsing from the stats service to the `Game` class, before it was simply saving and passing a string).

.ci/init-rabbitmq.sh
```diff
@@ -0,0 +1,13 @@
...
+
+source .ci/faf-rabbitmq.env
+docker run --rm -d -p 5672:5672 --env-file .ci/faf-rabbitmq.env --name faf-rabbitmq rabbitmq:3.8.2-management-alpine
+
+# Create RabbitMQ users
+docker exec faf-rabbitmq rabbitmqctl wait --timeout ${MAX_WAIT} "${RABBITMQ_PID_FILE}"
+
+docker exec faf-rabbitmq rabbitmqctl add_vhost "${RABBITMQ_LOBBYSERVER_VHOST}"
+docker exec faf-rabbitmq rabbitmqctl add_user "${RABBITMQ_LOBBYSERVER_USER}" "${RABBITMQ_LOBBYSERVER_PASS}"
+docker exec faf-rabbitmq rabbitmqctl set_permissions -p "${RABBITMQ_LOBBYSERVER_VHOST}" "${RABBITMQ_LOBBYSERVER_USER}" ".*" ".*" ".*"
```
norraxx, line 13: imho, this should be solved by `faf-stack`, like it is solved with databse. see `.travis.yml`  `./scripts/init-db.sh`

Pipfile
```diff
@@ -24,12 +24,13 @@ twilio = "*"
...
 aiomysql = {editable = true, git = "https://github.com/aio-libs/aiomysql"}
 pyyaml = "*"
+aio_pika = "*"
 
 [dev-packages]
 pytest = "*"
 pytest-mock = "*"
 pytest-cov = "*"
-pytest-asyncio = "*"
+pytest-asyncio = "==0.10.0"
```
norraxx, line 33: Reason?

server/games/game.py
```diff
@@ -764,8 +774,8 @@ def get_player_outcome(self, player):
 
         return self._results.outcome(army)
 
-    def report_army_stats(self, stats):
-        self._army_stats = stats
+    def report_army_stats(self, stats_json):
+        self._army_stats_list = json.loads(stats_json)["stats"]
         self._process_pending_army_stats()
```
cleborys, line 779: I am hoping that this isn't called to often, so the json parsing doesn't happen too often, ideally only once at game end (or at least only once per player).
Let me know if that's not the case, then this should probably be saved as string and only parsed on-demand.

server/gameconnection.py
```diff
@@ -223,7 +223,10 @@ async def handle_game_option(self, key, value):
             return
 
         if key == "Victory":
-            self.game.gameOptions["Victory"] = Victory.from_gpgnet_string(value)
+            try:
+                self.game.gameOptions["Victory"] = Victory[value.upper()]
+            except KeyError:
+                self.game.gameOptions["Victory"] = None
```
norraxx, line 229: `Victory.__members__.get("DEMORALIZATION", None)`

server/games/game.py
```diff
@@ -260,7 +263,7 @@ def _process_pending_army_stats(self):
     def _process_army_stats_for_player(self, player):
         try:
             if (
-                self._army_stats is None
+                len(self._army_stats_list) == 0
```
norraxx, line 266: `if self._army_stats_list` ;-)

.ci/init-rabbitmq.sh
Brutus5000: No it should not. https://github.com/FAForever/server/issues/579
Tying to faf-stack brings severe issues in the release process as you have a circular dependency.

server/game_service.py
```diff
@@ -210,5 +216,22 @@ def __getitem__(self, item: int) -> Game:
...
+            "game_results",
+            ".".join([
+                "results",
+                result_dict["rating_type"],
+                result_dict["featured_mod"],
+                result_dict["validity"],
+            ]),
+            result_dict,
+        )
+
```
Brutus5000, line 231: Selecting a useful topic is not easy :smile: What is contained in rating_type?

In general at work we came up with the following approach:

## "The shitting camels" 
Use dots to distinguish sub-parts, use camel case between the dots.
(The dots allow wildcard filtering on certain parts of the routing key)

Example: request.meterData.create

**Routing key names:**
Schema: \<messageType>.\<entity>.\<operation>

**Message type** can be:
- request → messages intending to perform a state change
- success → confirmation/notification of a state change
- error → notification about a denied state change
  an error is like a 4xx HTTP status code: inform the sender about an error
  an error is not like 5xx HTTP status code: in case of "internal" errors, use the dead letter queue

**Entity** describes the object that is being operated on

**Operation**:
  - create
  - update
  - delete
  - use update if no distinction is possible (e.g. on ingestion)
  - \<to be discussed>

**Examples:**
- request.meter.create → success.meter.create
- request.meter.update → error.meter.update

Pipfile
cleborys: The current version breaks some other dependency, this should be temporary until that is resolved.

server/gameconnection.py
cleborys: Hm, I like this, but I'm not sure whether we want to rely on the `__members__` implementation. As far as I'm aware the library retains the right to change them as they see fit, as if they were private in other programming languages.

cleborys:
Concerning how to start rabbitmq:
Since we don't want to depend on `faf-stack` anymore, I guess we also don't want to start the container from there.
There is a `pytest-rabbitmq` module, which provides starting the rabbitmq process as a session-scoped fixture, but it has a few drawbacks: Firstly it doesn't provide a way to recreate the same setup as in faf-stack (running on a vhost, using a custom user), secondly and more importantly it relies on you having rabbitmq installed in the appropriate places (or at least that is what it looks like to me). I much prefer running rabbitmq from a public container repo, so everyone including travis has the same version and doesn't need to install it themselves.

Instead of running the init-script before running the tests, I'm now calling it in a session-scoped fixture (and an appropriate teardown script when the session ends.)
The drawbacks I can see are 1. possibly requiring more rights to run the tests, if your docker is installed such that you have to call it with sudo and 2. significant delays on hitting the first test that requires rabbitmq running (a couple of seconds, normally, but a _lot_ more if you don't yet have the rabbitmq container).

The benefits of the fixture over the manually run script I think is only that you wouldn't forget to run it and that it's cleaned up after exiting.
With either we obviously still need to maintain the `.ci/inti-rabbitmq.sh` script, but luckily it's not too much to maintain, only really the user information and host path.
Currently I don't see a good way around that though.

cleborys:
Tests that require a RabbitMQ server to run are now moved to `integration_tests/test_message_queue.py` and only those that don't will be run with the regular tests from `tests/unit_tests/test_message_queue.py` and via travis.

To set up RabbitMQ one should run `scripts/init-rabbitmq.sh` from faf-stack first, like for the database. I am unsure where to document that.

cleborys:
Tests that require a RabbitMQ server to run are now moved to integration_tests/test_message_queue.py and only those that don't will be run with the regular tests from tests/unit_tests/test_message_queue.py and via travis.

To set up RabbitMQ one should run `init-rabbitmq.sh` in `integration_tests/test_message_queue`. I am unsure where to document that.

server/gameconnection.py
Askaholic: I think it's fine. I read through the Enum source code a while back and I think it's mostly a dunder to signify that it is a special member and not an enum variant. There are some other ones too like `__order__` and they are meant to be used by client code. In python usually having underscores in the front and back means `magic` not `private` (like `__init__`, `__eq__`, etc).

server/__init__.py
```diff
@@ -20,7 +20,9 @@
 from .lobbyconnection import LobbyConnection
 from .protocol import QDataStreamProtocol
 from .servercontext import ServerContext
-from .configuration_service import ConfigurationService  # noqa: F401
+from .configuration_service import ConfigurationService   # noqa: F401
+from .message_queue_service import MessageQueueService    # noqa: F401
+from .rating_service.rating_service import RatingService  # noqa: F401
```
Askaholic, line 25: Add them to `__all__`

server/game_service.py
```diff
@@ -210,5 +216,17 @@ def __getitem__(self, item: int) -> Game:
...
-        await self._rating_service.enqueue(summary)
+    async def publish_game_results(self, game_results: EndedGameInfo):
+        result_dict = game_results.to_dict()
+        await self._message_queue_service.publish(
+            "faf-rabbitmq",
+            "success.gameResults.create",
+            result_dict,
+        )
+
+        # To be removed when rating service starts listening to message queue
```
Askaholic, line 227: Make it a `TODO`?

server/games/game.py
Askaholic: I think `report_army_stats` is called from `GameConnection` each time a player sends army stats. I'm also pretty sure that army stats are reported in a similar way to game results where every player reports stats for every other player. The  difference is it might happen only at game end.

server/games/ladder_game.py
```diff
@@ -2,9 +2,9 @@
 
 from server.abc.base_game import InitMode
 from server.players import Player
+from server.rating import RatingType
```
Askaholic, line 5: why?

server/message_queue_service.py
```diff
@@ -0,0 +1,124 @@
...
+
+
+@with_logger
+class MessageQueueService(Service):
+    def __init__(self) -> None:
+        """
+        Service handling connection to the message queue
+        and providing an interface to publish messages.
+        """
+        self._logger.info("Message queue service created.")
```
Askaholic, line 21: TBH I'm not sure if all these creation messages are necessary. They kindof just clutter the log output and make it harder to see actual important messages like "missing geoip database" or something. I'd either remove it entirely or demote it to `DEBUG` level.

Askaholic: (Same for rating serviceI

server/message_queue_service.py
```diff
@@ -0,0 +1,124 @@
...
+        if exchange is None:
+            raise KeyError(f"Unknown exchange {exchange_name}.")
+
+        message = aio_pika.Message(
+            json.dumps(payload).encode(), delivery_mode=delivery_mode
+        )
+
+        async with self._channel.transaction():
+            await exchange.publish(message, routing_key=routing)
+            self._logger.debug(
```
Askaholic, line 115: Hmm, maybe `TRACE` log level?

tests/unit_tests/test_game.py
```diff
@@ -861,3 +849,56 @@ async def test_team_sets_missing_team_disallowed(game: Game, game_add_players):
...
+    await game.add_result(host_id, 0, "victory", 1)
+    await game.add_result(join_id, 1, "defeat", 0)
+
+    game_results =  await game.resolve_game_results()
+    result_dict = game_results.to_dict()
+
+    assert result_dict["validity"] == "VALID"
+    assert result_dict["rating_type"] == "GLOBAL"
+    assert len(result_dict["teams"]) == 2
+    print(result_dict)
```
Askaholic, line 872: Need?

server/games/game.py
cleborys: I did that, and then only noticed that it actually should be `if not self._army_stats_list or ...` when the tests failed. That promptly reminded me of why I find the `len(stuff) == 0` more readable.

server/gameconnection.py
cleborys: Alright!

tests/unit_tests/test_message_queue_service.py
```diff
@@ -0,0 +1,86 @@
...
+async def test_publish(mq_service):
+    payload = {"msg": "test message"}
+    exchange_name = "test_exchange"
+    routing_key = "test.routing.key"
+    delivery_mode = aio_pika.DeliveryMode.NOT_PERSISTENT
+
+    await mq_service.publish(exchange_name, routing_key, payload, delivery_mode)
+
+
+async def test_incorrect_credentials(mocker):
```
Askaholic, line 41: I think you can use the `caplog` fixture to assert that certain warning messages appear in the log

Askaholic: Then you don't need to mock the logger

tests/unit_tests/test_game.py
cleborys: :fearful:

tests/unit_tests/test_message_queue_service.py
cleborys: I don't think that exists in the current stable pytest version yet (5.4.3)?

Askaholic: Well the `caplog` fixture does exist, I'm using it in some of the tests to change the log level. I haven't tried to assert contents of log messages though

cleborys: Yes, I noticed, I just followed the link and then clicked on the "latest stable version" banner, which tries to get you to the logging page of 5.4.3 which doesn't exist, so I wrongly assumed that's because they added that feature.
Just changed it, works like a charm.

server/games/game.py
cleborys: As long as it's only called at the game end I think that's fine? I just didn't want to do unnecessary parsing of partial stats every minute if they are going to be overwritten anyways.

The parsing needed to be moved so that stats can be included in the `gameResults` message.

server/message_queue_service.py
```diff
@@ -0,0 +1,126 @@
...
+    async def initialize(self) -> None:
+        if self._connection is not None:
+            return
+
+        await self._connect()
+
+    async def _connect(self) -> None:
+        try:
+            self._connection = await aio_pika.connect_robust(
+                "amqp://{user}:{password}@localhost:{port}/{vhost}".format(
```
Askaholic, line 41: Are there supposed to be two slashes between port and `vhost`? The `MQ_VHOST` variable starts with a slash.

server/message_queue_service.py
```diff
@@ -0,0 +1,126 @@
...
+                    user=config.MQ_USER,
+                    password=config.MQ_PASSWORD,
+                    vhost=config.MQ_VHOST,
+                    port=config.MQ_PORT,
+                ),
+                loop=asyncio.get_running_loop(),
+            )
+        except ConnectionError:
+            self._logger.warning("Unable to connect to RabbitMQ. Is it running?")
```
Askaholic, line 50: I would include the exception trace here so it's obvious something went wrong. Otherwise it's really easy to miss this warning.

server/message_queue_service.py
```diff
@@ -0,0 +1,126 @@
...
+            return
+        except ProbableAuthenticationError:
+            self._logger.warning(
+                "Unable to connect to RabbitMQ. Incorrect credentials?"
```
Askaholic, line 54: Same here

server/message_queue_service.py
```diff
@@ -0,0 +1,126 @@
...
+            )
+            return
+        except Exception as e:
+            self._logger.warning(
+                "Unable to connect to RabbitMQ due to unhandled excpetion %s. Incorrect vhost?",
+                e,
```
Askaholic, line 60: And I think same here

server/message_queue_service.py
cleborys: Indeed it's weird, but that's what was done for `/faf-postal` so I did the same. 
https://github.com/FAForever/faf-stack/blob/b6582560c77ddce549c0e099b1265b0d481c6805/config.template/faf-postal/faf-postal.env#L11

Issue #587, opened 2020-05-21
Workaround for game reporting bad results


Askaholic:
It seems that stopping the game from reporting bad results is not so easily done as it is caused by something in the engine and not the lua code. So we will probably need to work around this on the server side :(

My proposition is to try to figure out which player is sending the bad results and then simply throw out all results reported by that player. Maybe we could even throw out only results reported past a certain point in time if we can automatically identify that a user has left the game early and is reporting bogus `defeat`.

### How can we determine that results are bad?
When a player leaves with `Exit to windows`, their game very rapidly sends `defeat` for all players in the game (except for themselves) and then sends `GameState: Ended` signalling that FA has terminated. If we track time stamps for when a result is reported we could presumably define a window around when `GameState: Ended` is reported during which game results will be ignored. Maybe 1-2 seconds would be a good amount of time.

### Checklist
- [ ] Track timestamps for game results
- [x] Track which player reported a result
- [ ] Track timestamps for `GameState: Ended` or "disconnection" from the game
- [ ] Throw out a player's reports if they happen too close to a disconnection (1-2 seconds) and the disconnect happens before `GameEnded` is reported.
- [ ] Adjust tests to report game results at specific times.
- [ ] Gather some test data from real games

### Related 
#580 
https://github.com/FAForever/fa/issues/3087

Askaholic:
Not going to do this

Issue #580, opened 2020-04-29
Some ladder games are not being ranked


ChaosWarrior80:
Hi
I played some ladder games yesterday 28.04.2020 and noticed that 3 of them have "Rating change is not yet available" status on them.
https://content.faforever.com/faf/vault/replay_vault/replay.php?id=11601739
https://content.faforever.com/faf/vault/replay_vault/replay.php?id=11602403
https://content.faforever.com/faf/vault/replay_vault/replay.php?id=11605225

Askaholic:
Related: https://github.com/FAForever/fa/issues/3087

PR #588, opened 2020-05-23
Unset lobby_connection reference to game_connection on abort

Askaholic:
Should be able to solve #580 with this small change and another really small change in the FA code (sending `GameState: Ended` when `SafeQuit` is triggered). I think this is a better solution (because its so much simpler) than what I described in #587.

When investigating the `SafeQuit` issue, I noticed that `GameConnection` objects would not be deleted after games ended because `lobby_connection` would still hold a reference to them. This reference would only be removed once a new game was started or the player disconnected from the server. Since `GameConnection` holds a strong reference to `Game` this also means that `Game` objects would not be deleted for a possibly very long time.

Another possible solution would be to make `LobbyConnection` only hold a weak reference to `GameConnection`.

### DIFF
--- a/server/lobbyconnection.py
+++ b/server/lobbyconnection.py
@@ -124,7 +124,6 @@ class LobbyConnection:
             )
         if self.game_connection:
             await self.game_connection.abort()
-            self.game_connection = None
 
         await self.protocol.close()
 
--- a/tests/integration_tests/test_game.py
+++ b/tests/integration_tests/test_game.py
@@ -313,3 +313,92 @@ async def test_ladder_game_not_joinable(lobby_server):
         "style": "error",
         "text": "The game cannot be joined in this way."
     }
+
+
+@fast_forward(30)
+async def test_gamestate_ended_clears_references(
+    lobby_server,
+    game_service,
+    player_service
+):
+    test_id, _, test_proto = await connect_and_sign_in(
+        ("test", "test_password"), lobby_server
+    )
+    rhiza_id, _, rhiza_proto = await connect_and_sign_in(
+        ("Rhiza", "puff_the_magic_dragon"), lobby_server
+    )
+    await read_until_command(test_proto, "game_info")
+    await read_until_command(rhiza_proto, "game_info")
+
+    # Set up the game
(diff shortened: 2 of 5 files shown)


server/lobbyconnection.py
```diff
@@ -124,7 +124,6 @@ async def abort(self, logspam=""):
             )
         if self.game_connection:
             await self.game_connection.abort()
-            self.game_connection = None
```
cleborys, line 127: If a client sends a `GameState ended` message and immediately afterwards a wrong game result, it feels like it could happen that the first message triggers the `game_connection.abort()`, then the thread sleeps before the game_connection is set to `None` and then the second messages' result still gets set.
I think that's very unlikely, given that the result thread will also `await` something, but at this point it feels like we're relying on asyncio internals a bit - what if there are a lot more `await`s in the `game_connection.abort()` than in the result processing?

So maybe some flag should still be set immediately here in `lobbyconnection.py`, or wherever `GameState ended` is processed. (Or unset the `game_connection` reference, but then it's slightly awkward to call `abort()` afterwards.)

Whether you think that's an issue or not, maybe a test checking that when you send `GameState ended` and then immediately a result, the result doesn't get saved is a good idea?

tests/integration_tests/test_game.py
```diff
@@ -272,3 +272,55 @@ async def test_partial_game_ended_rates_game(lobby_server, tmp_user):
     # The game should only be rated once
     with pytest.raises(asyncio.TimeoutError):
         await read_until_command(host_proto, "player_info", timeout=10)
+
+
+@fast_forward(5)
+async def test_gamestate_ended_clears_references(
```
norraxx, line 278: btw: I have practice at work, where I'm describing the problem that you're trying to test.
I know, that test naming not every time can be descriptive.
Like it's here: https://github.com/norraxx/faf-aio-replayserver/blob/master/tests/server/unitttests/test_replay.py#L47

server/lobbyconnection.py
Askaholic: I'll add some game results to the test

PR #591, opened 2020-05-27
Feature/#440 Accept authentication via OAuth token [new]

Askaholic:
Extracts the relevant parts of #473 (that PR has a lot of refactors which are unrelated).

Adds a new command auth which allows users to log in with an OAuth token. Users authenticating in this way will have their IRC password set to a random value and sent to them with the command:
```
{
    "command": "irc_password",
    "password": "<password>"
}
```

Related: #440

### DIFF
--- a/server/config.py
+++ b/server/config.py
@@ -50,6 +50,12 @@ class ConfigurationStore:
         self.API_TOKEN_URI = "https://api.test.faforever.com/oauth/token"
         self.API_BASE_URL = "https://api.test.faforever.com/"
         self.USE_API = True
+        # Always set this key. It can be either the public key its self, or a
+        # path pointing to a pub key file.
+        self.API_JWT_PUBLIC_KEY = ""
+        # Resolved public key. If API_JWT_PUBLIC_KEY is a file path then this
+        # will contain the contents of that file.
+        self._API_JWT_PUBLIC_KEY_VALUE = ""
 
         self.MQ_USER = "faf-lobby"
         self.MQ_PASSWORD = "banana"
@@ -153,5 +159,18 @@ def set_log_level():
     logger.setLevel(config.LOG_LEVEL)
 
 
+def read_api_pub_key():
+    pub_key = config.API_JWT_PUBLIC_KEY
+
+    is_key = pub_key.startswith("-----BEGIN") or pub_key.startswith("ssh-rsa")
+
+    if pub_key and not is_key:  # pragma: no cover
+        with open(pub_key) as f:
+            config._API_JWT_PUBLIC_KEY_VALUE = f.read()
+    else:
+        config._API_JWT_PUBLIC_KEY_VALUE = pub_key
+
+
 config = ConfigurationStore()
 config.register_callback("LOG_LEVEL", set_log_level)
+config.register_callback("API_JWT_PUBLIC_KEY", read_api_pub_key)
--- a/server/config.py
+++ b/server/config.py
@@ -50,8 +50,9 @@ class ConfigurationStore:
         self.API_TOKEN_URI = "https://api.test.faforever.com/oauth/token"
         self.API_BASE_URL = "https://api.test.faforever.com/"
         self.USE_API = True
-        # Always set this key. It can be either the public key itself, or a
-        # path pointing to a pub key file.
+        # A path pointing to a pub key file.
+        self.API_JWT_PUBLIC_KEY_FILE = ""
+        # For setting the public key directly. Takes precedence over ..._FILE
         self.API_JWT_PUBLIC_KEY = ""
         # Resolved public key. If API_JWT_PUBLIC_KEY is a file path then this
         # will contain the contents of that file.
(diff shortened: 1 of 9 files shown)


server/lobbyconnection.py
```diff
@@ -530,14 +539,75 @@ async def check_policy_conformity(self, player_id, uid_hash, session, ignore_res
...
+        token = message["token"]
+        unique_id = message['unique_id']
+
+        try:
+            token = jwt.decode(token, pub_key, algorithms="RS256")
+            player_id = token["user_id"]
+        except (InvalidTokenError, KeyError):
+            raise AuthenticationError("Token signature was invalid")
+
+        new_irc_password = b64encode(os.urandom(30)).decode()
```
Brutus5000, line 553: I'm not super familiar with IRC, but it might be possible that a Base64 encoded String contains control commands. Maybe go for sha256 or random UUID instead?

server/lobbyconnection.py
```diff
@@ -666,6 +709,24 @@ async def command_hello(self, message):
...
+        irc_pass = "md5:" + str(m.hexdigest())
+
+        try:
+            await conn.execute(
+                "UPDATE anope.anope_db_NickCore SET pass = %s WHERE display = %s",
+                (irc_pass, login)
+            )
+        except (pymysql.OperationalError, pymysql.ProgrammingError):
+            self._logger.error("Failure updating NickServ password for %s", login)
+
```
Brutus5000, line 729: Reminder to myself: Remove feature from API.

axel1200:
Let's deploy it on test server and test it before merge

server/lobbyconnection.py
Askaholic: Oh because of the `/` character.

server/lobbyconnection.py
```diff
@@ -530,14 +539,75 @@ async def check_policy_conformity(self, player_id, uid_hash, session, ignore_res
...
+    ):
+        conforms_policy = await self.check_policy_conformity(
+            player_id, unique_id, self.session,
+            ignore_result=(
+                steamid is not None or
+                self.player_service.is_uniqueid_exempt(player_id)
+            )
+        )
+        if not conforms_policy:
+            return
```
cleborys, line 603: I see that the logic is staying and I'm not completely sure what the policy check is actually doing (making sure a single steam account has only one associated faf account, except for very old accounts?),
but does failing the policy check count as failed login attempt and if so, does it increase the `metrics.user_logins.labels("failure")` counter?

server/config.py
```diff
@@ -50,6 +49,12 @@ def __init__(self):
         self.API_TOKEN_URI = "https://api.test.faforever.com/oauth/token"
         self.API_BASE_URL = "https://api.test.faforever.com/"
         self.USE_API = True
+        # Always set this key. It can be either the public key its self, or a
+        # path pointing to a pub key file.
+        self.API_JWT_PUBLIC_KEY = ""
```
cleborys, line 54: Might be worth adding it to `tests/data/test_conf.yaml` for documentation.
(That file is not used for regular tests, only those that check configuration refreshes work correctly, so it won't mess with the api_pub_key fixture.)

tests/integration_tests/test_login.py
```diff
@@ -213,3 +216,83 @@ async def test_server_double_login(lobby_server):
...
+    assert msg == {
+        'command': 'welcome',
+        'me': {
+            'clan': '678',
+            'country': '',
+            'global_rating': [2000.0, 125.0],
+            'id': 1,
+            'ladder_rating': [2000.0, 125.0],
+            'login': 'test',
+            'number_of_games': 5
```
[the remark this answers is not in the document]
Askaholic: Well the numbers come from the test data so I think it's fine. I like checking the entire message when possible because it means if anything changes at all you will get a failing test. Like if suddenly the global rating started being sent as `1500, 500` then that probably means some code is not loading from the database correctly anymore.

server/lobbyconnection.py
Askaholic: Yea metrics are a little off, although I think this branch is not taken very often since most of the time we ignore policy responses anyways.

server/config.py
Askaholic: Good idea

server/lobbyconnection.py
```diff
@@ -529,14 +538,75 @@ async def check_policy_conformity(self, player_id, uid_hash, session, ignore_res
...
+
+        async with self._db.acquire() as conn:
+            result = await conn.execute(
+                select([t_login.c.login, t_login.c.steamid])
+                .where(t_login.c.id == player_id)
+            )
+            row = await result.fetchone()
+
+            username = row[t_login.c.login]
+            steamid = row[t_login.c.steamid]
```
norraxx, line 566: access to row property straight like `row.login` or `row.steamid` ;-)

server/lobbyconnection.py
```diff
@@ -665,6 +708,20 @@ async def command_hello(self, message):
 
         await self.send_game_list()
 
+    async def update_irc_password(self, conn, login, password):
+        passwordmd5 = md5(password.encode()).hexdigest()
```
norraxx, line 712: `password_md5`? :innocent:

server/lobbyconnection.py
```diff
@@ -665,6 +708,20 @@ async def command_hello(self, message):
...
+        # Since the password is hashed on the client, what we get at this point
+        # is really md5(md5(sha256(password))). This is entirely insane.
```
norraxx, line 714: :facepalm: nono! :-D

tests/unit_tests/test_configuration_refresh.py
```diff
@@ -93,3 +80,9 @@ async def test_config_no_callback_without_change(config_service, monkeypatch):
 
     assert config.DB_PASSWORD == "banana"
     callback.assert_not_called()
+
+
+async def test_config_api_key_value(config_service):
+    config.refresh()
```
norraxx, line 86: I would test, value before refresh + i think, config should be a fixture.

server/lobbyconnection.py
axel1200: From a security perspective  applying hash algorithms on hashes is bad

Sheikah45:
We need to test this now that I have gotten up the oauth for the client. Additionally this should be changed to using the hydra service that brutus set up. I imagine that with the hydra service we would only need the token based login flow with the server.

Sheikah45:
See https://github.com/FAForever/downlords-faf-client/pull/2175

PR #592, opened 2020-05-29
Cleanup project files

Askaholic:
Seems like pipenv changed something and now refuses to do system installs from the root directory. So just copy pipfile* to the `code` directory as well.

I also removed some old files which I'm pretty sure are not needed.


Brutus5000:
Wasn't the server certificate used for some signature stuff in uid verification?

Askaholic:
As far as I can tell there is nowhere in the code where the lobby server reads that file or even does any crypto besides md5 hashing the IRC password. (In fact we removed the crypto library from the dependencies a while ago because it was giving us security warnings on github). The UID stuff is all handled by the policy server. The lobby server simply forwards data from the client to the policy server and reads the responses.

PR #597, opened 2020-06-02
Remove unneeded rating type conversion

axel1200:


Askaholic:
This requires more changes and is already done in #428.

PR #598, opened 2020-06-02
Travis pull from faf-db directly

### LINKED ISSUE #579: Travis set up database from faf-db directly

Askaholic:
Also bump to db version 93.

Closes #579.

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -11,14 +11,16 @@ dist: bionic
 services:
   - docker
 
+env:
+  global:
+    - FAF_DB_VERSION="v93"
+
 before_install:
   - sudo /etc/init.d/mysql stop
-  - git clone https://github.com/FAForever/faf-stack.git faf-stack
-      && pushd faf-stack
-      && git checkout v20.5.9.3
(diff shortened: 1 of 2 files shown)


.travis.yml
```diff
@@ -11,14 +11,16 @@ dist: bionic
 services:
   - docker
 
+env:
+  global:
+    - FAF_DB_VERSION="v93"
```
norraxx, line 16: Is it tag name? If yes, i would name variable with suffix _tag

Askaholic: I guess. I mean technically it can be anything that can follow `git checkout` so it could be a commit hash or something. Although in practice we will probably always put a tag there

norraxx: I'd name it with tag suffix, cause you mean it to be tag.

PR #604, opened 2020-06-05
Issue/#602 search start confirmation

### LINKED ISSUE #602: Add search start confirmation
Need to add a message when a player starts searching which confirms that a player has entered into a certain queue. For TMM this includes when a player is a regular party member as it is needed for the UI to display which queues the party is in.

Might be an opportunity to clean up the `match_found` and `game_launch_cancelled` messages. I think `match_cancelled` would actually be a better name, and it can also include the associated queue name.

Make sure to update https://github.com/FAForever/downlords-faf-client/issues/1750.

Askaholic:
Add server message to tell clients when they have been entered into a queue.

Rename `game_launch_cancelled` to `match_cancelled`.

Closes #602

### DIFF
--- a/server/ladder_service.py
+++ b/server/ladder_service.py
@@ -338,7 +338,7 @@ class LadderService(Service):
                 if not hosted:
                     raise TimeoutError("Host left lobby")
             finally:
-                # TODO: Once the client supports `game_launch_cancelled`, don't
+                # TODO: Once the client supports `match_cancelled`, don't
                 # send `launch_game` to the client if the host timed out. Until
                 # then, failing to send `launch_game` will cause the client to
                 # think it is searching for ladder, even though the server has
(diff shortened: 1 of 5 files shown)


Askaholic:
> Wouldn't it be better to send the current search status (in which queues a player is) instead of the search starting? How do we indicate when a party left a queue but is still present in others?

Good point, I didn't think about it very hard :P

> I assume a "match_cancelled" (or it's game launch workaround) is also used for stopping the search and sent to every player in the party when the "initiator" cancels the search?

No, that is just for when you found a match, but the game was cancelled because of an error during match setup (player timing out).

Askaholic:
Mmkay. This way was easier because of how the data is organized. Otherwise as @cleborys said, I think I'd probably want to add the queues to the player objects which is a bit more work.

Geosearchef:
Ok, this is fine as well, I can easily handle that client sided.

PR #608, opened 2020-06-20
Feature/#607 match confirmation

### LINKED ISSUE #607: Replace party `ready` check with match confirmation
## Overview
For the purposes of a matchmaker it makes more sense to perform the "are players ready to start" check when a match is found instead of when the party is entered into the queue. I suggest that we add a `match_info` message that look something like this:

```json
{
    "command": "match_info",
    "expires_at": "2020-06-18 20:30:26.677218",
    "players_total": 4,
    "players_ready": 1,
    "ready": false
}
```
Where `expires_at` is a timestamp in ISO format and `ready` refers the the ready state of the player that the message is being sent to (this way the client can synchronize its UI state with the server).

The client can display this information and provide a "Ready up" button. When a player clicks the button, the client will send a `match_ready` command that will look something like this:

```json

{
    "command": "match_ready"
}
```

Once all players have readied up, the server will start the match launching procedure. If some players fail to ready up by the expiry time (say `20` seconds from when the match was found) then the match will be cancelled, the server will send a `match_cancelled` message and return both searches to the matchmaker queue (notifying the client with a `search_info` command).

## Implementation
I suggest writing a small wrapper around a future that will keep track of which players have readied up:

```python3 
class MatchOffer:
    def __init__(self, players, expires_at):
        ...

    def ready_player(self, player):
        ...

    async def wait_ready(self):
        ...

    def to_dict(self):
        """Return contents of `match_info` message."""
```

and adding a weak reference to it on the `Player` objects. Then the player object can also get a method 

```python3
def ready_match(self):
    ...
```

which will check if the player has a match offer and proxy the call to `ready_player`.

Askaholic:
Adds the ability for players to accept matchmaker games before the server tries to launch them. Clients will need to respond with a `match_ready` command which has already been supported by the server with a dummy implementation for years. If some players don't accept the match within the timeout, the match will be cancelled, the players who failed to accept the match will receive a matchmaker violation and their parties will be removed from all queues. Any parties where all players accepted the match will be automatically re-entered into the queues they had been searching for before the match was found.

New server messages:
- `match_info`: sent to create or update the match offer UI. This might also include additional information about the match in the future.
  Example:
  ```json
  {
      "command": "match_info",
      "expires_at": "2020-01-31T14:30:36+00:00",
      "expires_in": 20.0,
      "players_total": 6,
      "players_ready": 2,
      "ready": false
  }
  ```
- `match_notice`: sent to players who are searching together to notify them which other player(s) in their party did not accept the match.
  Example:
  ```json
  {
      "command": "match_notice",
      "unready_players": [1697983, 149808, 1390]
  }
  ```

Required client messages:
- `match_ready`: response to `match_info` with attribute `ready=False` to mark ones self as ready. All players in the match must respond with a `match_ready` message before the game launch sequence will begin.
  Example:
  ```json
  {
      "command": "match_ready"
  }
  ```

Requires client changes as described in https://github.com/FAForever/downlords-faf-client/issues/1783.
Closes #607


tests/unit_tests/test_match_offer.py
```diff
@@ -0,0 +1,73 @@
...
+        await offer.wait_ready()
+
+
+@pytest.mark.asyncio
+async def test_wait_ready(offer):
+    offer.expires_at = datetime.now() + timedelta(seconds=5)
+
+    players = list(offer.get_unready_players())
+
+    for player in players:
```
cleborys, line 70: Does
```python
for player in offer.get_unready_players():
```
not work?

Askaholic: It works. I can't remember why I did it like this.

server/matchmaker/match_offer.py
```diff
@@ -0,0 +1,81 @@
...
+        Mark a player as ready.
+
+        Broadcasts the state change to other players.
+        """
+        if player not in self.player_ready_states:
+            raise KeyError(player)
+
+        self.player_ready_states[player] = True
+
+        self.write_broadcast_update()
```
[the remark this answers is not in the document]
Askaholic: Yea, if that happens I'll just send the message to the current player.

server/ladder_service.py
```diff
@@ -319,21 +328,67 @@ def start_queue_handlers(self):
...
+                    s2.unmatch()
+                    asyncio.create_task(queue.search(s2))
+        except Exception as e:
+            self._logger.exception(
+                "Error processing match between searches %s, and %s: %s",
+                s1, s2, e
+            )
+
+    def create_match_offer(self, players: Iterable[Player]):
+        offer = MatchOffer(players, datetime.now() + timedelta(seconds=20))
```
Askaholic, line 383: Probably should make this a config variable

server/matchmaker/match_offer.py
```diff
@@ -0,0 +1,88 @@
...
+        )
+
+    def ready_player(self, player: Player) -> None:
+        """
+        Mark a player as ready.
+
+        Broadcasts the state change to other players.
+        """
+        if player not in self._players_ready:
+            raise KeyError(player)
```
Askaholic, line 44: Not necessary

server/ladder_service.py
```diff
@@ -365,15 +375,75 @@ def on_match_found(
...
+            )
+            # TODO: Refactor duplication
+            msg = {"command": "match_cancelled"}
+            for player in all_players:
+                if player.state == PlayerState.STARTING_AUTOMATCH:
+                    player.state = PlayerState.IDLE
+                player.write_message(msg)
+
+            # Return any player that accepted the match back to the queue
+            # TODO: make this work with parties
```
Askaholic, line 413: Probably don't need to even add auto-requeue to begin with. It would be nice, but if its too hard, people can figure out how to requeue on their own.

server/ladder_service/ladder_service.py
```diff
@@ -458,17 +482,111 @@ def on_match_found(
...
+                    if player in search_players
+                ]
+                if not search_unready_players:
+                    new_search = Search(
+                        players=search.players,
+                        # TODO: What start time do we want to use?
+                        start_time=search.start_time,
+                        rating_type=search.rating_type,
+                        on_matched=search.on_matched
+                    )
```
Askaholic: Can we just reuse the same object? That way we preserve the failed matching attempts / threshold expansion which would give the auto requeue a real technical benefit over just having people rejoin the queue.

Askaholic: Seems like yes

BlackYps:
This really is ready? I missed that

Askaholic:
> This really is ready? I missed that

Yes. Certainly ready enough to build a client implementation against.

PR #615, opened 2020-06-29
Issue/#610 refactor protocol interface

### LINKED ISSUE #610: Refactor Protocol and QDataStreamProtocol classes
Move necessary interface definitions into `Protocol` as abstract functions and change type signatures to use `Protocol` everywhere. This should make it easier to add a different protocol implementation such as one that uses simple UTF8 encoded JSON. `QDataStreamProtocol` should only be used in a few places (servercontext.py and __init__.py) where the code actually depends on the QDataStream implementation.

The `Protocol` interface should look something like:
```python3
class Protocol():
    def __init__(self, reader, writer):
        """reader and writer should be public attributes"""

    def is_connected(self) -> bool:
        ...

    async def read_message(self):
        ...

    async def send_message(self, message: dict):
        ...

    def write_message(self, message: dict):
        ...

    async def close(self):
        ...

    async def drain(self):
        ...
```

Might also want to expose `send_messages` and `write_messages`, or refactor the one place where it's actually called.

Askaholic:
There were only a few places that were relying on methods from `QDataStreamProtocol`. Now it should be possible to simply plug a different `Protocol` implementation into `ServerContext` and everything should work correctly.

Also added `write_message` to player (extracted from other PRs).

Closes #610

### DIFF
--- a/server/__init__.py
+++ b/server/__init__.py
@@ -7,7 +7,7 @@ Copyright (c) 2015-2016 Michael Søndergaard <sheeo@faforever.com>
 Distributed under GPLv3, see license.txt
 """
 import logging
-from typing import Optional
+from typing import Optional, Tuple, Type
 
 from prometheus_client import start_http_server
 
@@ -18,14 +18,14 @@ from .configuration_service import ConfigurationService
 from .control import run_control_server
 from .game_service import GameService
 from .gameconnection import GameConnection
-from .games.game import GameState, VisibilityState
+from .games import GameState, VisibilityState
 from .geoip_service import GeoIpService
 from .ice_servers.nts import TwilioNTS
 from .ladder_service import LadderService
 from .lobbyconnection import LobbyConnection
 from .message_queue_service import MessageQueueService
 from .player_service import PlayerService
-from .protocol import QDataStreamProtocol
+from .protocol import Protocol, QDataStreamProtocol
 from .rating_service.rating_service import RatingService
 from .servercontext import ServerContext
 from .stats.game_stats_service import GameStatsService
@@ -62,11 +62,8 @@ if config.ENABLE_METRICS:
     start_http_server(config.METRICS_PORT)
 
 
-PING_MSG = QDataStreamProtocol.pack_message('PING')
-
-
 async def run_lobby_server(
-    address: (str, int),
+    address: Tuple[str, int],
     database: FAFDatabase,
     player_service: PlayerService,
     game_service: GameService,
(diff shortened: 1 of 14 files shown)


server/__init__.py
```diff
@@ -154,5 +150,7 @@ def make_connection() -> LobbyConnection:
         )
 
     ctx = ServerContext(make_connection, name="LobbyServer")
+    ping_msg = ctx.protocol_class.encode_message({"command": "ping"})
+
```
cleborys, line 154: Could this be defined closer to the usage in `ping_broadcast`?
(Still in the same scope, just so it's easier to understand what `ping_message` is when reading `ping_broadcast`. I guess that would mean moving the definition of `ping_broadcast`.)

server/lobbyconnection.py
```diff
@@ -190,7 +188,7 @@ async def on_message_received(self, message):
             await self.abort("Error processing command")
 
     async def command_ping(self, msg):
-        await self.protocol.send_raw(PONG_MSG)
+        await self.send({"command": "pong"})
```
cleborys, line 191: How much of an impact does this have? I see that you are still caching `PING` but no longer `PONG` for every protocol (although both are cached inside `QDataStreamProtocol`)?.

server/protocol/protocol.py
```diff
@@ -30,14 +63,68 @@ async def send_messages(self, messages: List[dict]) -> None:
...
+
+    async def drain(self) -> None:
+        """
+        Await the write buffer to empty.
+        See StreamWriter.drain()
+
+        :raises: DisconnectedError if the client disconnects while waiting for
+        the write buffer to empty.
+        """
+        # NOTE: This sleep is needed in python versions <= 3.6
```
cleborys, line 122: Did we not update to 3.7 recently?

server/__init__.py
cleborys: I guess you are caching here so that it's cached for every protocol (not just `QDataStreamProtocol`)?

server/lobbyconnection.py
Askaholic: Yea it will create a temporary dict every time, which is pretty negligible (I benchmarked a similar thing a while ago). It could be refactored but I don’t think it’s worth it. Just a lot nicer to read this way

server/__init__.py
Askaholic: Yea, if you were to create two `ServerContext` objects, each with a different protocol implementation, then the ping message would get encoded twice, once for each protocol implementation.

That's where I'd like to get to eventually. I'd like to be able to spin up a second `ServerContext` on a different port which uses a better protocol. There are still more refactors that need to happen though for the data to be broadcast over all contexts correctly. Right now, any broadcast would only go out over one context.

Issue #589, opened 2020-05-24
Disable auto reporting teamkills


Viba:
Moderation team would like to disable this feature, the generated reports are mostly garbage. If a "proper" TK happens players can still report them normally.

I think related commit here: https://github.com/FAForever/server/commit/beaa8f34176fb7720f6f902a366969c2d51c187c

Askaholic:
Via #616

PR #616, opened 2020-06-29
Issue/#589 disable auto reporting teamkills

MartijnPieterse:
Redid my changes, and added a more sensible git log. Tests are now passing.

I tried to solve issue #589, but i just want some feedback because i have close to no idea what i am doing. I reverted the change, and then kept the external interface, this should be removed only when the client no longer calls it, at least that seemed logical to me.

### DIFF
--- a/server/gameconnection.py
+++ b/server/gameconnection.py
@@ -1,13 +1,17 @@
 import asyncio
 import contextlib
 
-from server.db import FAFDatabase
 from sqlalchemy import or_, select, text
+
+from server.db import FAFDatabase
 
 from .abc.base_game import GameConnectionState
 from .config import TRACE
 from .db.models import (
-    coop_leaderboard, coop_map, login, teamkills
+    coop_leaderboard,
+    coop_map,
+    login,
+    teamkills
 )
 from .decorators import with_logger
 from .game_service import GameService
@@ -15,7 +19,9 @@ from .games.game import Game, GameError, GameState, ValidityState, Victory
 from .player_service import PlayerService
 from .players import Player, PlayerState
 from .protocol import (
-    DisconnectedError, GpgNetServerProtocol, QDataStreamProtocol
+    DisconnectedError,
+    GpgNetServerProtocol,
+    QDataStreamProtocol
 )
 
 
--- a/server/gameconnection.py
+++ b/server/gameconnection.py
@@ -1,19 +1,13 @@
 import asyncio
 import contextlib
 
-from sqlalchemy import or_, select, text
-
 from server.db import FAFDatabase
+from sqlalchemy import or_, select, text
 
 from .abc.base_game import GameConnectionState
 from .config import TRACE
 from .db.models import (
-    coop_leaderboard,
-    coop_map,
-    login,
-    moderation_report,
-    reported_user,
-    teamkills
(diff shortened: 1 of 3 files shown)


server/gameconnection.py
```diff
@@ -347,67 +347,12 @@ async def handle_teamkill_report(self, gametime, reporter_id, reporter_name, tea
...
-
-            result = await conn.execute(insert)
-
-            await conn.execute(
-                reported_user.insert().values(
-                    player_id=verified_teamkiller_id,
-                    report_id=result.lastrowid
-                )
-            )
+        pass
```
Askaholic, line 355: Maybe insert into `teamkills` like we do with `teamkill_happened`? I'm not exactly sure what the difference is. Will have to ask some mods or game dev guys.

Askaholic: @speed2CZ @viba

Brutus5000: The teamkills table just logs all teamkills, whether they are reported or not (collecting evidence for reports without anybody needing to confirm the report)

Askaholic: I know, but do we need to insert here? Or is it good enough that we're inserting in `handle_teamkill_happened`?

MartijnPieterse: I looked in the lua code, and the teamkillhappened is always called. Then optionally it can be reported. Thus, the handle_teamkill_report can safely do nothing.

See https://github.com/FAForever/fa/blob/a3654d6a12b29b67fedece934f142872764d5c97/lua/UserSync.lua#L80

Askaholic: K

tests/unit_tests/test_gameconnection.py
```diff
@@ -394,50 +394,7 @@ async def test_handle_action_TeamkillReport(game: Game, game_connection: GameCon
...
-    await game_connection.handle_action('TeamkillReport', ['270', 0, 'Dostya', 0, 'Geosearchef'])
-
-    async with database.acquire() as conn:
-        result = await conn.execute(
-            "select game_id,id from moderation_report where reporter_id=2 and game_id=%s and game_incident_timecode=270",
-            game.id)
-        report = await result.fetchone()
-        assert report is None
-
+        assert report == None
```
Askaholic, line 397: Equality comparison with `None`, `True`, and `False` should always use `is`.
```suggestion
        assert report is None
```

tests/unit_tests/test_gameconnection.py
```diff
@@ -384,6 +384,7 @@ async def test_json_stats(game_connection: GameConnection, game_stats_service, p
 async def test_handle_action_EnforceRating(game: Game, game_connection: GameConnection):
     await game_connection.handle_action('EnforceRating', [])
     assert game.enforce_rating is True
+## MARPIE
```
Askaholic: hmmmm

MartijnPieterse: Removed already, in the current state it is no longer there, my way of working might not be compatible. I did a reset HEAD~2 to remove all my commits, than made a fresh one. The github also says "Outdated".

Askaholic: Yea I know, I was just looking at the commit history

PR #618, opened 2020-06-30
Issue/#617 Add check for generated map

### LINKED ISSUE #617: Rank games played on generated maps
Currently games played on maps created by the generator are not ranked because the maps are not stored in the map vault. A check needs to be added to see if a game is played on a generated map and then not label that map as a bad map.

Sheikah45:
Adds in a check for if the map is a generated map by checking whether the start of the map name matches the map generator template name. If the map is marked as generated it will not be given a BAD_MAP Validity State.

Closes #617

### DIFF
--- a/server/games/game.py
+++ b/server/games/game.py
@@ -668,11 +668,12 @@ class Game:
                 "WHERE lower(filename) = lower(%s)", (self.map_file_path, )
             )
             row = await result.fetchone()
+            is_generated = self.map_file_path.startswith("neroxis_map_generator")
 
             if row:
                 self.map_id = row['id']
 
-            if (not row or not row['ranked']) and self.validity is ValidityState.VALID:
+            if (not row or not row['ranked']) and self.validity is ValidityState.VALID and not is_generated:
                 await self.mark_invalid(ValidityState.BAD_MAP)
 
             modId = self.game_service.featured_mods[self.game_mode].id
--- a/server/games/game.py
+++ b/server/games/game.py
@@ -668,14 +668,13 @@ class Game:
                 "WHERE lower(filename) = lower(%s)", (self.map_file_path, )
             )
             row = await result.fetchone()
-            is_generated = False
-            if self.map_file_path:
-                is_generated = "neroxis_map_generator" in self.map_file_path
+            is_generated = (self.map_file_path and "neroxis_map_generator" in self.map_file_path)
 
             if row:
                 self.map_id = row['id']
 
-            if (not row or not row['ranked']) and self.validity is ValidityState.VALID and not is_generated:
+            if (self.validity is ValidityState.VALID and
+               ((row and not row.ranked) or (not row and not is_generated))):
                 await self.mark_invalid(ValidityState.BAD_MAP)
 
             modId = self.game_service.featured_mods[self.game_mode].id
--- a/server/games/game.py
+++ b/server/games/game.py
@@ -667,23 +667,26 @@ class Game:
                 "SELECT id, ranked FROM map_version "
                 "WHERE lower(filename) = lower(%s)", (self.map_file_path, )
             )
-            row = await result.fetchone()
-            is_generated = (self.map_file_path and "neroxis_map_generator" in self.map_file_path)
 
-            if row:
-                self.map_id = row['id']
+        row = await result.fetchone()
+        is_generated = (self.map_file_path and "neroxis_map_generator" in self.map_file_path)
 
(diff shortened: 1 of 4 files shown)


server/games/game.py
```diff
@@ -668,11 +668,14 @@ async def update_game_stats(self):
...
             row = await result.fetchone()
+            is_generated = False
+            if self.map_file_path:
+                is_generated = "neroxis_map_generator" in self.map_file_path
 
             if row:
                 self.map_id = row['id']
 
-            if (not row or not row['ranked']) and self.validity is ValidityState.VALID:
+            if (not row or not row['ranked']) and self.validity is ValidityState.VALID and not is_generated:
```
Askaholic, line 678: Lets make sure that it only checks `is_generated` if the map does not exist in the database. Otherwise it would be possible for the `ranked` column to be overridden by putting "neroxis_map_generator" in the name. Something like this:
```suggestion
            if self.validity is ValidityState.VALID and (row and not row.ranked or not row and not is_generated):
```

Will want to format it so the line is not too long.

server/games/game.py
```diff
@@ -668,11 +668,14 @@ async def update_game_stats(self):
                 "WHERE lower(filename) = lower(%s)", (self.map_file_path, )
             )
             row = await result.fetchone()
+            is_generated = False
+            if self.map_file_path:
+                is_generated = "neroxis_map_generator" in self.map_file_path
```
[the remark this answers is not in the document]
Sheikah45: Changed

server/games/game.py
Sheikah45: Changed

tests/unit_tests/test_game.py
```diff
@@ -150,6 +150,35 @@ async def test_ffa_not_rated(game, game_add_players):
...
+
+
+async def test_unranked_generated_map_not_rated(database, game, game_add_players):
+    async with database.acquire() as conn:
+        await conn.execute(
+            "INSERT into map_version (id, max_players, "
+            "width, height, version, filename, ranked, hidden, map_id)"
+            "VALUES (0, 4, 512, 512, 1, \"maps/neroxis_map_generator_sneaky_map.zip\","
+            "0, 0, 0)"
+        )
```
Askaholic, line 171: Don't insert in the test, add it to `tests/data/test-data.sql`.

Sheikah45: Will do

server/games/game.py
```diff
@@ -669,19 +669,24 @@ async def update_game_stats(self):
...
-                self.map_id = row['id']
+        is_generated = (self.map_file_path and "neroxis_map_generator" in self.map_file_path)
 
-            if (not row or not row['ranked']) and self.validity is ValidityState.VALID:
-                await self.mark_invalid(ValidityState.BAD_MAP)
+        if row:
+            self.map_id = row['id']
 
-            modId = self.game_service.featured_mods[self.game_mode].id
+        if (self.validity is ValidityState.VALID and
```
cleborys, line 677: I like binary operators at the start of the new line rather than the end of the last (and I think we are doing that somewhat consistently?), like
```python
if (
  self.validity is ValidityState.VALID
  and ((row and not row.ranked) or (not row and not is_generated))
):
  await  self.mark_invalid(ValidityState.BAD_MAP)
```
but that's of course only a minor thing.

tests/data/test-data.sql
```diff
@@ -198,6 +199,9 @@ insert into map_version (id, description, max_players, width, height, version, f
   (16, 'SCMP 015', 8, 512, 512, 2, 'maps/scmp_015.v0002.zip', 0, 15),
   (17, 'SCMP 015', 8, 512, 512, 3, 'maps/scmp_015.v0003.zip', 0, 15);
 
+insert into map_version (id, description, max_players, width, height, version, filename, hidden, ranked, map_id) values
+  (18, 'Sneaky_Map', 8, 512, 512, 1, "maps/neroxis_map_generator_sneaky_map.zip", 0, 0, 16);
+
```
cleborys, line 204: Might be more straightforward to add that map to the previous insert statement (and set `ranked` of all maps that are there already).

server/games/game.py
Sheikah45: Consistent style is important noted

tests/data/test-data.sql
Sheikah45: Done

Askaholic:
Closes #617

PR #619, opened 2020-07-01
Add compatibility to db v96

Brutus5000:


tests/unit_tests/test_rating_service.py
```diff
@@ -205,7 +205,7 @@ async def test_get_new_player_rating_created(semiinitialized_service):
     tables, a new rating entry should be created.
     """
     service = semiinitialized_service
-    player_id = 999
+    player_id = 300
```
cleborys, line 208: Ah, I guess this test was only working before since foreign keys weren't properly enforced?

Askaholic: Yea, this test and the other one too

PR #622, opened 2020-07-04
Skip policy server check when running in dev mode

Askaholic:
When running through `pipenv run devserver`, you will always want to skip this check. I found myself commenting these lines out every time I needed to test something locally, and realized it should have a config option.


server/lobbyconnection.py
```diff
@@ -551,15 +551,16 @@ async def command_hello(self, message):
                 )
             )
 
-            conforms_policy = await self.check_policy_conformity(
-                player_id, message['unique_id'], self.session,
-                ignore_result=(
-                    steamid is not None or
-                    self.player_service.is_uniqueid_exempt(player_id)
+            if config.USE_POLICY_SERVER:
+                conforms_policy = await self.check_policy_conformity(
```
cleborys, line 555: Could also do the check inside `check_policy_conformity` as
```python
if not config.USE_POLICY_SERVER:
    return True
```
to save one indentation level.
Which one is easier to read is up to personal taste I think :D

Askaholic: True

PR #623, opened 2020-07-04
Issue/#581 Remove old ladder pool

### LINKED ISSUE #581: Remove old ladder map pool
After we've migrated to new tables we should delete the code that pulls from the legacy map pool.

RemcoSchrijver:
Removed legacy code using the old ladder pool and removed part of the test that checks for the legacy maps.

Closes #581

### DIFF
--- a/server/ladder_service.py
+++ b/server/ladder_service.py
@@ -77,19 +77,6 @@ class LadderService(Service):
 
     async def update_data(self) -> None:
         async with self._db.acquire() as conn:
-            # Legacy ladder1v1 map pool
-            # TODO: Remove this https://github.com/FAForever/server/issues/581
-            result = await conn.execute(
-                "SELECT ladder_map.idmap, "
-                "table_map.name, "
-                "table_map.filename "
-                "FROM ladder_map "
-                "INNER JOIN table_map ON table_map.id = ladder_map.idmap"
-            )
-            maps = [Map(*row.as_tuple()) async for row in result]
-
-            self.ladder_1v1_map_pool.set_maps(maps)
-
             map_pool_maps = await self.fetch_map_pools(conn)
             db_queues = await self.fetch_matchmaker_queues(conn)
 
--- a/server/ladder_service.py
+++ b/server/ladder_service.py
@@ -51,29 +51,21 @@ class LadderService(Service):
         self._db = database
         self._informed_players: Set[Player] = set()
         self.game_service = game_service
-
-        # Fallback legacy map pool and matchmaker queue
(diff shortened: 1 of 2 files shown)


server/ladder_service.py
```diff
@@ -77,19 +77,6 @@ async def initialize(self) -> None:
...
-            result = await conn.execute(
-                "SELECT ladder_map.idmap, "
-                "table_map.name, "
-                "table_map.filename "
-                "FROM ladder_map "
-                "INNER JOIN table_map ON table_map.id = ladder_map.idmap"
-            )
-            maps = [Map(*row.as_tuple()) async for row in result]
-
-            self.ladder_1v1_map_pool.set_maps(maps)
```
Askaholic, line 91: We can remove `ladder_1v1_map_pool` completely from `LadderService`. Which means we can also remove the hardcoded `ladder1v1` queue and the code that initializes it (line 73).

RemcoSchrijver: Ohh yes I see, removed that as well.

server/ladder_service.py
```diff
@@ -52,15 +52,13 @@ def __init__(
...
-        self.ladder_1v1_map_pool = MapPool(0, "ladder1v1")
+        # Matchmaker queue
         self.queues = {
             'ladder1v1': MatchmakerQueue(
                 game_service,
                 name="ladder1v1",
                 featured_mod=FeaturedModType.LADDER_1V1,
-                rating_type=RatingType.LADDER_1V1,
-                map_pools=[(self.ladder_1v1_map_pool, None, None)]
+                rating_type=RatingType.LADDER_1V1
```
Askaholic, line 61: Like I said, I think we should remove this whole hardcoded queue. There is no way to use the new map pool format without inserting this queue into the database, so there's no reason to keep it around.

That will also mean that the `TODO` in `initialize` can be resolved, since there wont be any more hardcoded queues to initialize.

RemcoSchrijver: Removing both of those generates a slew of crashes in multiple unit and integration tests but if those just expect the existing information from the hardcoded queue I'll change those.

Askaholic: Which tests are failing? Why? Are they tests that are specifically checking for the hardcoded queue? Then they can be deleted too.

Askaholic:
Any progress on this?

RemcoSchrijver:
Am a bit strapped for time at the moment, will look at it next week.

Askaholic:
Superseded by #656

PR #625, opened 2020-07-09
Issue/#282 Game titles seem to be quoted needlessly

### LINKED ISSUE #282: Game titles seem to be quoted needlessly
The lobby connection uses `cgi.escape` to escape game titles. From what I can tell the server does not need this encoding for anything, as it doesn't use the "title" field other than receive it and send it. The client always receives a game as a "game_info" command - apart from passing it around and sending it back, it prints it in a couple of places where it uses HTML escaping (tooltips, game widgets), and places where it doesn't (live replay tabs). This causes escape sequences to print in a couple places.

Escaping the game title seems needless, and there are places where we would need to unescape it in the client. We should fix the client and start sending the title unquoted.

MartijnPieterse:
Hej, i created a bunch of games using some silly tool i made and made sure they were correct in the client. All seems ok. I could not find any places where the &amp; i mean \&amp; would pop up though. Nice to see that html injection works in this box.

Checked the client (a little bit) for any unescape code, but java isn't really my cup of tea.

Fixes #282

### DIFF
--- a/server/lobbyconnection.py
+++ b/server/lobbyconnection.py
@@ -1,7 +1,6 @@
 import asyncio
 import contextlib
 import hashlib
-import html
 import json
 import random
 import urllib.parse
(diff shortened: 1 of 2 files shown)


MartijnPieterse:
No i didn't, will try to do some more test. Ideally these tests belong to the client code ofcourse, we don't trust the client, and the client shouldn't trust us. :-)

PR #630, opened 2020-07-14
More readme updates

Askaholic:
Since there have been a number of new contributors recently, I have realized it would be a lot easier for people to start contributing if the readme was better. I added some more details about running tests and added a `CONTRIBUTING.md` file.

### DIFF
--- a/CONTRIBUTING.md
+++ b/CONTRIBUTING.md
@@ -1,0 +1,73 @@
+# Contributing
+
+When contributing for the first time, try to find an issue labeled with
+[good first issue](https://github.com/FAForever/server/issues?q=is%3Aissue+is%3Aopen+label%3A%22good+first+issue%22),
+and leave a comment stating that you would like to work on
+it, that way we can assign it to you. Fork this repository and base your pull
+request on the `develop` branch. Note that we squash on merge, so pull requests
+that have been open for a while may need to be rebased before they can be merged.
+
+## Pull Request Process
+
+1.  Create a branch on your fork based on `develop`. Our convention is to name
+branches like `issue/#123-fix-thing`, or `feature/#456-add-thing` depending on
+whether the referenced issue pertains to a bug or an enhancement.
+2.  Before committing, make sure that the unit tests pass, and that you have
+checked your code with a linter.
+    1. Check changes with [flake8](https://pypi.org/project/flake8/).
+    2. Sort imports with [isort](https://pypi.org/project/isort/), our settings
+    are already checked into the setup.cfg file.
+    3. Run all tests with `pipenv run tests`.
+3.  Open a pull request with the issue number in the title, and `Closes #123` in
+the description. GitHub will default the title to the name of your branch, but
+feel free to expand it a little if you have the space, e.g. `Issue/#123 Fix this
+one thing` or `Feature/#456 add that other thing`.
+
+### Committing
+
+Use the normal git conventions for commit messages, with the following rules:
+-   Subject line shorter than 80 characters
+-   Proper capitalized sentence as subject line, with no trailing period
+-   For non-trivial commits, include a commit message body, describing the
+change in detail
+
+For further tips on writing good commit messages have a read through
+[this post](https://chris.beams.io/posts/git-commit/#seven-rules).
+
+## Code Style
+
+Historically, this project has lacked an established coding style, so the
+formatting may be inconsistent in places. If you are unsure of how to format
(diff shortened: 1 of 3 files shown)


CONTRIBUTING.md
```diff
@@ -0,0 +1,72 @@
+# Contributing
+
+When contributing for the first time, try to find an issue labeled with
+"good first issue", and leave a comment stating that you would like to work on
```
Brutus5000, line 4: Why not directly link it? https://github.com/FAForever/server/issues?q=is%3Aissue+is%3Aopen+label%3A%22good+first+issue%22

README.md
```diff
@@ -4,59 +4,75 @@
 [![Coveralls Status](https://img.shields.io/coveralls/FAForever/server/develop.svg)](https://coveralls.io/github/FAForever/server)
 [![semver](https://img.shields.io/badge/license-GPLv3-blue)](license.txt)
 
-This is the source code for the [Forged Alliance Forever](http://www.faforever.com/) server.
+This is the source code for the
+[Forged Alliance Forever](https://www.faforever.com/) server.
```
Brutus5000, line 8: lobby server, not server (I guess many people get confused by this)

CONTRIBUTING.md
Askaholic: Didn't think about it. Good idea

README.md
Askaholic: Yea, eventually I might add a section about what the code actually does. Hosting games, matchmaking, player status, etc.

README.md
```diff
@@ -4,59 +4,75 @@
...
- - For non-trivial commits, always include a commit message body, describing the change in detail
- - If there are related issues, reference them in the commit message footer
+Before opening a pull request, please take a moment to look over the
+[contributing guidelines](CONTRIBUTING.md).
 
 ## Setting up for development
 
-First make sure you have an instance of [faf-db](https://github.com/FAForever/db) setup. Then install the dependencies to a virtual environment
-using pipenv by running:
+First, follow the instructions on the [faf-db repo](https://github.com/FAForever/db)
```
cleborys, line 17: Sometimes (but rarely) I've had to use a db version different from the latest and the only way we document which version we are assuming is by what is checked out in the `.travis.yml`, I believe. Not sure if that's worth a mention here.

README.md
```diff
@@ -4,59 +4,75 @@
...
+running:
 
     $ pipenv sync --dev
 
 You can then start the server in development mode with:
 
     $ pipenv run devserver
 
-You will see some exception trace when the server fails to connect to RabbitMQ,
+You will see an exception trace when the server fails to connect to RabbitMQ,
```
cleborys, line 28: Another error to ignore (during tests mostly?) is 
```
ERROR    LobbyConnection:lobbyconnection.py:580 Failure updating NickServ password for test
```
I think I wasted 30min on that at the start.

cleborys:
@BlackYps pointed out that if you are using PyCharm and want to run the tests from within PyCharm, you need to add `--mysql_database=faf` in the "Additional Arguments" field in the "Run Configuration". I think it already understands the `pipenv` setup out-of-the-box.
Maybe that's also a good addition to the readme if any new people are used to using PyCharm.

README.md
Askaholic: I thought about that too, but I think it's too much of an edge case to mention right here. If anything it would go in a section with more details about database migrations, but I'm not even sure a section like that is appropriate for this readme. Maybe it would belong to the readme on the db repo.

CONTRIBUTING.md
```diff
@@ -0,0 +1,73 @@
...
+that have been open for a while may need to be rebased before they can be merged.
+
+## Pull Request Process
+
+1.  Create a branch on your fork based on `develop`. Our convention is to name
+branches like `issue/#123-fix-thing`, or `feature/#456-add-thing` depending on
+whether the referenced issue pertains to a bug or an enhancement.
+2.  Before committing, make sure that the unit tests pass, and that you have
+checked your code with a linter.
+    1. Check changes with [flake8](https://pypi.org/project/flake8/).
```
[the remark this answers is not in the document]
Askaholic: I feel like people can figure this out for themselves if they need to. Everyone will have their own setup and their own workflow. I just want to mention the tools we've agreed to use.

README.md
Askaholic: Added a list of errors that can be ignored

PR #631, opened 2020-07-15
Fix problems with no repetition code

### LINKED ISSUE #624: Problems with No Repetition Code
It seems that the no repetition code that has people not play one of their last 3 maps is no longer functioning with the introduction of the new pools. I'm noticing people playing the same map repeatedly. 

I don't think it has anything to do with people transitioning between pools and the code considering the same map in two different pools to be two different maps as the example below consists of two people facing one another for two games in a row on the same map twice in a row. They should have had the same map pool for both games. They were both also 1400 (rounded) in both games, meaning they were not on the edge of switching between two pools.

I also don't think it has anything to do with the code bugging out due to there being no viable map to pick, as the smallest map pool is 7 which means in the worst case scenario, there is still 1 viable map for the code to pick with the 3 map restriction. 

repeat

devopsdeluxe:
Potentially fixes #624 

I.E. I see no reason why @Askaholic's proposed [fix](https://github.com/FAForever/server/issues/624#issuecomment-655192848) _wouldn't_ work.

### DIFF
--- a/tests/data/test-data.sql
+++ b/tests/data/test-data.sql
@@ -228,7 +228,11 @@ insert into game_stats (id, startTime, gameName, gameType, gameMod, host, mapId,
   (41941, NOW() + interval 6 minute, 'MapRepetition', '0', 6, 1, 6, 0),
   (41942, NOW(), 'OldRatingNull',  '0', 6, 1, NULL, 0),
   (41943, NOW(), 'OldRatingLose', '0', 6, 1, NULL, 0),
-  (41944, NOW(), 'OldRatingWin', '0', 6, 1, NULL, 0);
+  (41944, NOW(), 'OldRatingWin', '0', 6, 1, NULL, 0),
+  (41945, NOW() + interval 7 minute, 'MapRepetition', '0', 6, 2, 6, 0),
+  (41946, NOW() + interval 8 minute, 'MapRepetition', '0', 6, 2, 5, 0),
+  (41947, NOW() + interval 9 minute, 'MapRepetition', '0', 6, 2, 4, 0),
+  (41948, NOW() + interval 10 minute, 'MapRepetition', '0', 6, 2, 3, 0);
 
 insert into game_player_stats (gameId, playerId, AI, faction, color, team, place, mean, deviation, scoreTime) values
   (1, 1, 0, 0, 0, 2, 0, 1500, 500, NOW()),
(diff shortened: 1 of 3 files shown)


Askaholic:
We’ll need to add at least one unit test to catch this in the future.

devopsdeluxe:
> We’ll need to add at least one unit test to catch this in the future.

@Askaholic let me know if I'm on the right track here...

As I understand, this would be accomplished by adding a few additional rows into the DB testing [seed](https://github.com/FAForever/server/blob/develop/tests/data/test-data.sql) so the resulting table would look something like:

```sql
 MySQL [faf]> SELECT * FROM game_stats;
+-------+---------------------+---------+--------------+----------+---------+------+-------+---------------+----------+
| id    | startTime           | endTime | replay_ticks | gameType | gameMod | host | mapId | gameName      | validity |
+-------+---------------------+---------+--------------+----------+---------+------+-------+---------------+----------+
|     1 | 2020-07-18 20:18:18 | NULL    |         NULL | 0        |       6 |    1 |     1 | Test game     |        0 |
| 41935 | 2020-07-18 20:18:18 | NULL    |         NULL | 0        |       6 |    1 |  NULL | MapRepetition |        0 |
| 41936 | 2020-07-18 20:19:18 | NULL    |         NULL | 0        |       6 |    1 |     1 | MapRepetition |        0 |
| 41937 | 2020-07-18 20:20:18 | NULL    |         NULL | 0        |       6 |    1 |     2 | MapRepetition |        0 |
| 41938 | 2020-07-18 20:21:18 | NULL    |         NULL | 0        |       6 |    1 |     3 | MapRepetition |        0 |
| 41939 | 2020-07-18 20:22:18 | NULL    |         NULL | 0        |       6 |    1 |     4 | MapRepetition |        0 |
| 41940 | 2020-07-18 20:23:18 | NULL    |         NULL | 0        |       6 |    1 |     5 | MapRepetition |        0 |
| 41941 | 2020-07-18 20:24:18 | NULL    |         NULL | 0        |       6 |    1 |     6 | MapRepetition |        0 |
| 41942 | 2020-07-18 20:18:18 | NULL    |         NULL | 0        |       6 |    1 |  NULL | OldRatingNull |        0 |
| 41943 | 2020-07-18 20:18:18 | NULL    |         NULL | 0        |       6 |    1 |  NULL | OldRatingLose |        0 |
| 41944 | 2020-07-18 20:18:18 | NULL    |         NULL | 0        |       6 |    1 |  NULL | OldRatingWin  |        0 |
| 41945 | 2020-07-18 20:24:18 | NULL    |         NULL | 0        |       6 |    2 |   NULL | MapRepetition |        0 |
| 41946 | 2020-07-18 20:24:18 | NULL    |         NULL | 0        |       6 |    2 |     6 | MapRepetition |        0 |
| 41947 | 2020-07-18 20:24:18 | NULL    |         NULL | 0        |       6 |    2 |     5 | MapRepetition |        0 |
| 41948 | 2020-07-18 20:24:18 | NULL    |         NULL | 0        |       6 |    2 |     4 | MapRepetition |        0 |
| 41949 | 2020-07-18 20:24:18 | NULL    |         NULL | 0        |       6 |    2 |     3 | MapRepetition |        0 |
+-------+---------------------+---------+--------------+----------+---------+------+-------+---------------+----------+
```

Which would allow for a second player's map history to be pulled, and the intersection two players map history could be found similar to how [`test_get_ladder_history_many_maps`](https://github.com/FAForever/server/blob/develop/tests/unit_tests/test_ladder_service.py#L462) works (albeit with 2 players instead of 1 :wink: ).

Askaholic:
Yea, inserting test data is the way to go. You'll also need to insert entries into `game_player_stats` for the second player though, the host of the game is not important.

devopsdeluxe:
@Askaholic ready for review!

tests/data/test-data.sql
```diff
@@ -238,7 +242,11 @@ insert into game_player_stats (gameId, playerId, AI, faction, color, team, place
   (41938, 1, 0, 0, 0, 2, 0, 1500, 500, NOW() + interval 3 minute),
   (41939, 1, 0, 0, 0, 2, 0, 1500, 500, NOW() + interval 4 minute),
   (41940, 1, 0, 0, 0, 2, 0, 1500, 500, NOW() + interval 5 minute),
-  (41941, 1, 0, 0, 0, 2, 0, 1500, 500, NOW() + interval 6 minute);
+  (41941, 1, 0, 0, 0, 2, 0, 1500, 500, NOW() + interval 6 minute),
+  (41945, 2, 0, 0, 0, 1, 0, 1500, 500, NOW() + interval 1 minute),
```
cleborys, line 246: The `scoreTime` you enter is *before* the `startTime` of the game.
Probably won't ever be a problem but maybe it's nice to keep them in sync as before (i.e. increment them all by six).

devopsdeluxe: Updated those lines for consistency

PR #635, opened 2020-07-28
Issue/#605 add game type to game info message

MartijnPieterse:
I think all is done now. Correct me if i'm wrong.

### DIFF
--- a/server/games/game.py
+++ b/server/games/game.py
@@ -28,6 +28,7 @@ from .typedefs import (
     EndedGameInfo,
     FeaturedModType,
     GameState,
+    GameType,
     ValidityState,
     Victory,
     VisibilityState
@@ -46,6 +47,7 @@ class Game:
     The initialization mode to use for the Game.
     """
     init_mode = InitMode.NORMAL_LOBBY
+    game_type = GameType.CUSTOM
 
     def __init__(
         self,
@@ -807,6 +809,7 @@ class Game:
             "uid": self.id,
             "title": self.name,
             "state": client_state,
+            "game_type": GameType.to_string(self.game_type),
             "featured_mod": self.game_mode,
             "sim_mods": self.mods,
             "mapname": self.map_folder_name,
--- a/server/games/typedefs.py
+++ b/server/games/typedefs.py
@@ -20,6 +20,31 @@ class Victory(Enum):
     ERADICATION = 2
     SANDBOX = 3
 
+@unique
+class GameType(Enum):
+    COOP = 0
+    CUSTOM = 1
+    MATCHMAKER = 2
+
+    @staticmethod
+    def from_string(value: str) -> Optional["GameType"]:
+        """
+        :param value: The string to convert from
+
+        :return: GameType or None if the string is not valid
+        """
+        return {
+            "coop": GameType.COOP,
+            "custom": GameType.CUSTOM,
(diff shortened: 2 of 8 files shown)


server/games/game.py
```diff
@@ -800,13 +802,19 @@ def to_dict(self):
             GameState.ENDED: 'closed',
             GameState.INITIALIZING: 'closed',
         }.get(self.state, 'closed')
+        game_type = {
+            GameType.COOP: 'coop',
+            GameType.CUSTOM: 'custom',
+            GameType.MATCHMAKER: 'matchmaker',
+        }.get(self.game_type, 'custom')
```
Askaholic, line 809: Just make the enum values be strings and do `self.game_type.value`

server/games/typedefs.py
```diff
@@ -20,6 +20,12 @@ class Victory(Enum):
     ERADICATION = 2
     SANDBOX = 3
 
+@unique
+class GameType(Enum):
+    NONE = 0
```
Askaholic, line 25: Do we need `NONE`?

server/games/game.py
Askaholic: Or maybe make this a method of `GameType`, we might need the integer values for hooking it up to the database later.

tests/integration_tests/test_server.py
```diff
@@ -192,6 +192,30 @@ async def test_host_missing_fields(event_loop, lobby_server, player_service):
     assert msg['map_file_path'] == 'maps/scmp_007.zip'
     assert msg['featured_mod'] == 'faf'
 
+@fast_forward(5)
+async def test_host_game_type(event_loop, lobby_server, player_service):
```
Askaholic, line 196: Might be able to get away with just using the existing tests and adding the `game_type` field to the assertions. I think there is already one for hosting a game.

server/games/typedefs.py
MartijnPieterse: No, in my first iteration i added None because i did not know when this would be used, so i'll remove this.

tests/integration_tests/test_server.py
MartijnPieterse: Ok, i'll take a closer look at the testcases. There should be some already which i can use.

server/games/game.py
MartijnPieterse: I copied the method of gameState, i'll make it more nice and neat.

Askaholic:
Another annoyance is that the client already has a mapping for `game_type`, probably left over from some old code, which actually raises an error when it tries to deserialize the game_info message. Will make an issue for it:
https://github.com/FAForever/downlords-faf-client/blob/9107f842d1529997a242aa5a984db7fecdc048d9/src/main/java/com/faforever/client/remote/domain/GameInfoMessage.java#L27-L28

Askaholic:
I made a PR to remove the conflict on the client side: https://github.com/FAForever/downlords-faf-client/pull/1851

Brutus5000:
@Askaholic Did you test if this field breaks the java client? Depending on whether Jackson or Gson is used by default an exception is thrown on encountering an unexpected field. Then this would become a breaking change.

Askaholic:
Yes I did, I tested it with the PR where I removed the old `gameType` field.

PR #638, opened 2020-08-02
Issue/#637 mq service initialization race condition

### LINKED ISSUE #637: Race condition in MQService initialization
Looks like there is a pesky concurrency issue in the message service:

```bash
  File "/usr/local/lib/python3.7/asyncio/runners.py", line 43, in run
    return loop.run_until_complete(main)
  File "/usr/local/lib/python3.7/asyncio/base_events.py", line 587, in run_until_complete
    return future.result()
  File "server.py", line 73, in main
    service.initialize() for service in services.values()
  File "/code/server/game_service.py", line 63, in initialize
    await self._message_queue_service.declare_exchange("faf-rabbitmq")
  File "/code/server/message_queue_service.py", line 80, in declare_exchange
    new_exchange = await self._channel.declare_exchange(
AttributeError: 'NoneType' object has no attribute 'declare_exchange'

During handling of the above exception, another exception occurred:

  File "/usr/local/lib/python3.7/site-packages/aiormq/connection.py", line 374, in __reader
    weight, channel, frame = await self.__receive_frame()
  File "/usr/local/lib/python3.7/site-packages/aiormq/connection.py", line 326, in __receive_frame
    frame_header = await self.reader.readexactly(1)
  File "/usr/local/lib/python3.7/asyncio/streams.py", line 679, in readexactly
    await self._wait_for_data('readexactly')
  File "/usr/local/lib/python3.7/asyncio/streams.py", line 473, in _wait_for_data
    await self._waiter
concurrent.futures._base.CancelledError
```

1. Game service initialize and message queue service initialize are both started and awaited with `asyncio.gather`
2. message queue service successfully connects and yields on `_connection.channel`
3. Game service calls `declare_exchange`

Now the message queue service is in a state where `_connection` is initialized, but `_channel` is still None

cleborys:
Adds `_is_ready` to `MessageQueueService` keeping track of whether a connection is fully established (i.e. connection *and* channel). Methods that touch ths are `initialize()`, `shutdown()`, and `reconnect()`.
Since spamming e.g. reconnect could potentially leave things in a bad state, I also added an `_initialization_lock` and `_is_ready` is only modified by methods holding the lock.

Closes #637

### DIFF
--- a/integration_tests/test_message_queue/test_message_queue_service.py
+++ b/integration_tests/test_message_queue/test_message_queue_service.py
@@ -100,6 +100,7 @@ async def test_reconnect(mq_service):
     assert mq_service._exchange_types["test_topic"] == aio_pika.ExchangeType.TOPIC
     assert "test_direct" in mq_service._exchanges
     assert mq_service._exchange_types["test_direct"] == aio_pika.ExchangeType.DIRECT
+    assert mq_service._is_ready
 
 
 async def test_incorrect_credentials(mocker, caplog):
@@ -109,6 +110,7 @@ async def test_incorrect_credentials(mocker, caplog):
     await service.initialize()
     expected_warning = "Unable to connect to RabbitMQ. Incorrect credentials?"
     assert expected_warning in [rec.message for rec in caplog.records]
+    assert service._is_ready is False
     caplog.clear()
 
     await service.declare_exchange("test_exchange")
--- a/integration_tests/test_message_queue/test_message_queue_service.py
+++ b/integration_tests/test_message_queue/test_message_queue_service.py
@@ -144,3 +144,16 @@ async def test_incorrect_vhost(mocker, caplog):
     await service.initialize()
 
     assert any("Incorrect vhost?" in rec.message for rec in caplog.records)
+
+
+async def test_initialize_declare_exchange_race_condition():
+    service = MessageQueueService()
+
+    async def keep_declaring_exchanges():
+        i = 0
+        while service._channel is None:
+            await service.declare_exchange(f"exchange_{i}")
+            await asyncio.sleep(0)
+            i += 1
+
+    await asyncio.gather(keep_declaring_exchanges(), service.initialize())
--- a/integration_tests/test_message_queue/test_message_queue_service.py
+++ b/integration_tests/test_message_queue/test_message_queue_service.py
@@ -159,3 +159,17 @@ async def test_initialize_declare_exchange_race_condition():
             i += 1
 
     await asyncio.gather(keep_declaring_exchanges(), service.initialize())
+
+
+async def test_declaring_exchange_without_initialization():
+    service = MessageQueueService()
+    exchange_name = "test_exchange"
+
+    assert service._is_ready is False
+    assert service._connection is None
+
+    await service.declare_exchange(exchange_name)
+
+    assert service._is_ready
+    assert service._connection is not None
+    assert service._exchanges.get(exchange_name) is not None
--- a/server/message_queue_service.py
+++ b/server/message_queue_service.py
@@ -23,6 +23,7 @@ class MessageQueueService(Service):
         self._channel = None
         self._exchanges = {}
         self._exchange_types = {}
+        self._is_ready = False
 
         config.register_callback("MQ_USER", self.reconnect)
         config.register_callback("MQ_PASSWORD", self.reconnect)
@@ -34,9 +35,10 @@ class MessageQueueService(Service):
         if self._connection is not None:
             return
 
-        await self._connect()
+        self._is_ready =  await self._connect()
 
-    async def _connect(self) -> None:
+    async def _connect(self) -> bool:
+        """ Returns True on success. """
         try:
             self._connection = await aio_pika.connect_robust(
                 "amqp://{user}:{password}@{server}:{port}/{vhost}".format(
@@ -50,28 +52,29 @@ class MessageQueueService(Service):
             )
         except ConnectionError:
             self._logger.warning("Unable to connect to RabbitMQ. Is it running?", exc_info=True)
-            return
+            return False
         except ProbableAuthenticationError:
             self._logger.warning(
                 "Unable to connect to RabbitMQ. Incorrect credentials?",
                 exc_info=True
             )
-            return
+            return False
         except Exception as e:
             self._logger.warning(
                 "Unable to connect to RabbitMQ due to unhandled excpetion %s. Incorrect vhost?",
                 e,
                 exc_info=True
             )
-            return
+            return False
 
         self._channel = await self._connection.channel(publisher_confirms=False)
         self._logger.debug("Connected to RabbitMQ %r", self._connection)
+        return True
 
     async def declare_exchange(
         self, exchange_name: str, exchange_type: ExchangeType = ExchangeType.TOPIC
     ) -> None:
-        if self._connection is None:
+        if not self._is_ready:
             self._logger.warning(
                 "Not connected to RabbitMQ, unable to declare exchange."
             )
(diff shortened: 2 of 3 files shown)


server/message_queue_service.py
```diff
@@ -23,6 +23,8 @@ def __init__(self) -> None:
         self._channel = None
         self._exchanges = {}
         self._exchange_types = {}
+        self._is_ready = False
+        self._initialization_lock = asyncio.Lock()
```
Askaholic, line 27: Btw I wrote a decorator for this here https://github.com/FAForever/server/pull/629/commits/cda403335f72b7ca428ef6cb553c8bbc1c963193. Maybe I should extract it to a separate PR?

It lets you do:
```python3
class SomeClass:

    @synchronizedmethod
    async def some_func(self, args):
        ... 
```
to wrap a method with one lock per instance, or 
 
```python3
@synchronized
async def some_func(args):
    ...
```
to lock a function with a global lock (works for methods too, but you will only be able to call the method on one instance at a time).

server/message_queue_service.py
```diff
@@ -31,12 +33,14 @@ def __init__(self) -> None:
...
 
     async def initialize(self) -> None:
-        if self._connection is not None:
-            return
+        async with self._initialization_lock:
+            if self._connection is not None:
+                return
 
-        await self._connect()
+            self._is_ready = await self._connect()
```
Askaholic, line 40: why not make `_connect` set the attribute directly?

server/message_queue_service.py
Askaholic: Oh I guess you are using the lock in more than one function so maybe it's not useful here.

cleborys: I guess one could add something like `@synchronized_group(lock_name)` managing a global dictionary of locks that can be shared between functions/methods, if you want.

server/message_queue_service.py
cleborys: Because I also want to call `_connect` from `reconnect` and there I don't want to set the flag until after the previous exchanges have been re-declared.

Askaholic: Hmm, maybe it would be better to propagate exceptions then. I feel like passing around a "
success" flag as return value is pretty clunky. Gives me nightmares about C.

server/message_queue_service.py
Askaholic: I worked on it a bit more, and I came up with that same solution, except I just made it an optional parameter for the existing decorator. So it would be:

```python3
class Foo:
    def __init__(self):
        self._the_lock = asyncio.Lock()

    @synchronizedmethod("_the_lock")
    async def bar(self):
        ...

    @synchronizedmethod(lock="_the_lock")
    async def baz(self):
        ...

```

server/message_queue_service.py
cleborys: Sure thing.

server/message_queue_service.py
```diff
@@ -11,6 +11,10 @@
 from .decorators import with_logger
 
 
+class ConnectionAttemptFailed(Exception):
```
Askaholic, line 14: Maybe inherit from ConnectionError?

```suggestion
class ConnectionAttemptFailed(ConnectionError):
```

server/message_queue_service.py
```diff
@@ -31,12 +37,18 @@ def __init__(self) -> None:
...
 
-        await self._connect()
+            try:
+                await self._connect()
+            except ConnectionAttemptFailed:
+                return
+            self._is_ready = True
 
     async def _connect(self) -> None:
+        """ Returns True on success. """
```
Askaholic, line 51: Nope
```suggestion
```

server/message_queue_service.py
```diff
@@ -49,29 +61,30 @@ async def _connect(self) -> None:
...
+            )
+            raise ConnectionAttemptFailed
         except ProbableAuthenticationError:
             self._logger.warning(
-                "Unable to connect to RabbitMQ. Incorrect credentials?",
-                exc_info=True
+                "Unable to connect to RabbitMQ. Incorrect credentials?", exc_info=True
             )
-            return
+            raise ConnectionAttemptFailed
```
Askaholic, line 72: Maybe use the `raise from` syntax.

```suggestion
            raise ConnectionAttemptFailed from e
```
(need to change the `except` block to name the exception)

server/message_queue_service.py
```diff
@@ -31,12 +37,18 @@ def __init__(self) -> None:
         config.register_callback("MQ_PORT", self.reconnect)
 
     async def initialize(self) -> None:
-        if self._connection is not None:
-            return
+        async with self._initialization_lock:
+            if self._connection is not None:
```
Askaholic, line 41: Should we be checking `is_ready` here instead?

cleborys: Yes that seems better. TBH one could probably also replace the flag by `self._connection is not None` checks everywhere else, but having the flag feels more readable.

Askaholic:
Do we need to update something in GameService too to wait for the MQService to be ready before calling `declare_exchange`?

cleborys:
Hm... Instead I chose to have `declare_exchange` call the initialization if necessary and added a test to make sure there is only one initialization, even if you call it several times.
(The public `declare_exchange` has a check whether initialization is necessary, the private `_declare_exchange` does not, since you want to be able to re-declare all old exchanges when reconnecting, before you mark the service as ready again.)

Coincidentally, I assume the `asyncio.Lock` is not like a `threading.RLock`, so calling a `@synchronizedmethod("the_lock")` from another `@synchronizedmethod("the_lock")` will deadlock?

Askaholic:
Yes I expect it would

integration_tests/test_message_queue/test_message_queue_service.py
```diff
@@ -144,3 +146,30 @@ async def test_incorrect_vhost(mocker, caplog):
...
+        i = 0
+        while service._channel is None:
+            await service.declare_exchange(f"exchange_{i}")
+            await asyncio.sleep(0)
+            i += 1
+
+    await asyncio.gather(keep_declaring_exchanges(), service.initialize())
+
+
+async def test_declaring_exchange_without_initialization():
```
Askaholic, line 164: What was the deal with these tests again? Are we not able to get them to run in travis?

Askaholic (approved):
Looks fine!

integration_tests/test_message_queue/test_message_queue_service.py
cleborys: We can, but they rely on running the faf-rabbitmq container, which I believe we decided not to want to depend on.
It would be quick to add them back to travis, for example the league service is doing that:
https://github.com/cleborys/faf-league-service/blob/9bec96d5f1b0f9bdf69baff014489fc7b3482a9f/.travis.yml#L27
https://github.com/cleborys/faf-league-service/blob/develop/ci/init-rabbitmq.sh

Askaholic: Hmm, yea it might be nice to do that in a separate PR. I can't remember why we decided not to before. I remember we put that stuff in the integration tests because we didn't want people to need to install rabbitmq just to develop the server, but I would think on Travis it would be fine.

cleborys: Ah, I see. I didn't see that distinction between contributors running it and travis running it, that's why I moved it to the integration tests. Moving it to travis would also mean the docker repository won't evict `faf-rabbitmq` because of six months of inactivity :stuck_out_tongue_closed_eyes: 

By the way, "installing rabbitmq" is literally only running the script linked above which pulls the docker image from the repository and adds the appropriate user, so the only drawback I'd see for contributors is that the image takes up some space.

Askaholic: Hmm, yea that's not a big deal.

PR #641, opened 2020-08-03
Asyncio lock helpers

Askaholic:
Wrote some reusable functionality to add locks to functions and methods. We can use these in a few upcoming PRs:

1. `ServerInstance._start_services` to prevent double initialization
2. `MessageQueueService` to prevent double initialization
3. `MatchmakerQueue.find_matches` to allow multiqueuing by preventing double matching

and probably more in the future.


server/asyncio_extensions.py
```diff
@@ -0,0 +1,152 @@
...
+    lock_name: Optional[str] = None
+) -> AsyncFunc:
+    """Wrap an async method with an async lock stored on the instance."""
+    if lock_name is None:
+        lock_name = f"#{function.__name__}_lock"
+
+    @wraps(function)
+    async def wrapped(obj, *args, **kwargs):
+        lock = obj.__dict__.get(lock_name, None)
+        if not lock:
```
cleborys, line 145: I prefer `if lock is not None:` (although there is probably no difference in practice).

tests/unit_tests/test_asyncio_extensions.py
```diff
@@ -0,0 +1,185 @@
...
+            assert self.in_call is False, "Multiple concurrent executions!"
+
+            self.in_call = True
+            await asyncio.sleep(1)
+            self.in_call = False
+
+    a = Test()
+    b = Test()
+    # Calls to different instances should not block eachother
+    await asyncio.wait(
```
cleborys, line 130: Why did you choose `wait` over `gather` as before?

tests/unit_tests/test_asyncio_extensions.py
```diff
@@ -0,0 +1,185 @@
...
+        [a.sleep_for_1s() for _ in range(500)] +
+        [b.sleep_for_1s() for _ in range(500)],
+        timeout=502
+    )
+
+
+@fast_forward(500)
+async def test_synchronizedmethod_class_invocation():
```
cleborys, line 165: Is this not identical to `test_synchronizedmethod` since the interpreter parses `a.sleep_for_1s()` as `Test.sleep_for_1s(a)` anyways? Or is there a way to make one of these tests pass but not the other?

server/asyncio_extensions.py
```diff
@@ -0,0 +1,152 @@
...
+        self.func = func
+        self.args = args
+
+    def __call__(self, *args):
+        return self.func(*args, *self.args)
+
+
+@overload
+def synchronized() -> AsyncDecorator:
+    ...
```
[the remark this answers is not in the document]
Askaholic: Similar to `None` it is a singleton object. See the `Ellipsis` section https://docs.python.org/3/reference/datamodel.html#the-standard-type-hierarchy. It's frequently used in typing stubs and numpy array indexing https://numpy.org/doc/stable/reference/arrays.indexing.html

server/asyncio_extensions.py
```diff
@@ -0,0 +1,152 @@
...
+def synchronized(function: AsyncFunc) -> AsyncFunc:
+    ...
+
+
+@overload
+def synchronized(lock: Optional[asyncio.Lock]) -> AsyncDecorator:
+    ...
+
+
+def synchronized(*args):
```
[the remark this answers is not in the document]
Askaholic: Eh, it's not too bad. I based it off of the `pytest.fixture` implementation which does a similar thing.

tests/unit_tests/test_asyncio_extensions.py
Askaholic: Because `wait` has a `timeout` parameter.

tests/unit_tests/test_asyncio_extensions.py
Askaholic: Riiight. This can probably be deleted. This was necessary when I was still manually implementing a descriptor (object with `__get__` method) because I needed to verify that my descriptor implementation would correctly for both invocations. But then I realized I didn't need to use a custom descriptor at all, and went back to just using normal functions.

Still, I learned a lot about how python objects actually work under the hood. In fact the calls `a.foo()` and `A.foo(a)` don't do *exactly* the same thing. For instance if you fire up a python interpreter you can easily check that `A.foo is A.foo` is true, but `a.foo is a.foo` is not. I found this documentation quite helpful https://docs.python.org/3/howto/descriptor.html

cleborys: Thanks for the link!

PR #642, opened 2020-08-05
Rework ladder search internals/interface

Askaholic:
The interface for searching/cancelling is now asymmetric. 
- Searches are started on a list of players and one queue name
- Searches are cancelled for one player and one queue name (or all queues)

This should allow pretty trivial integration with the party system. 
- When a player starts a search, call `start_search` with the players entire party
- When a player cancels a search, call `cancel_search` with just the requesting player

Permissions still need to be handled by the caller of `LadderService` (which is always `LobbyConnection`)

### DIFF
--- a/server/matchmaker/matchmaker_queue.py
+++ b/server/matchmaker/matchmaker_queue.py
@@ -52,7 +52,7 @@ class MatchmakerQueue:
         self.team_size = team_size
         self.map_pools = {info[0].id: info for info in map_pools}
 
-        self.queue: Dict[Search, Search] = OrderedDict()
+        self._queue: Dict[Search, None] = OrderedDict()
         self._matches: Deque[Match] = deque()
         self._is_running = True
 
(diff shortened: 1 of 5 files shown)


server/matchmaker/matchmaker_queue.py
```diff
@@ -139,13 +139,13 @@ async def search(self, search: Search) -> None:
...
-            if search in self.queue:
-                del self.queue[search]
+            if search in self._queue:
+                del self._queue[search]
 
     async def find_matches(self) -> None:
         self._logger.info("Searching for matches: %s", self.name)
 
-        if len(self.queue) < 2 * self.team_size:
+        if len(self._queue) < 2 * self.team_size:
```
MartijnPieterse, line 148: This assumes it will always be 2 teams. Should it be flexible to allow for 2v2v2 or 3v3v3v3?

Askaholic:
As there haven't been any complaints, I'm going to merge this

PR #644, opened 2020-08-25
Issue/#603 implement tmm multiqueuing

### LINKED ISSUE #603: TMM: Multiqueue support
Add the ability for a player to join multiple matchmaker queues at the same time.

1. Make sure that when a player finds a match, all other searches that the player is a part of are cancelled.
2. Make sure that only one queue can run the matchmaking algorithm at a time. This might "just work" without changing anything, or it might need an mutex because of asyncio.

Askaholic:
Some more relatively minor ladder refactors that should enable multiqueuing to function correctly. In order to give the LadderService a chance to cancel players extra searches when a match is found, the MatchmakerQueue object now triggers a callback from within the globally synchronized function `find_matches`. As the old way of asynchronously yielding matches was not used anywhere else, I decided to remove it entirely in favor of the callback API.

Also players will now get the "welcome to the matchmaker" message for all queues, not just ladder1v1. I suppose this could be annoying at first, but it will be good for letting veteran players know that they should not expect TMM games to be balanced off of their existing rating.

Closes #603


tests/integration_tests/test_teammatchmaker.py
```diff
@@ -105,6 +111,55 @@ async def test_game_matchmaking(lobby_server):
...
+    }
+    msgs = await asyncio.gather(*[client_response(proto) for proto in protos])
+
+    uid = set(msg["uid"] for msg in msgs)
+    assert len(uid) == 1
+    for msg in msgs:
+        assert msg["init_mode"] == 1
+        assert "None" not in msg["name"]
+        assert msg["mod"] == "faf"
+        assert msg["expected_players"] == 4
```
cleborys, line 157: So you expect the server to give priority to the tmm2v2 queue, rather than e.g. making two 1v1 matches? Does that maybe make this test a bit brittle in case something in the future changes in which order the queue matches are handled?

server/ladder_service.py
```diff
@@ -259,10 +257,10 @@ def _cancel_search(self, initiator: Player, queue_name: str) -> None:
             "%s stopped searching for %s", cancelled_search, queue_name
         )
 
-    def inform_player(self, player: Player):
+    def inform_player(self, player: Player, rating_type: str) -> None:
```
cleborys, line 260: TBH I think this could have a more descriptive name. Maybe `notify_about_initial_learning_phase` (which is the shortest I could think of in one minute).

cleborys: `send_rating_progress_msg`?

server/ladder_service.py
```diff
@@ -289,34 +287,35 @@ def inform_player(self, player: Player):
...
-                asyncio.create_task(
-                    self.start_game(s1.players, s2.players, queue)
-                )
-            except Exception as e:
-                self._logger.exception(
-                    "Error processing match between searches %s, and %s: %s",
-                    s1, s2, e
-                )
+            for player in s1.players + s2.players:
+                player.write_message(msg)
```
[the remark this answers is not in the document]
Askaholic: If the player has disconnected it's a noop.

tests/integration_tests/test_teammatchmaker.py
Askaholic: No, only a single player has queued for ladder, so it's not possible for ladder1v1 to generate a match.

server/ladder_service.py
Askaholic: IMO this function shouldn't exist at all :D, it should be done in the client, but yea that name sounds better

tests/integration_tests/test_teammatchmaker.py
cleborys: Ah, of course.

server/ladder_service.py
```diff
@@ -289,34 +287,35 @@ def inform_player(self, player: Player):
                     )
                 })
 
-    def start_queue_handlers(self):
-        for queue in self.queues.values():
-            asyncio.ensure_future(self.handle_queue_matches(queue))
+    def on_match_found(
+        self,
+        s1: Search,
+        s2: Search,
```
norraxx, line 293: for me, s1 and s2, etc is very bad naming. Don't name variables by type, name them by meaning

server/ladder_service.py
```diff
@@ -289,34 +287,35 @@ def inform_player(self, player: Player):
...
-                ])
-                asyncio.create_task(
-                    self.start_game(s1.players, s2.players, queue)
-                )
-            except Exception as e:
-                self._logger.exception(
-                    "Error processing match between searches %s, and %s: %s",
-                    s1, s2, e
-                )
+            for player in s1.players + s2.players:
```
norraxx, line 299: operator `+`, is there magic inside? I'm offering you to not to override magic methods like `def __add__`.

server/matchmaker/matchmaker_queue.py
```diff
@@ -187,10 +191,19 @@ def match(self, s1: Search, s2: Search) -> bool:
...
         :param s2:
-        :return:
+        :return: True if matching succeeded or False if matching failed
         """
-        if (s1.is_matched or s2.is_matched) or (s1.is_cancelled or s2.is_cancelled):
+        if s1.is_matched or s2.is_matched:
             return False
+        if s1.is_cancelled or s2.is_cancelled:
+            return False
+        # Additional failsafe. Ideally this check will never fail.
```
norraxx, line 200: any logging for return false?

server/ladder_service.py
Askaholic: There is no other meaning. It's search 1 and search 2. Just like player 1 and player 2 except it's not just one player, it's a search.

server/ladder_service.py
Askaholic: No. it's list + list.

Askaholic:
>I guess it is very important that on_match_found and other callbacks passed to it are synchronous/fast, since they'll block the Matchmaker Timer, but I don't know if there's a good way to document that.
Well it's a synchronous function already. I also added a note in the docstring.

> Maybe a test that mocks start_game as an asyncio.sleep(10) and then you could test that the matchmaker still handles 10 players in less than a second.
Since `on_match_found` is a synchronous function this is basically checking that `loop.run_until_complete` is not being called? Because there's no way for `start_game` to block the callback anyways as `start_game` is asynchronous.

> Maybe a new (non-integration) tests that makes sure players are cancelled from other queues when multiqueueing is also a good idea? 
I think I can add something to `test_ladder_service` but I'm not sure what you mean by asserting that everyone's callback was triggered at most once. The callback function is going to get called once per match, so if you add a bunch of searches you would expect it to be called multiple times.

cleborys:
> Since on_match_found is a synchronous function this is basically checking that loop.run_until_complete is not being called? 
Or that they are not being awaited. But yes, it's probably an unnecessary test to add - I just tried to think of ways in which things could go wrong. :stuck_out_tongue_closed_eyes: 

> but I'm not sure what you mean by asserting that everyone's callback was triggered at most once. 
I think what I meant was a test that would check that if you queue up search A for 1v1 and 2v2, search B for 1v1, and search C for 2v2 then if search A is matched with B in the 1v1 queue on the first pop it will correctly be cancelled in the 2v2 queue and no longer match with C on the next pop.

Askaholic:
I see. What I did was have A search for 1v1 and then A, B, C, D search for 2v2 and assert that A's 1v1 search was cancelled. I could queue a second player into the 1v1 queue, but the problem is I can't make any guarantees as to which players will get a game then as it depends on what queue happens to pop first. Either A, B get a 1v1 game or A, B, C, D get a 2v2 game. It makes the assertions kinda awkward. I'll try it and see how it goes.

server/matchmaker/matchmaker_queue.py
```diff
@@ -144,7 +127,15 @@ async def search(self, search: Search) -> None:
             if search in self._queue:
                 del self._queue[search]
 
+    @synchronized(SpinLock(sleep_duration=1))
     async def find_matches(self) -> None:
+        """
+        Perform the matchmaking algorithm.
+
+        Note that this funciton is synchronized such that only one instance of
```
cleborys, line 135: funciton :star_struck:

cleborys (approved):
It's a bit hard to guess at all intentions for long PRs like this one, but this definitely looks good to go to me, even more so since it's been playtested.

server/asyncio_extensions.py
```diff
@@ -32,6 +46,50 @@ async def gather_without_exceptions(
     return results
 
 
+# Based on python3.8 asyncio.Lock
+# https://github.com/python/cpython/blob/6c6c256df3636ff6f6136820afaefa5a10a3ac33/Lib/asyncio/locks.py#L106
+class SpinLock(_ContextManagerMixin):
```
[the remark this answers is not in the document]
Askaholic: Yea it just calls `acquire` and `release` https://github.com/python/cpython/blob/3.9/Lib/asyncio/locks.py. The other option would be to inherit from `Lock` and override the functions, but that's messy because they share very little implementation.

server/matchmaker/matchmaker_queue.py
Askaholic: Ha! Typing too fast again

Issue #640, opened 2020-08-03
Replace ban text with localizable message


Askaholic:
The server should not be sending any strings that get rendered in the client, because these strings are in English and we can't be sure that the user on the other end has their client set to English (in fact a majority of people probably knows English only as a second language, or does not know it very well at all). In fact issue #444 already covers this, but I'm making another ticket just for this message because changing it would allow us to drop the `humanize` dependency.

Here is a link to the message:
https://github.com/FAForever/server/blob/e4ca993a9b7ea8123c64f6917bf9edd05f69b0ce/server/lobbyconnection.py#L1094

It should be replaced with something like:
```
{
    "command": "banned",
    "expires_at": <ISO timestamp>
}
```

See #183 for inspiration, but note that we should be sending ISO timestamps (in UTC) and not timedeltas.

cb0806151:
Hey @Askaholic, could I try fixing this issue or do you already have someone lined up to work on it?

Askaholic:
Yea go for it, note that it will also require a PR to https://github.com/FAForever/downlords-faf-client

cb0806151:
@Askaholic Are you and the team using Linux when developing this application?

Askaholic:
Yea I’ve only run the server on Linux, but I think @Sheikah45 was able to get it working on Windows. If you’re on the FAF discord you can also PM me there.

Sheikah45:
I can confirm I was able to get it working had to use docker desktop and WSL

cb0806151:
@Askaholic I wasn't originally aware that this was going to need to be handled with linux, and WSL is being uncooperative with my machine. I can set up a linux OS on a spare laptop I have to fix this, but that would take quite a while; because of that, I think it'd be best if I recuse myself from fixing this issue. I should have done further research into it before volunteering and apologize for the time of yours that I have wasted

Askaholic:
No worries, just let me know if you want to work on it in the future!

therajdeepbiswas:
Is this still open? I could take this up.

Askaholic:
@therajdeepbiswas Yea for sure! I should edit the issue because I think we should make it a bit simpler than I originally described. I think the right approach would be to just add an i18n key to the message in addition to the English text.

```json
{
    "command": "notice",
    "style": "error"
    "text": "You are banned from FAF for 1 month"
    "i18n_key": "login.error.banned",
    "i18n_args": ["<ISO timestamp"]
}
```

[...]

PR #645, opened 2020-08-25
Issue/#639 Show ban duration to user with higher resolution

### LINKED ISSUE #639: Show ban duration to user with higher resolution
Currently users might get ban duration shown as "1 day" when the duration is e.g. 44 hours. It would be better for the users if they'd see at least the days + hours.

I believe this could be done simply by changing the naturaldelta [here ](https://github.com/FAForever/server/blob/ecb4fd7e570a4ceaeb2fe1c34b3302f80b5b565f/server/lobbyconnection.py#L1092)to more precise and cut it to min hours or minutes or something?

46bit:
Hi. This is my first PR to FAF.

This PR addresses #639: "Show ban duration to user with higher resolution." I think it would be better to implement #640 in the long term, but I don't know how to do that yet and this PR still gives an improvement.

The tests outline the new behaviour, and test more of what was already there. A temporarily banned user will now see a message saying:

```
You are banned from FAF for 1 day and 8.5 hours.
Reason :
[ban reason here]
```

I haven't actually been banned myself 😇 but this seemed like a good first contribution.

Closes #639.

### DIFF
--- a/server/lobbyconnection.py
+++ b/server/lobbyconnection.py
@@ -391,7 +391,7 @@ class LobbyConnection:
             self._logger.debug('Rejected login from banned user: %s, %s, %s',
                                player_id, username, self.session)
 
-            await self.send_ban_message_and_abort(ban_expiry - now, ban_reason)
+            raise_ban_error(ban_expiry - now, ban_reason)
 
         # New accounts are prevented from playing if they didn't link to steam
 
@@ -1079,15 +1079,20 @@ class LobbyConnection:
                 return
 
             ban_expiry = data[ban.c.expires_at]
+            if now >= ban_expiry:
+                return
 
-            if now < ban_expiry:
-                self._logger.debug('Aborting connection of banned user: %s, %s, %s',
-                                   self.player.id, self.player.login, self.session)
-                self.send_ban_message_and_abort(ban_expiry - now, data[ban.c.reason])
+            self._logger.debug("Aborting connection of banned user: %s, %s, %s",
+                               self.player.id, self.player.login, self.session)
+            raise_ban_error(ban_expiry - now, data[ban.c.reason])
 
-    def send_ban_message_and_abort(self, ban_time, reason):
-        ban_time_text = (f"for {humanize.naturaldelta(ban_time)}"
-                         if ban_time.days < 365 * 100 else "forever")
-        raise ClientError((f"You are banned from FAF {ban_time_text}.\n "
-                           f"Reason :\n "
-                           f"{reason}"), recoverable=False)
+def raise_ban_error(ban_duration, ban_reason):
+    raise ClientError((f"You are banned from FAF {ban_duration_text(ban_duration)}.\n "
+                       f"Reason :\n "
+                       f"{ban_reason}"), recoverable=False)
+
+def ban_duration_text(ban_duration):
+    if ban_duration.days > 365 * 100:
+        return "forever"
+    humanized_ban_duration = humanize.precisedelta(ban_duration, minimum_unit='hours')
+    return f"for {humanized_ban_duration}"
--- a/server/lobbyconnection.py
+++ b/server/lobbyconnection.py
@@ -1087,8 +1087,8 @@ class LobbyConnection:
             raise_ban_error(ban_expiry - now, data[ban.c.reason])
 
 def raise_ban_error(ban_duration, ban_reason):
-    raise ClientError((f"You are banned from FAF {ban_duration_text(ban_duration)}.\n "
-                       f"Reason :\n "
+    raise ClientError((f"You are banned from FAF {ban_duration_text(ban_duration)}. <br>"
+                       f"Reason : <br>"
                        f"{ban_reason}"), recoverable=False)
 
 def ban_duration_text(ban_duration):
--- a/server/lobbyconnection.py
+++ b/server/lobbyconnection.py
@@ -56,6 +56,26 @@ class ClientError(Exception):
         self.recoverable = recoverable
 
 
+class BanError(Exception):
+    def __init__(self, ban_expiry, ban_reason, *args, **kwargs):
+        super().__init__(*args, **kwargs)
+        self.ban_expiry = ban_expiry
+        self.ban_reason = ban_reason
+
+    def message(self):
+        return (f"You are banned from FAF {self._ban_duration_text()}. <br>"
+                f"Reason : <br>"
+                f"{self.ban_reason}")
+
+    def _ban_duration_text(self):
+        ban_duration = self.ban_expiry - datetime.now()
+        if ban_duration.days > 365 * 100:
+            return "forever"
+        humanized_ban_duration = humanize.precisedelta(ban_duration, minimum_unit="hours")
+        return f"for {humanized_ban_duration}"
+
+
+
 class AuthenticationError(Exception):
     def __init__(self, message, *args, **kwargs):
         super().__init__(*args, **kwargs)
@@ -166,6 +186,13 @@ class LobbyConnection:
                 'command': 'authentication_failed',
                 'text': ex.message
             })
+        except BanError as ex:
+            await self.send({
+                "command": "notice",
+                "style": "error",
+                "text": ex.message()
+            })
+            await self.abort(ex.message())
         except ClientError as ex:
             self._logger.warning("Client error: %s", ex.message)
             await self.send({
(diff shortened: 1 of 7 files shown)


Askaholic:
Awesome! I totally agree that this is a worthwile improvement, and also that it should be replaced in the future :P. I'll take a closer look at it tonight.

46bit:
@Askaholic Brilliant. I had quite a few rounds of fixing CI 🙈 but this is ready for review now.

tests/unit_tests/test_lobbyconnection.py
```diff
@@ -949,3 +950,29 @@ async def test_check_policy_conformity_fatal(lobbyconnection, policy_server):
...
+    await lobbyconnection.abort_connection_if_banned()
+
+    lobbyconnection.player.id = 201 # test user whose ban has been revoked
+    await lobbyconnection.abort_connection_if_banned()
+
+    lobbyconnection.player.id = 202 # test user whose ban has expired
+    await lobbyconnection.abort_connection_if_banned()
+
+    lobbyconnection.player.id = 203 # test user who is permabanned
+    with pytest.raises(ClientError) as banned_error:
```
Askaholic, line 969: Hah, I didn't even know pytest yields the error back to you

server/lobbyconnection.py
```diff
@@ -391,7 +391,7 @@ async def check_user_login(self, conn, username, password):
             self._logger.debug('Rejected login from banned user: %s, %s, %s',
                                player_id, username, self.session)
 
-            await self.send_ban_message_and_abort(ban_expiry - now, ban_reason)
+            await send_ban_message_and_abort(ban_expiry - now, ban_reason)
```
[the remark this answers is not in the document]
46bit: Done

server/lobbyconnection.py
```diff
@@ -1079,15 +1079,20 @@ async def abort_connection_if_banned(self):
...
             ban_expiry = data[ban.c.expires_at]
+            if now >= ban_expiry:
+                return
+
+            self._logger.debug('Aborting connection of banned user: %s, %s, %s',
+                               self.player.id, self.player.login, self.session)
+            raise_ban_error(ban_expiry - now, data[ban.c.reason])
+
+def raise_ban_error(ban_duration, ban_reason):
+    raise ClientError((f"You are banned from FAF {ban_duration_text(ban_duration)}.\n "
```
Askaholic, line 1090: Sorry, I thought I commented on this earlier, but I must have forgotten to click the "Add" button or something. The client actually expects these messages to be HTML
```suggestion
    raise ClientError((f"You are banned from FAF {ban_duration_text(ban_duration)}.</br> "
```
right now it looks like this (minus the white background which has something to do with linux):
image
The first one happens to line up correctly, but you can see there is no newline after the reason.

tests/unit_tests/test_lobbyconnection.py
```diff
@@ -949,3 +950,29 @@ async def test_check_policy_conformity_fatal(lobbyconnection, policy_server):
...
+        await lobbyconnection.abort_connection_if_banned()
+    assert banned_error.value.message == 'You are banned from FAF forever.\n Reason :\n Test permanent ban'
+    assert banned_error.value.recoverable is False
+
+    lobbyconnection.player.id = 204 # test user who is banned for another 46 hours
+    with pytest.raises(ClientError) as banned_error:
+        await lobbyconnection.abort_connection_if_banned()
+    assert re.match('You are banned from FAF for 1 day and 2[0-3]\.[0-9]+ hours.\n Reason :\n Test ongoing ban with 46 hours left', banned_error.value.message)
```
Askaholic, line 977: I think you want a raw string
```suggestion
    assert re.match(r'You are banned from FAF for 1 day and 2[0-3]\.[0-9]+ hours\.\n Reason :\n Test ongoing ban with 46 hours left', banned_error.value.message)
```

Askaholic: Oh, also use double quotes for strings please

Askaholic (commented):
I was getting a warning when running the tests:
```
tests/unit_tests/test_lobbyconnection.py:977: DeprecationWarning: invalid escape sequence \.
    assert re.match('You are banned from FAF for 1 day and 2[0-3]\.[0-9]+ hours.\n Reason :\n Test ongoing ban with 46 hours left', banned_error.value.message)
```

server/lobbyconnection.py
Askaholic: Oh I keep clicking the wrong button I guess. I keep clicking "Add single comment" and not the green button which adds it to my review...

46bit: Hopefully this is fixed now :)

tests/unit_tests/test_lobbyconnection.py
46bit: Thanks--I haven't used Python in a long time so a lot of this is new to me.

46bit:
@Askaholic I think I've addressed everything.

I also refactored a new `BanError` type and started outputting the ban reason and expiry time for future use. Unfortunately the test suite is failing with some sort of problem parsing the protocol :/

server/lobbyconnection.py
```diff
@@ -56,6 +56,31 @@ def __init__(self, message, recoverable=True, *args, **kwargs):
         self.recoverable = recoverable
 
 
+class BanError(Exception):
+    """
+    Represents a BanError
```
Askaholic, line 61: I don't think this comment adds any value :P

server/lobbyconnection.py
```diff
@@ -56,6 +56,31 @@ def __init__(self, message, recoverable=True, *args, **kwargs):
...
+    def message(self):
+        return (f"You are banned from FAF {self._ban_duration_text()}. <br>"
+                f"Reason : <br>"
+                f"{self.ban_reason}")
+
+    def _ban_duration_text(self):
+        ban_duration = self.ban_expiry - datetime.now()
+        if ban_duration.days > 365 * 100:
+            return "forever"
+        humanized_ban_duration = humanize.precisedelta(ban_duration, minimum_unit='hours')
```
Askaholic, line 79: Right now the code is still pretty inconsistent on quote usage, but I've been trying to convert everything over to double quotes (unless the string contains a double quote).
```suggestion
        humanized_ban_duration = humanize.precisedelta(ban_duration, minimum_unit="hours")
```

server/lobbyconnection.py
```diff
@@ -166,6 +191,15 @@ async def on_message_received(self, message):
                 'command': 'authentication_failed',
                 'text': ex.message
             })
+        except BanError as ex:
+            await self.send({
+                'command': 'notice',
+                'style': 'error',
+                'text': ex.message(),
+                'ban_reason': ex.ban_reason,
+                'ban_expiry': ex.ban_expiry.isoformat()
```
Askaholic, line 200: It's a bit weird attaching these fields to the `notice` message, and the way that the client is written does not lend its self well to having dynamic messages (every message gets deserialized into a class, so the class needs to have members defined for all expected fields). So IMO, it's probably better to hold off on this until we add a dedicated ban message, otherwise if someone tries to implement it, it will turn into a mess.

tests/data/test-data.sql
```diff
@@ -323,7 +324,8 @@ insert into ban (player_id, author_id, reason, level, expires_at, revoke_reason,
 insert into ban (player_id, author_id, reason, level, expires_at, revoke_time) values
   (201, 201, 'Test revoked ban', 'GLOBAL', NULL, now() - interval 1 day),
   (202, 202, 'Test expired ban', 'GLOBAL', now() - interval 1 day, NULL),
-  (203, 203, 'Test permanent ban', 'GLOBAL', now() + interval 1000 year, NULL)
+  (203, 203, 'Test permanent ban', 'GLOBAL', make_timestamptz(3020, 01, 01, 12, 0, 0), NULL),
```
Askaholic, line 327: `make_timestampz` doesn't exist I guess. You could just use `TIMESTAMP` instead

Askaholic (commented):
Yea the refactor is great! The travis build is failing because of the `make_timestampz` function you added to the test data file.

server/lobbyconnection.py
46bit: I really think this should be done later as a big PR on the whole codebase rather than trying to get it right here.

46bit:
I've removed exporting the fields since that wasn't helpful. Hopefully good to go now.

server/lobbyconnection.py
46bit: (Am volunteering.)

Askaholic: The tricky thing about doing it all at once is that it could cause conflicts with other PR's. If you want to take a stab at that, go ahead, maybe it won't be too bad. But that still does't justify committing new code with single quotes ;)

Askaholic: It's even coding style bullet number 1 as listed in the contributing document: [CONTRIBUTING.md](https://github.com/FAForever/server/blob/develop/CONTRIBUTING.md)

46bit: cool, pushed again. opened a PR for mass-replacing too to see if it's workable: #648.

Askaholic (approved):
Perfect!

PR #646, opened 2020-08-26
[WIP] Issue/#609 Add metrics for certain data structure sizes

### LINKED ISSUE #609: Add metrics for certain data structure sizes
There are some dictionaries that hold strong references to their values which we should post some metrics about to make sure they aren't growing without bound. Mostly I want to know "how many elements are in the dictionary".

https://github.com/FAForever/server/blob/ccf349e3b4bbaee95373d79a0e2f4ee9c5d064a2/server/ladder_service.py#L53

https://github.com/FAForever/server/blob/ccf349e3b4bbaee95373d79a0e2f4ee9c5d064a2/server/player_service.py#L25

https://github.com/FAForever/server/blob/ccf349e3b4bbaee95373d79a0e2f4ee9c5d064a2/server/game_service.py#L46

https://github.com/FAForever/server/blob/ccf349e3b4bbaee95373d79a0e2f4ee9c5d064a2/server/matchmaker/matchmaker_queue.py#L49

46bit:
This PR closes #609. It adds Prometheus metrics to track the number of items in various datastructures. This will be useful to check that they are not growing unboundedly.

The metrics are updated re-using whatever periodic update functions already existed, rather than adding new schedules and functions to do it.

### DIFF
--- a/server/ladder_service.py
+++ b/server/ladder_service.py
@@ -8,6 +8,7 @@ from typing import Dict, List, Optional, Set, Tuple
 import aiocron
 from sqlalchemy import and_, func, select, text, true
 
+import server.metrics as metrics
 from .abc.base_game import InitMode
 from .config import config
 from .core import Service
@@ -133,6 +134,8 @@ class LadderService(Service):
             if queue_name not in db_queues:
                 self.queues[queue_name].shutdown()
                 del self.queues[queue_name]
+
+        metrics.ladder_service_searches_dict_length.set(len(self._searches))
 
     async def fetch_map_pools(self, conn) -> Dict[int, Tuple[str, List[Map]]]:
         result = await conn.execute(
--- a/server/matchmaker/matchmaker_queue.py
+++ b/server/matchmaker/matchmaker_queue.py
@@ -102,6 +102,8 @@ class MatchmakerQueue:
 
             number_of_matches = len(self._matches)
             metrics.matches.labels(self.name).set(number_of_matches)
+            number_of_searches_in_queue = len(self.queue)
+            metrics.matchmaker_queue_length.labels(self.name).set(number_of_searches_in_queue)
 
             # TODO: Move this into algorithm, then don't need to recalculate quality_with?
             # Probably not a major bottleneck though.
--- a/server/metrics.py
+++ b/server/metrics.py
@@ -37,6 +37,12 @@ matchmaker_queue_pop = Gauge(
     ["queue"],
 )
 
+matchmaker_queue_length = Gauge(
+    "server_matchmaker_queue_length",
+    "Number of items in queue",
+    ["queue"],
+)
+
 # =====
 # Users
 # =====
@@ -109,4 +115,13 @@ active_games = Gauge(
 # ==============
 rating_service_backlog = Gauge(
     "server_rating_service_backlog", "Number of games remaining to be rated",
+)
+
+
+# ==============
+# Ladder Service
+# ==============
+ladder_service_searches_dict_length = Gauge(
+    "ladder_service_searches_dict_length",
+    "Number of items in LadderService._searches",
 )


server/matchmaker/matchmaker_queue.py
```diff
@@ -102,6 +102,8 @@ async def queue_pop_timer(self) -> None:
 
             number_of_matches = len(self._matches)
             metrics.matches.labels(self.name).set(number_of_matches)
+            number_of_items_in_queue = len(self.queue)
```
cleborys, line 105: +1 for descriptive names! Maybe replace `items` by `searches`?

server/metrics.py
```diff
@@ -103,10 +113,24 @@
     ["game_mode", "game_state"],
 )
 
+game_service_dict_length = Gauge(
```
cleborys, line 116: Is this the same as the sum over all reported numbers for `active_games` above?

cleborys (commented):
Welcome, nice to have you here!
I took the liberty to reply because I was the one to add the prometheus module a while back - I am currently super busy with work, but hopefully getting back to FAF next week, actually!

These metrics will eventually be collected on a grafana dashboard (https://grafana.faforever.com), @Askaholic can provide you with an account if you want.
There is also a Zulip chat (instead of slack) for developers if you want to stick around.

server/metrics.py
```diff
@@ -61,6 +67,10 @@
     "Number of users currently online as per lobbyconnection.player_service",
 )
 
+player_service_dict_length = Gauge(
```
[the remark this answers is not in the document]
46bit: Turns out they are. I'll drop this extra metric.

server/game_service.py
```diff
@@ -82,8 +82,10 @@ async def initialise_game_counter(self):
     async def update_data(self):
         """
         Loads from the database the mostly-constant things that it doesn't make sense to query every
-        time we need, but which can in principle change over time.
+        time we need, but which can in principle change over time. Also update some metrics.
         """
+        metrics.game_service_dict_length.set(len(self._games))
```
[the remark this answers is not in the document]
46bit: Hey. I didn't want to add an extra scheduled job, but now I know the codebase more I can try and integrate this a bit better.

server/metrics.py
46bit: Yep.

server/matchmaker/matchmaker_queue.py
46bit: Done

46bit:
`LadderService._searches` is an interesting one to monitor. We probably want to count:

1. the number of items in `LadderService._searches` (the total number of different players)
2. the number of sub-items (the total number of searches from all players)

We could file the player names as labels but I'm pretty sure that Prometheus doesn't like having that many different labels. So both of these would be simple numbers. Not sure what to call (2).

46bit:
And hi @cleborys! Thanks for that. @Askaholic It'd be great to get onto Zulip (and Grafana if that's OK.)

Askaholic:
Yea I can arrange that, I just need an email address. I can use the one from your website, dig one out of your git commits or you can pm one to me on the FAF discord.

cleborys:
> We could file the player names as labels but I'm pretty sure that Prometheus doesn't like having that many different labels.

Yes, they specifically warn you of using user IDs as labels: https://prometheus.io/docs/practices/naming/#labels

server/matchmaker/matchmaker_queue.py
```diff
@@ -102,6 +102,8 @@ async def queue_pop_timer(self) -> None:
 
             number_of_matches = len(self._matches)
             metrics.matches.labels(self.name).set(number_of_matches)
+            number_of_searches_in_queue = len(self.queue)
```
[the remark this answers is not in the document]
cleborys: ...and is `number_of_searches_in_queue` then equal to `number_of_unmatched_searches`?
I cannot tell whether `number_of_searches_in_queue` is meant to report the number of searches before matching was done or after. As it is now I believe it is reporting "after", which we already report.

I think we are currently reporting the number of matches made (during the current queue pop),
as well as the number of searches remaining in the queue after the queue pop.
The number of searches in the queue before matching would then (currently) be `2*number_of_matches + number_of_unmatched_searches`, although that won't hold anymore as soon as the matchmaker handles non-2v2 games.

What you could do instead is update the `searches_in_queue` metric in `MatchmakerQueue.search`, to have a value that is reported not only during queue pops, but also in between?
I'm thinking of these lines:
https://github.com/FAForever/server/blob/a024a2fe8c8b7b905a5bdd41795f8134ee01e677/server/matchmaker/matchmaker_queue.py#L134-L135
https://github.com/FAForever/server/blob/a024a2fe8c8b7b905a5bdd41795f8134ee01e677/server/matchmaker/matchmaker_queue.py#L144-L145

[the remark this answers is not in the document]
cleborys: Would we then expect the sum over all matchmaker name labels to be the what `ladder_service_searches_dict_length` is currently reporting? Equivalently, is every `ladder_service` search being registered with a `MatchmakerQueue` or can they ever get out of sync?

It might still be good to report `ladder_service_searches_dict_length` separately as you are currently suggesting, in case there will be some bug causing them to get out of sync even if we don't expect them to.

server/ladder_service.py
```diff
@@ -134,6 +135,8 @@ async def update_data(self) -> None:
                 self.queues[queue_name].shutdown()
                 del self.queues[queue_name]
 
+        metrics.ladder_service_searches_dict_length.set(len(self._searches))
```
cleborys, line 138: I would suggest moving this to `LadderService.start_search` and `LadderService.cancel_search` for real-time updates.

46bit:
It's pretty clear that what needs doing isn't as figured out as I'd understood from #609. I'm going to close this and focus on issues a bit readier-to-go.

PR #648, opened 2020-08-28
Use consistent quote marks

46bit:
Replaces Python `'` with `"`.


server/ladder_service.py
```diff
@@ -115,7 +115,7 @@ async def update_data(self) -> None:
                 map_pool_name, map_list = map_pool_maps[map_pool_id]
                 if not map_list:
                     self._logger.warning(
-                        "Map pool '%s' is empty! Some %s games will "
+                        "Map pool \"%s\" is empty! Some %s games will "
```
Askaholic, line 118: You can leave single quotes inside of strings

server/lobbyconnection.py
```diff
@@ -419,34 +445,34 @@ def _set_user_agent_and_version(self, user_agent, version):
...
 
-        if not self.user_agent or 'downlords-faf-client' not in self.user_agent:
+        if not self.user_agent or "downlords-faf-client' not in self.user_agent:
             await self.send_warning(
                 "You are using an unofficial client version! "
                 "Some features might not work as expected. "
                 "If you experience any problems please download the latest "
                 "version of the official client from "
-                f'<a href="{config.WWW_URL}">{config.WWW_URL}</a>'
+                f"<a href=\"{config.WWW_URL}\">{config.WWW_URL}</a>"
```
Askaholic, line 459: And here's a situation where I think single quotes are appropriate in order to avoid backslash escapes.

tests/integration_tests/test_server.py
```diff
@@ -153,69 +153,69 @@ async def test_game_info_broadcast_to_friends(lobby_server):
...
 
-    msg = await read_until_command(proto, 'game_info')
+    msg = await read_until_command(proto, "game_info")
 
-    assert msg['title'] == 'test\'s game'
-    assert msg['game_type'] == 'custom'
-    assert msg['mapname'] == 'scmp_007'
-    assert msg['map_file_path'] == 'maps/scmp_007.zip'
-    assert msg['featured_mod'] == 'faf'
+    assert msg["title"] == "test\"s game"
```
Askaholic, line 190: can single quote this one

tests/integration_tests/test_server.py
```diff
@@ -153,69 +153,69 @@ async def test_game_info_broadcast_to_friends(lobby_server):
...
 
-    msg = await read_until_command(proto, 'game_info')
+    msg = await read_until_command(proto, "game_info")
 
-    assert msg['title'] == 'test\'s game'
-    assert msg['mapname'] == 'scmp_007'
-    assert msg['map_file_path'] == 'maps/scmp_007.zip'
-    assert msg['featured_mod'] == 'coop'
-    assert msg['game_type'] == 'coop'
+    assert msg["title"] == "test\"s game"
```
Askaholic, line 214: and this one

tests/unit_tests/test_game.py
```diff
@@ -725,15 +725,15 @@ async def test_partial_stats_not_affecting_rating_persistence(
...
     old_mean = players[0].ratings[RatingType.GLOBAL][0]
 
     await game.launch()
     game.launched_at = time.time() - 60 * 60
-    await game.add_result(0, 0, 'victory', 10)
-    await game.add_result(0, 1, 'defeat', -10)
-    game.report_army_stats('{"stats": []}')
+    await game.add_result(0, 0, "victory", 10)
+    await game.add_result(0, 1, "defeat", -10)
+    game.report_army_stats("{\"stats\": []}")
```
Askaholic, line 736: and here

tests/unit_tests/test_protocol.py
```diff
@@ -84,42 +84,42 @@ async def test_types():
 
 
 async def test_QDataStreamProtocol_recv_small_message(protocol, reader):
-    data = QDataStreamProtocol.pack_block(b''.join([QDataStreamProtocol.pack_qstring('{"some_header": true}'),
-                                                    QDataStreamProtocol.pack_qstring('Goodbye')]))
+    data = QDataStreamProtocol.pack_block(b"".join([QDataStreamProtocol.pack_qstring("{\"some_header\": true}"),
```
Askaholic, line 87: and here

tests/unit_tests/test_protocol.py
```diff
@@ -84,42 +84,42 @@ async def test_types():
...
 
     with pytest.raises(asyncio.IncompleteReadError):
         await protocol.read_message()
 
 
 async def test_QDataStreamProtocol_recv_large_array(protocol, reader):
-    reader.feed_data(QDataStreamProtocol.pack_block(b''.join(
-        [QDataStreamProtocol.pack_qstring('{"some_header": true}')] +
+    reader.feed_data(QDataStreamProtocol.pack_block(b"".join(
+        [QDataStreamProtocol.pack_qstring("{\"some_header\": true}")] +
```
Askaholic, line 106: here

Askaholic (commented):
Nice! It will be good to get this out of the way. I marked a few place where single quotes would be appropriate to make the string more readable by eliminating backslash escapes.

46bit:
@Askaholic I've been anticipating that you'd use a linter (Codacy?) to prevent any future use of single quotes. That's why I changed absolutely everything to double quotes. Can change the messy things back to single quotes if you'd still prefer.

Askaholic:
I wasn't going to add any CI tools to check for it. I wouldn't even know of any which have such a check, plus codacy is unfortunately not very configurable AFAIK. I think it uses pylint and you can't change that, but that's just what I heard from the org admin. 

Ultimately the goal is clean code, and I think enforcing double quotes all the time isn't the right call as there will be times where using single quotes is more readable. However, i think the decision of which quote to use should be consistent, so I've decided that the rule should be: default to double quotes unless a different quote makes the string more readable.

46bit:
Should be good to go now.

Askaholic (approved):
Great! BTW what tool did you use for this?

PR #650, opened 2020-08-29
Use UTC instead of local timezone for determining remaining ban time

Askaholic:
Luckily since the server is located in Germany (at most UTC+2) this hasn't had any meaningful impact on ban durations. Also I would imagine that Travis builds are run in Germany too, so our tests wouldn't pick up on this either. It's just kindof funny that I keep finding these timezone discrepancies.

Also reformatted some stuff long lines.

### DIFF
--- a/server/lobbyconnection.py
+++ b/server/lobbyconnection.py
@@ -68,12 +68,14 @@ class BanError(Exception):
                 f"{self.ban_reason}")
 
     def _ban_duration_text(self):
-        ban_duration = self.ban_expiry - datetime.now()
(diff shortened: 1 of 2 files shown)


Brutus5000:
This shouldn't make a difference, the hetzner server operates in utc.

Askaholic:
Well it’s interesting because it seems that the MySQL `now` function called in the test data file returns utc times, but it’s good to know that it won’t change anything on prod.

cleborys (approved):
LGTM

46bit:
Good catch

PR #652, opened 2020-08-29
Issue/#636 Refactor game validity checks

### LINKED ISSUE #636: Refactor game validity checks
When I added the coop validity checks initially, I added them all on the base `Game` class with a switch that checked the featured mod. However in hindsight, I should have made use of the existing object oriented design and put the coop checks in the `CoopGame` subclass. This would have the added benefit of decoupling the coop validity checks from the featured mod. We should also do a similar refactor for the "faf validity checks" as well.

Related code:
https://github.com/FAForever/server/blob/e4ca993a9b7ea8123c64f6917bf9edd05f69b0ce/server/games/game.py#L587

46bit:
This commit closes #636. It restructures the validity checks that are performed on games. It starts taking advantage of the OOP setup of Game classes. Child classes of Game can redefine the `_validate_game_mode_settings` method to set custom game validity rules.

As discussed in #636 there is probably more tidyup to be done of these classes and how featured mods/game types work. I'm leaving that for later because I don't feel confident attempting it yet.

### DIFF
--- a/server/games/coop.py
+++ b/server/games/coop.py
@@ -20,3 +20,17 @@ class CoopGame(Game):
             "Difficulty": 3,
             "Expansion": 1
         })
+
+    async def validate_game_mode_settings(self):
+        """
+        Checks which only apply to the coop mode
+        """
+
+        valid_options = {
+            "Victory": (Victory.SANDBOX, ValidityState.WRONG_VICTORY_CONDITION),
+            "TeamSpawn": ("fixed", ValidityState.SPAWN_NOT_FIXED),
+            "RevealedCivilians": ("No", ValidityState.CIVILIANS_REVEALED),
+            "Difficulty": (3, ValidityState.WRONG_DIFFICULTY),
+            "Expansion": (1, ValidityState.EXPANSION_DISABLED),
+        }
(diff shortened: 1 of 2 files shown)


server/games/coop.py
```diff
@@ -20,3 +20,17 @@ def __init__(self, *args, **kwargs):
             'Difficulty': 3,
             'Expansion': 1
         })
+
+    async def _validate_game_mode_settings(self):
```
Askaholic, line 24: Would it make sense to make this a "public" function?

Askaholic (approved):
Simple enough changes, seems good to me.

cleborys (approved):
Great!
I guess this interface only supports restrictions like "Option X must be set to value Y and no other value" rather than "Option X must be set to one of Y1, Y2, Y3", but it seems that's all we need!

server/games/coop.py
46bit: I've fixed that. Hopefully good to merge now.

Askaholic:
Need to rebase please

46bit:
@Askaholic good to go

Askaholic:
Thanks!

PR #654, opened 2020-09-03
Feature/#599 Link games with their originating matchmaker queue

### LINKED ISSUE #599: Link games with their originating matchmaker queue
For the anti map repetition code to work we need a way of telling when a game was started from a particular queue. Right now there is only one queue games started from it have the gametype `ladder1v1`. But for future tmm queues this will not be the case. 

## Proposed solutions

1. Add a nullable `matchmaker_queue_id` column to `game_stats` table.
The problem is that `game_stats` is a big table, and the majority of games will have a null value here. This is a lot of wasted space.
2. Introduce some sort of m:n table for linking queues to games. 
Takes less space, but requires an extra join when querying. Can also link multiple queues to the same game, but I'm not sure what the use case for that would be.

cleborys:
I went with proposed option 2 in issue #599: Add a many-to-many relationship table linking`matchmaker_queue` to `game_stats` (= less space than a nullable field, but one more join when reading or writing).

I thought about adding the database_id to the `MatchmakerQueue` object in addition to the `name` field, but then decided against it since there is still the legacy `ladder1v1` fallback matchmaker queue, which would not have a meaningful `id`.
That means one additional additional join/select statement when reading/writing the relationship, since we need to translate the name to an id.

If we wanted to make sure every game has at most one queue attached, we could change the primary key of `matchmaker_queue_game` to the `game_stats_id`.

I'm including the database migration here for now.

Closes #599


matchmaker_queue_game_migration.sql
```diff
@@ -0,0 +1,10 @@
+DROP TABLE IF EXISTS matchmaker_queue_game;
+
+CREATE TABLE matchmaker_queue_game
+(
+  matchmaker_queue_id INT(10) UNSIGNED NOT NULL,
+  game_stats_id       INT(10) UNSIGNED NOT NULL,
+  FOREIGN KEY (matchmaker_queue_id) REFERENCES matchmaker_queue (id),
+  FOREIGN KEY (game_stats_id) REFERENCES game_stats (id),
+  PRIMARY KEY (game_stats_id, matchmaker_queue_id)
```
Askaholic, line 9: I think you'll want to add an `id` column anyways. AFAIK @Brutus5000 ended up needing these for the API, because of Elide.

server/games/game.py
```diff
@@ -703,6 +710,17 @@ async def update_game_stats(self):
                 )
             )
 
+            if self.matchmaker_queue_name is not None:
+                queue_id_select = select([matchmaker_queue.c.id]).where(
+                    matchmaker_queue.c.technical_name == self.matchmaker_queue_name
+                )
```
Askaholic, line 716: Why not take a `MatchmakerQueue` object instead of just the name, then you will already have access to the id?

cleborys: The object doesn't store the id currently. We can add and either pass it or the id only, but right now there is still the hardcoded `ladder1v1` fallback queue, which wouldn't have a database id attached:
https://github.com/FAForever/server/blob/1ec2a7bd145ea4360ed38f76ea5705560098472e/server/ladder_service.py#L54-L64

We could of course make the id field optional and give this one a `None` id if you think that is the better way.

Askaholic: We don't need the fallback anymore, we can remove it. It's already in progress here #623, but hasn't been quite finished.

cleborys: Maybe I'll take over #623 then, since it's been some time? If that's alright with @RemcoSchrijver.
Then afterwards I'll add the id field.

Askaholic: Yea that's a good idea. You can just pull his commits into a new branch and finish deleting the related fallback stuff.

cleborys: Added the id, now depends on #656

server/matchmaker/matchmaker_queue.py
```diff
@@ -40,13 +40,15 @@ def __init__(
         self,
         game_service: "GameService",
         name: str,
+        id: int,
```
Askaholic, line 43: Do we worry about shadowing a builtin here? It's definitely fine for the attribute, but idk about function parameter

Askaholic (commented):
Do we have some tests for anti map repetition for tmm games?

cleborys:
Added tests and removed the manual migration necessary to add the new table, since it is now in FAForever/db.
Since it's not released as a proper version with tag yet, I set the `FAF_DB_VERSION` to in `.travis.yml` to the appropriate commit hash.

Askaholic (approved):
Looks good, I think we wait for the DB version to be released before merging. Then I can test it on the test server first. Shouldn't be too long before it gets tagged.

PR #656, opened 2020-09-03
Issue/#581 Remove hardcoded ladder1v1 queue and map pool

### LINKED ISSUE #581: Remove old ladder map pool

cleborys:
Closes #581

### DIFF
--- a/server/ladder_service.py
+++ b/server/ladder_service.py
@@ -58,8 +58,6 @@ class LadderService(Service):
         await self.update_data()
         self._update_cron = aiocron.crontab("*/10 * * * *", func=self.update_data)
 
-        self.start_queue_handlers()
-
     async def update_data(self) -> None:
         async with self._db.acquire() as conn:
             map_pool_maps = await self.fetch_map_pools(conn)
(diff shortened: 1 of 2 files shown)


server/ladder_service.py
```diff
@@ -50,46 +50,18 @@ def __init__(
...
 
     async def initialize(self) -> None:
-        # TODO: Starting hardcoded queues here
-        for queue in self.queues.values():
-            queue.initialize()
-
         await self.update_data()
         self._update_cron = aiocron.crontab("*/10 * * * *", func=self.update_data)
 
         self.start_queue_handlers()
```
Askaholic, line 61: This function needs to go as well. Right now each queue ends up getting two handlers trying to pull from it at the same time.

Askaholic (changes requested):
I also noticed the double queue handler problem in my multiqueue refactor. Not sure if there's a nice way to test for it, but since I'm reworking that stuff anyways there isn't really a need to test it.

Askaholic (approved):

PR #665, opened 2020-09-17
Refactor weakrefs in Player object

Askaholic:
I stumbled across this `WeakAttribute` implementation while looking up something unrelated and realized that we actually have a use case for it. This will also make it easier to add weak references on other objects too, not just Player. I also refactored a few other things in Player while I was at it.

### DIFF
--- a/server/players.py
+++ b/server/players.py
@@ -1,5 +1,6 @@
 from contextlib import suppress
 from enum import Enum, unique
+from typing import Optional, Union
 
 from server.config import config
 from server.rating import PlayerRatings, RatingType, RatingTypeMap
@@ -11,11 +12,11 @@ from .weakattr import WeakAttribute
 
 @unique
 class PlayerState(Enum):
-    IDLE = 1,
-    PLAYING = 2,
-    HOSTING = 3,
-    JOINING = 4,
-    SEARCHING_LADDER = 5,
+    IDLE = 1
+    PLAYING = 2
+    HOSTING = 3
+    JOINING = 4
+    SEARCHING_LADDER = 5
 
 
 class Player:
@@ -38,8 +39,8 @@ class Player:
         ratings=None,
         clan=None,
         game_count=None,
-        lobby_connection: "LobbyConnection" = None
-    ):
+        lobby_connection: Optional["LobbyConnection"] = None
+    ) -> None:
         self._faction = Faction.uef
 
         self.id = player_id
@@ -74,11 +75,11 @@ class Player:
             self.lobby_connection = lobby_connection
 
     @property
-    def faction(self):
+    def faction(self) -> Faction:
         return self._faction
 
     @faction.setter
-    def faction(self, value):
+    def faction(self, value: Union[str, int, Faction]) -> None:
         if isinstance(value, str):
             self._faction = Faction.from_string(value)
         elif isinstance(value, int):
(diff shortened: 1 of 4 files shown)


server/players.py
```diff
@@ -66,21 +66,19 @@ def __init__(
...
-
-        self._lobby_connection = lambda: None
         if lobby_connection is not None:
             self.lobby_connection = lobby_connection
 
-        self._game = lambda: None
-        self._game_connection = lambda: None
+    lobby_connection = WeakAttribute["LobbyConnection"]()
+    game = WeakAttribute["Game"]()
+    game_connection = WeakAttribute["GameConnection"]()
```
cleborys, line 74: I would have put these before the `__init__`?

server/weakattr.py
```diff
@@ -0,0 +1,29 @@
...
+    """
+
+    def __set_name__(self, _: object, name: str) -> None:
+        self.name = name
+
+    def __get__(self, obj: object, objclass: type) -> Optional[T]:
+        ref = obj.__dict__.get(self.name)
+        if ref:
+            return ref()
+        return None
```
cleborys, line 22: Here is a shorter alternative that might or might not be more readable :stuck_out_tongue_closed_eyes: 
```suggestion
        ref = obj.__dict__.get(self.name, lambda: None)
        return ref()
```

cleborys (approved):
Cool! Took me a bit to understand what's happening under the hood, but the resulting syntax is very clear!

I noticed that there is no test checking that the attribute is indeed a weak reference, e.g. setting it, running `gc.collect()` and checking whether it was collected.

Askaholic:
There is a test for that actually, it already exists and is called `test_weak_references`, so it doesn't show up in the git diff

server/players.py
Askaholic: Yea maybe that's better, I wasn't sure where to put them since they're replacing the property definitions and those were after init.

tests/unit_tests/test_weakattr.py
```diff
@@ -0,0 +1,41 @@
...
+import pytest
+
+from server.weakattr import WeakAttribute
+
+
+@pytest.fixture(params=("typed", "untyped"))
+def obj(request):
+    """An object with a WeakAttribute"""
+    if request.param == "untyped":
+        class Foo(object):
```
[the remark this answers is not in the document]
Askaholic: Every class inherits from `object` implicitly, this is just making it explicit I guess. BTW metaclasses inherit from `type` instead of `object`, that's what makes them different.

server/weakattr.py
Askaholic: I don't think it's unreadable enough to do that. Besides, my way is marginally faster because it avoids creating a temporary function and avoids one function call when the item isn't found :P 

```python
>>> timeit.timeit("""
... a = {}.get(1)
... if a:
...     a()
... """)
0.07964258699212223
>>> timeit.timeit("{}.get(1, lambda: None)()")
0.13527880399487913
>>> timeit.timeit("""
... a = {1: func}.get(1)
... if a:
...     a()
... """, "func = lambda: 10")
0.16154761402867734
>>> timeit.timeit("{1: func}.get(1, lambda: None)()", "func = lambda: 10")
0.18628269800683483
```

I added explicit `return None` because mypy was complaining, although it isn't technically necessary.

cleborys: touché

PR #670, opened 2020-09-24
Hopefix for infinite login

Askaholic:
We were not able to reproduce the issue yet, but in theory it might be possible that the await statement on sending the warning to old player connections could prevent the welcome message from being delivered to the new connection.

Uses the `write` variant (instead of `send`) when notifying the old connection of the new login.

### DIFF
--- a/server/lobbyconnection.py
+++ b/server/lobbyconnection.py
@@ -361,20 +361,17 @@ class LobbyConnection:
             if await self.player_service.has_permission_role(
                 self.player, "ADMIN_BROADCAST_MESSAGE"
             ):
-                tasks = []
                 for player in self.player_service:
                     # Check if object still exists:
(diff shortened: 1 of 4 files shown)


axel1200 (approved):

server/lobbyconnection.py
```diff
@@ -362,20 +362,17 @@ async def command_admin(self, message):
...
             ):
-                tasks = []
                 for player in self.player_service:
                     # Check if object still exists:
                     # https://docs.python.org/3/library/weakref.html#weak-reference-objects
                     if player.lobby_connection is not None:
-                        tasks.append(
-                            player.lobby_connection.send_warning(message_text)
-                        )
+                        with contextlib.suppress(DisconnectedError):
```
Askaholic, line 369: Either this is needed in which case it should be used in both places, or it's not needed in which case it shouldn't be used in either place

PR #673, opened 2020-09-26
Issue/#621 Detect when some players don't join a matchmaker game

### LINKED ISSUE #621: Wait for all guests to join a ladder game
The `start_game` function in `LadderService` should wait for all expected guests to have joined the before printing the "game started successfully" message. If some players do not join within a certain timeout (maybe a minute or two), then the match should be aborted and the normal match timeout logic should be triggered (sending `match_cancelled`, etc).

Would require some modification of `Game` to possibly trigger a future when a player joins, and then adding some code to wait for those futures in `LadderService`.

Askaholic:
Adds logic for sending `match_cancelled` messages if the matchmaker game doesn't start within about a minute. This should fix point 4 mentioned here https://github.com/FAForever/downlords-faf-client/pull/1778#issuecomment-660403082. 

Seems like some of the mocking code in the tests was not working correctly, and they were waiting for 30 seconds for the game timeout to trigger before passing. Now that's been fixed and the tests run about 30 seconds faster.

Closes #621

### DIFF
--- a/server/games/game.py
+++ b/server/games/game.py
@@ -115,6 +115,7 @@ class Game:
         }
         self.mods = {}
         self._is_hosted = asyncio.Future()
+        self._launch_fut = asyncio.Future()
 
         self._logger.debug("%s created", self)
         asyncio.get_event_loop().create_task(self.timeout_game())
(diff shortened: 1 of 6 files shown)


0x647262 (approved):
Tests pass, code is clean.

LGTM!

PR #674, opened 2020-09-29
#663 - Phantom games report format

### LINKED ISSUE #663: Phantom games report results in a different format
Right now the result reporting assumes that results always have the form `string number`, which is correct for vanilla games. However, as it turns out, the phantom mod also sends an additional string indicating whether or not the player was a phantom. Maybe it would be worth rewriting the game result processing to handle this format as well. At the very least it would make the scores be saved in the database for phantom games.

```
WARNING  Sep 05  03:10:18 GameConnection                 Invalid result for 5 reported: innocent defeat -3
WARNING  Sep 05  03:17:05 GameConnection                 Invalid result for 3 reported: phantom defeat -1
```

Source code:
https://github.com/FAForever/server/blob/ccf349e3b4bbaee95373d79a0e2f4ee9c5d064a2/server/gameconnection.py#L292-L296

ahsanbagwan:
Closes #663

### DIFF
--- a/tests/unit_tests/test_gameconnection.py
+++ b/tests/unit_tests/test_gameconnection.py
@@ -545,3 +545,9 @@ async def test_handle_action_invalid(game_connection: GameConnection):
 
     game_connection.abort.assert_not_called()
     game_connection.protocol.send_message.assert_not_called()
+
+async def test_result_format_phantom(game: Game, game_connection: GameConnection):
+    game_connection.connect_to_host = CoroutineMock()
+
+    await game_connection.handle_action("GameResult", [0, "phantom score -5"])
+    game.add_result.assert_called_once_with(game_connection.player.id, 0, "score", -5)
--- a/tests/unit_tests/test_gameconnection.py
+++ b/tests/unit_tests/test_gameconnection.py
@@ -547,7 +547,5 @@ async def test_handle_action_invalid(game_connection: GameConnection):
     game_connection.protocol.send_message.assert_not_called()
 
 async def test_result_format_phantom(game: Game, game_connection: GameConnection):
-    game_connection.connect_to_host = CoroutineMock()
-
     await game_connection.handle_action("GameResult", [0, "phantom score -5"])
     game.add_result.assert_called_once_with(game_connection.player.id, 0, "score", -5)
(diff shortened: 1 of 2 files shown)


Askaholic:
Can you add a unit test to `test_gameconnextion.py` please?

ahsanbagwan:
I'm not sure how we could effectively test this behaviour at this point. Code was changed around `result` and it being local to function, there is no way to intercept it iirc.

Extracting this to a separate method and then testing its behavior seems like a way to go.

Askaholic:
Use a mock object (CoroutineMock for add_result because it's async). Dispatch a message similar to the other tests in that file and assert that `game.add_result` was called with the appropriate arguments.

tests/unit_tests/test_gameconnection.py
```diff
@@ -545,3 +545,9 @@ async def test_handle_action_invalid(game_connection: GameConnection):
 
     game_connection.abort.assert_not_called()
     game_connection.protocol.send_message.assert_not_called()
+
+async def test_result_format_phantom(game: Game, game_connection: GameConnection):
+    game_connection.connect_to_host = CoroutineMock()
```
Askaholic, line 550: Is this needed?

tests/unit_tests/test_gameconnection.py
```diff
@@ -545,3 +545,9 @@ async def test_handle_action_invalid(game_connection: GameConnection):
...
+
+    await game_connection.handle_action("GameResult", [0, "phantom score -5"])
+    game.add_result.assert_called_once_with(game_connection.player.id, 0, "score", -5)
```
Askaholic, line 553: Add newline please

Askaholic (changes requested):
Yes, that's all we need! Just some housekeeping things.

tests/unit_tests/test_gameconnection.py
ahsanbagwan: I'm sorry I don't follow which line you are pointing to.

tests/unit_tests/test_gameconnection.py
ahsanbagwan: Will do.

tests/unit_tests/test_gameconnection.py
Askaholic: 550, The one where you're setting `connect_to_host` to a mock object. I don't think `connect_to_host` is triggered by the `GameResult` command.

Askaholic (approved):

PR #677, opened 2020-10-02
Enhancement/#676 remove python version check

### LINKED ISSUE #676: Remove python client version check
I would like to drop the table `version_lobby` in the database. The whole version check is sort of pointless today. The database records point to the download of the java client and the minimum version is 0.13.0 which is like 5 years old. Seems like it's completely useless.

Both clients have an auto-update mechanism outside of the faf-database.

Related code (didn't check for tests):
https://github.com/FAForever/server/blob/2932f863432ff2e9ff0d33d203be2f24546c8aee/server/player_service.py#L40
https://github.com/FAForever/server/blob/2932f863432ff2e9ff0d33d203be2f24546c8aee/server/player_service.py#L259-L265
https://github.com/FAForever/server/blob/2932f863432ff2e9ff0d33d203be2f24546c8aee/server/lobbyconnection.py#L458-L478

andrewmiko:
Removed `client_version_info` from project.

1. `isort` show sorting error in file which I didn't edited
2. `flake8` throws tons of errors on local env
3. Didn't update Pipfile only because it's trying to upgrade all packages (it's ignoring `--keep-outdated` flag)

I can update PR later with all necessary changes.

Closes #676

### DIFF
--- a/server/lobbyconnection.py
+++ b/server/lobbyconnection.py
@@ -11,7 +11,6 @@ from typing import Optional
 import aiohttp
 import humanize
 import pymysql
-import semver
 from sqlalchemy import and_, func, select
 
 import server.metrics as metrics
@@ -454,45 +453,6 @@ class LobbyConnection:
             metrics.user_agent_version.labels(str(version)).inc()
         self._version = version
 
-    async def _check_version(self):
-        versionDB, updateFile = self.player_service.client_version_info
-        update_msg = {
-            "command": "update",
-            "update": updateFile,
-            "new_version": versionDB
-        }
-
-        if not self.user_agent or "downlords-faf-client" not in self.user_agent:
-            await self.send_warning(
-                "You are using an unofficial client version! "
-                "Some features might not work as expected. "
-                "If you experience any problems please download the latest "
-                "version of the official client from "
-                f'<a href="{config.WWW_URL}">{config.WWW_URL}</a>'
-            )
-
-        if not self._version or not self.user_agent:
-            update_msg["command"] = "welcome"
-            # For compatibility with 0.10.x updating mechanism
-            await self.send(update_msg)
-            return False
-
-        # Check their client is reporting the right version number.
-        if "downlords-faf-client" not in self.user_agent:
-            try:
-                version = self._version
-                if "-" in version:
-                    version = version.split("-")[0]
-                if "+" in version:
-                    version = version.split("+")[0]
-                if semver.compare(versionDB, version) > 0:
-                    await self.send(update_msg)
-                    return False
-            except ValueError:
-                await self.send(update_msg)
-                return False
-        return True
-
     async def check_policy_conformity(self, player_id, uid_hash, session, ignore_result=False):
         if not config.USE_POLICY_SERVER:
             return True
@@ -738,9 +698,7 @@ class LobbyConnection:
         user_agent = message.get("user_agent")
         version = message.get("version")
         self._set_user_agent_and_version(user_agent, version)
-
-        if await self._check_version():
-            await self.send({"command": "session", "session": self.session})
+        await self.send({"command": "session", "session": self.session})
 
     async def command_avatar(self, message):
         action = message["action"]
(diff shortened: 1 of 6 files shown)


server/lobbyconnection.py
```diff
@@ -454,45 +453,6 @@ def _set_user_agent_and_version(self, user_agent, version):
...
-        }
-
-        if not self.user_agent or "downlords-faf-client" not in self.user_agent:
-            await self.send_warning(
-                "You are using an unofficial client version! "
-                "Some features might not work as expected. "
-                "If you experience any problems please download the latest "
-                "version of the official client from "
-                f'<a href="{config.WWW_URL}">{config.WWW_URL}</a>'
-            )
```
Askaholic, line 472: I think we do want to keep the user agent check

tests/integration_tests/test_server.py
```diff
@@ -20,21 +20,6 @@
 TEST_ADDRESS = ("127.0.0.1", None)
 
 
-async def test_server_deprecated_client(lobby_server):
```
Askaholic, line 23: Same reason, I'd keep this test. We'll still have some legacy client users which will report the user agent `faf-client`.

Askaholic (commented):
@Brutus5000 Thoughts on keeping the user agent check?

Askaholic:
Oh and I can take care of the Pipfile and Pipfile.lock in a separate PR. That's no problem

server/lobbyconnection.py
Brutus5000: Yes, I agree.

andrewmiko: OK, I going to fix that.

tests/integration_tests/test_server.py
andrewmiko: OK

Askaholic (commented):
Another thing I just noticed was that there are references to the `version_lobby` table in our test-data.sql file. We should remove those too so our tests don't break when the table is dropped.

And you'll need to rebase.

andrewmiko:
@Askaholic 

OK, I'm first time doing rebase and I need little help here.

Current branch is ahead of develop to 10 commits:
+ 7aa21c60392125e518d4f74a5772a332cb1b7be9 Removed semver package deprecation warning from pytest
+ 6c26c31f26959dee6f7dfcbac854cbf33a4ce112 Removed _check_version method and semver package
+ 5105dbdabde65001d158dd7c6e79c655dff00dbb Removed client_version_info from PlayerService
+ 6f9742e7bf832f6330b4d53fcaae7ea03e97f799 Removed client_version_info from test_update_data unittest
+ e21242be0188360600189bfd30a3c4db009305b6 Removed test_server_deprecated_client test
+ b088c48e6afdf9dc72543cef7c7f6fd7e0fccd8b #663 - Phantom games report format (#674)
+ a321937f7119b7dbed7bb4d993335745c6d80005 Restored and renamed _check_version to _check_user_agent
+ fc9c508db503b7e8cd4fd03d24c47c1d2e32f85b Restored test_server_deprecated_client test
+ 537c86ffdb03e5fd40b66d2ed90c68bf71fac79f Feature/#599 Link games with their originating matchmaker queue (#654)
+ 446a9f4e71f1cad1de4779deadf704215cfa83b8 Removed version_lobby from test-data.sql

Should I go back to 10 commits and squash them? `git rebase -i HEAD~10`
Then `push --force`

Right?

Askaholic:
Sorry for taking so long to respond, somehow I overlooked this. 

I usually do this:
```
$ git checkout develop
$ git pull upstream develop
$ git checkout my-branch
$ git rebase develop
$ git push --force
```

Where `upstream` is the remote that points to this repository (the FAForever one, not my fork). I'm not entirely sure what the interaction with merge commits is here, because I never make any, but I would expect them to just disappear during the rebase.

andrewmiko:
@Askaholic Thanks a lot. Today I will finish it then. 👍

Askaholic (approved):
Looks good!

andrewmiko:
🙌  🎉

PR #680, opened 2020-10-05
Ignore cancellation of non-existing search without error

Askaholic:
During the TMM test we saw a lot of this exception:
```
ERROR    Oct 04  18:19:01 LobbyConnection                'ladder1v1'
 File "/code/server/lobbyconnection.py", line 147, in on_message_received
   await handler(message)
 File "/code/server/lobbyconnection.py", line 789, in wrapper
   return await func(self, message)
 File "/code/server/lobbyconnection.py", line 856, in command_game_matchmaking
   self.ladder_service.cancel_search(self.player, queue_name)
 File "/code/server/ladder_service.py", line 207, in cancel_search
   self._cancel_search(initiator, queue_name)
 File "/code/server/ladder_service.py", line 213, in _cancel_search
   cancelled_search = self._searches[initiator][queue_name]
KeyError: 'ladder1v1'
```

Which would cause players to be disconnected if their client sent a matchmaking stop command while the server already considered them out of queue. Now extra messages will just be ignored, as they likely just mean the client state was out of date.

### DIFF
--- a/server/ladder_service.py
+++ b/server/ladder_service.py
@@ -211,7 +211,15 @@ class LadderService(Service):
         """
         Cancel search for a specific player/queue.
         """
-        cancelled_search = self._searches[initiator][queue_name]
+        cancelled_search = self._searches[initiator].get(queue_name)
+        if cancelled_search is None:
+            self._logger.debug(
+                "Ignoring request to cancel a search that does not exist: "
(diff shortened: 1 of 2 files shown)


0x647262 (approved):
LGTM!

PR #686, opened 2020-10-18
WIP: Add game_type field to game_launch message

### LINKED ISSUE #685: Use game_type instead of init_mode in game_launch message

angelobzsouza:
Closes #685 

I'm having problem to config db connection. I don't know if it is because I'm using MacOS, but I still work on it. I just opened the PR to show I keep trying to close this issue.

### DIFF
--- a/server/lobbyconnection.py
+++ b/server/lobbyconnection.py
@@ -41,7 +41,7 @@ from .players import Player, PlayerState
 from .protocol import DisconnectedError, Protocol
 from .rating import InclusiveRange, RatingType
 from .types import Address, GameLaunchOptions
-
+from .games.typedefs import GameType
 
 class ClientError(Exception):
     """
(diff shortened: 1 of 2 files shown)


server/lobbyconnection.py
```diff
@@ -938,7 +938,8 @@ async def launch_game(
             # options are. Currently, options for ladder are hardcoded into the
             # client.
             "name": game.name,
-            "init_mode": game.init_mode.value,
+            "game_type": GameType.to_string(self.game_type),
+            "init_mode": game.init_mode.value, # init_mode is deprecated
```
Askaholic, line 942: You should add "use game_type instead" to the comment, and I'd put the comment on its own line (like the other comment)

server/lobbyconnection.py
```diff
@@ -938,7 +938,8 @@ async def launch_game(
             # options are. Currently, options for ladder are hardcoded into the
             # client.
             "name": game.name,
-            "init_mode": game.init_mode.value,
+            "game_type": GameType.to_string(self.game_type),
```
Askaholic, line 941:
```suggestion
            "game_type": GameType.to_string(game.game_type),
```
This will fix a lot of the failing tests

Askaholic (commented):
Like I said it's probably easier just to look at the output from Travis since you are not going to be changing a lot of code.

Other than these things it looks fine. You could also refactor `GameType` if you felt so inclined. There's really no need for these `to_string` and `from_string` methods if we make use of existing Enum features.

Askaholic:
Closing this as I have see no progress in a month

PR #689, opened 2020-11-21
Optimization based on profiling data

Askaholic:
The profiling data shows that `is_visible_to_player` is causing a bit more load than necessary. This function is called often, and was calling the game's `players` property which would recompute the set of players in the game. This was generating millions of calls to other parts of the code, which can all be avoided by using `self._connection` instead.

I also went ahead and refactored `players` to get rid of one redundant call to `get_player_option`.

### DIFF
--- a/server/games/game.py
+++ b/server/games/game.py
@@ -161,9 +161,11 @@ class Game:
         else:
             return frozenset(
                 player
-                for player in self._players
-                if self.get_player_option(player.id, "Army") is not None
-                and self.get_player_option(player.id, "Army") >= 0
+                for player, army in (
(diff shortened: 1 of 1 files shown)


Wesmania (approved):
lgtm

Wesmania:
How often is that property called? Do you have profiling info? We could cache it and update it when appropriate.

Askaholic:
Based on the profile I have, `players` was called around 3200 times per second, and about 3000 of those times were from `is_visible_to_player`. So I would expect that after these changes it will be down to somewhere around 200 times per second. I don't think it will be worth optimizing further.

0x647262 (approved):
LGTM!

PR #698, opened 2020-12-29
Issue/#691 prevent joining a party while it is queued

### LINKED ISSUE #691: TMM: Don't let players join a party while the party owner is searching
It's possible to get into a weird state where only a portion of players in a party are queued.

1. Party owner invites someone
2. Party owner joins a queue
3. Invited player accepts

In this situation, the party owner is queued, the invited player is not queued, however both players are in the same party. This causes weird issues in the client where the owner cannot kick players because he's queued, and cannot leave the queue because the party has too many players.

The server should check party members state when accepting an invite, and only allow the invite to be accepted if the players are searching.

Askaholic:
For now I'm still using the `ClientError` copout, but I have been thinking about ways to do it better in a future PR.

Fixes #691

### DIFF
--- a/server/party_service.py
+++ b/server/party_service.py
@@ -5,7 +5,7 @@ from .decorators import with_logger
 from .exceptions import ClientError
 from .factions import Faction
 from .game_service import GameService
-from .players import Player
+from .players import Player, PlayerState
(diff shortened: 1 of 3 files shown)


0x647262 (approved):
LGTM!

PR #701, opened 2021-01-07
Shuffle teams before assigning start spots

### LINKED ISSUE #700: TMM: Randomize Party Member Slots
The server should assign slots randomly for party members. Currently it is giving slot 1 and slot 3 directly to the respective party leaders, which means that every game people are playing the same spawns. Since many 2v2 maps (and later other teamgame maps) tend to have totally different roles/style of play for each slot, it would be nice if the server automatically let people experience other slots.

Askaholic:
Randomizes players start spot, but makes sure to randomize both teams in the same way, i.e. if player 1 on team 1 gets moved to slot 2, then player 1 on team 2 will be moved to slot 3, the opposing slot. Note that it does this after choosing the game name, so the party leader will still always appear in the game title.

Closes #700

### DIFF
--- a/server/ladder_service.py
+++ b/server/ladder_service.py
@@ -1,5 +1,6 @@
 import asyncio
 import itertools
+import random
 import re
 from collections import defaultdict
 from typing import Dict, List, Optional, Set, Tuple
@@ -335,6 +336,8 @@ class LadderService(Service):
         team2: List[Player],
         queue: MatchmakerQueue
     ) -> None:
+        assert len(team1) == len(team2)
+
         self._logger.debug(
             "Starting %s game between %s and %s", queue.name, team1, team2
         )
@@ -374,9 +377,19 @@ class LadderService(Service):
             game.init_mode = InitMode.AUTO_LOBBY
             game.map_file_path = map_path
 
-            for i, player in enumerate(alternate(team1, team2)):
-                if player is None:
-                    continue
+            def get_player_mean(player):
+                return player.ratings[queue.rating_type][0]
+
+            team1 = sorted(team1, key=get_player_mean)
+            team2 = sorted(team2, key=get_player_mean)
+
+            # Shuffle the teams such that direct opponents remain the same
+            zipped_teams = list(zip(team1, team2))
+            random.shuffle(zipped_teams)
+
+            for i, player in enumerate(
+                player for pair in zipped_teams for player in pair
+            ):
                 # FA uses lua and lua arrays are 1-indexed
                 slot = i + 1
                 # 2 if even, 3 if odd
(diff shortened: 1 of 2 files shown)


server/ladder_service.py
```diff
@@ -374,7 +377,12 @@ async def start_game(
             game.init_mode = InitMode.AUTO_LOBBY
             game.map_file_path = map_path
 
-            for i, player in enumerate(alternate(team1, team2)):
+            # Shuffle the teams such that direct opponents remain the same
+            shuffled_teams = zip(*random.sample(
+                list(zip(team1, team2)),
+                len(team1)
+            ))
+            for i, player in enumerate(alternate(*shuffled_teams)):
```
cleborys, line 385: I guess this makes sense if the `alternate` method already exists for other reasons, but to me it looks like you have the result of `random.sample`, which is a list of tuples and you flatten it as `alternate(*zip(*random.sample(...)))` instead of e.g. `[player for player_pair in random.sample(...) for player in player_pair]`. I find the latter easier to reason about, but that might of course be a matter of taste :)

server/ladder_service.py
```diff
@@ -335,6 +336,8 @@ async def start_game(
         team2: List[Player],
         queue: MatchmakerQueue
     ) -> None:
+        assert len(team1) == len(team2)
```
cleborys, line 339: How is this handled if it fails? (just asking since it seems to be somewhat unrelated)

cleborys (approved):
Happy New Year :) My main resolution is to try to make time for FAF again, so here is a slow start! Glad to see that TMM is doing so well!

The issue #700 mentions matching players up by rating - I assume that's not already done implicitly in the order of players on a team? (I.e. it wasn't done before and this produces the same matchups as before, just on randomized spots.)
Then maybe that needs to be a separate feature request that ftx meant to make in the same issue.

Askaholic:
> I assume that's not already done implicitly in the order of players on a team?

Hmm, for some reason I thought I had written code for that already, but now I'm not so sure anymore. It probably makes sense to do that inside of `start_game` so maybe I should add it there.

server/ladder_service.py
Askaholic: If it does fail, I think there would be a warning about an unhandled exception in a coroutine, but I only put it there because it is a precondition of the function. It's basically a way of saying to the programmer "never call this function with teams of different sizes", plus if the programmer ever does that, then hopefully it will cause a unit test to fail. 

I would like to actually run the server with python's `-O` option which removes all assert statements at some point, except that I do know there are 1 or 2 assert statements in the code (which were there when I took over) that do occasionally fail and show up in the prod server logs. So those would need to be fixed first.

server/ladder_service.py
Askaholic: Yea I think I like that better too. `alternate` made more sense before because the teams were in two separate lists. But since the best way I could think of for shuffling both lists in the same way was zipping them together, we have the result as a list of tuples anyways so it doesn't really make sense anymore.

0x647262 (approved):
LGTM!

PR #702, opened 2021-01-07
Use correct rating type when choosing a map pool

Askaholic:
The map pool selection was still hardcoded to use ladder rating, I must have missed this spot when switching to getting the rating type from the queue.


cleborys (approved):
You probably did this already, but I quickly checked that all further occurrences of `RatingType.LADDER_1V1` in the code look as intended👍

Askaholic:
Cool, yea I did another search through the project for the hardcoded uses as well.

0x647262 (approved):
LGTM!

PR #709, opened 2021-01-29
Add version to server user connections

Sheikah45:


Askaholic:
You also need to change the code where the metric is called to include the version

Sheikah45:
Is this correct?

Askaholic:
I think so. One test is failing because a mock object is missing the version field

Sheikah45:
Alright I fixed the test. Also got the server tests running up on windows again so if anyone comes along that needs help with that I can help.

Sheikah45:
hmmm another test is failing that doesnt fail on my setup

Askaholic:
Yea, it's a flaky test. It will most likely pass if it runs again. I'll merge something else and you can rebase.

PR #710, opened 2021-01-30
Fix integration tests

Askaholic:
A couple of things that I forgot to update in previous PR's.

- Host does not always get map position 1 because it is now randomized
- No search_info stop message for the queue if a match was found


0x647262 (approved):
LGTM!

PR #711, opened 2021-02-01
Matchmaker Learning Popup

FtXCommando:
Made the learning popup a bit more clear.


cleborys:
There were two unit test in which the old "The system is still learning you" was hardcoded and so they needed to be adapted. Since it wasn't good style to hardcode the message anyways and I assumed you (@FtXCommando) didn't want to dig into the tests, I made a quick new PR with your changes plus those to the tests: #712

PR #712, opened 2021-02-01
Update matchmaker learning popup message and tests

cleborys:
See #711

### DIFF
--- a/server/ladder_service.py
+++ b/server/ladder_service.py
@@ -285,8 +285,11 @@ class LadderService(Service):
                     "command": "notice",
                     "style": "info",
                     "text": (
-                        "The system is still learning you.<b><br><br>The "
-                        f"learning phase is {progress:.0f}% complete<b>"
+                        "The matchmaking system is calibrating your skill level; "
+                        "your first few games may be slightly more imbalanced "
+                        "as the system attempts to learn your capability as a player."
+                        "<b><br><br>"
+                        f"The learning phase is {progress:.0f}% completed<b>"
                     )
                 })
 
--- a/server/ladder_service.py
+++ b/server/ladder_service.py
@@ -288,8 +288,8 @@ class LadderService(Service):
                         "The matchmaking system is calibrating your skill level; "
                         "your first few games may be slightly more imbalanced "
                         "as the system attempts to learn your capability as a player."
-                        "<b><br><br>"
-                        f"The learning phase is {progress:.0f}% completed<b>"
+                        "<br><br>"
+                        f"<b>The learning phase is {progress:.0f}% completed</b>"
                     )
                 })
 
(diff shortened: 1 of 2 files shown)


tests/unit_tests/test_ladder_service.py
```diff
@@ -827,14 +827,10 @@ async def test_write_rating_progress_message_2(
...
-    player.write_message.assert_called_once_with({
-        "command": "notice",
-        "style": "info",
-        "text": (
-            "The system is still learning you.<b><br><br>"
-            "The learning phase is 40% complete<b>"
-        )
-    })
+    player.write_message.assert_called_once()
+    assert player.write_message.call_args[0][0].get("command") == "notice"
```
cleborys, line 831: This didn't match what i expected from the documentation for a non-coroutine mock:
https://docs.python.org/3/library/unittest.mock.html#unittest.mock.Mock.call_args

In particular `player.write_message.call_args.args` and `player.write_message.call_args.kwargs` are empty. So I had to experiment a bit and this is what turned out to work, but there might be a slicker way to access it.

server/ladder_service.py
```diff
@@ -285,8 +285,11 @@ def write_rating_progress(self, player: Player, rating_type: str) -> None:
                     "command": "notice",
                     "style": "info",
                     "text": (
-                        "The system is still learning you.<b><br><br>The "
-                        f"learning phase is {progress:.0f}% complete<b>"
+                        "The matchmaking system is calibrating your skill level; "
+                        "your first few games may be slightly more imbalanced "
+                        "as the system attempts to learn your capability as a player."
+                        "<b><br><br>"
+                        f"The learning phase is {progress:.0f}% completed<b>"
```
Askaholic, line 292: What is that `<b>` tag for? If it's for making the text bold, shouldn't it come right before the text, and then shouldn't the closing tag be `</b>`?
```suggestion
                        "<br><br>"
                        f"<b>The learning phase is {progress:.0f}% completed</b>"
```

tests/unit_tests/test_ladder_service.py
Askaholic: Why not keep `assert_called_once_with`?. 

I think call_args is a list of NamedTuple's or something like that. `call_args[0]` gives you the args for the first call and `call_args[0][0]` gives you the first arg of the first call.

Askaholic (commented):
Ugh, that flaky test really needs to be fixed.

tests/unit_tests/test_ladder_service.py
cleborys: I thought for `assert_called_once_with` you have to provide the full argument list and that's what I wanted to change here so the test doesn't depend on the exact wording of the message anymore.

Askaholic: Right

Askaholic (approved):
Other than the obvious client issue, it looks good!

image

PR #718, opened 2021-02-06
Issue/#687 Fix coop leaderboards again

### LINKED ISSUE #687: Coop leaderboards are not updated anymore
*Requires Investigation*

This was reported on discord.

**Archsimkat** Nov. 5, 2020
> I have issues with coop times not being saved. None of my coop times have been saved for the last couple of months. Is that functionality supposed to be working?

> I haven't seen any new times from any other people. The leaderboards look exactly the same as it has. I've personally played a dozen or so games with 1/2/3 players and none of them have been recorded.

Askaholic:
Seems like these two things were broken:
- FA reports `Expansion` values as 'true' or 'false' now instead of 1 or 0
- Game ending code triggers before `OperationComplete`, and since coop games don't report any results they were getting marked as `UNKNOWN_RESULT`. This wasn't an issue before because they used to be marked as unknown only after having been inserted into the leaderboard.

If you have better ideas for how to solve point 2, let me hear them. Right now I'm overriding `on_game_end` in `CoopGame`, to remove the game result checking, but that leaves a bunch of other code duplicated.

Closes #687

### DIFF
--- a/server/games/coop.py
+++ b/server/games/coop.py
@@ -34,3 +34,9 @@ class CoopGame(Game):
             "Expansion": ("true", ValidityState.EXPANSION_DISABLED),
         }
         await self._validate_game_options(valid_options)
+
+    async def process_game_results(self):
+        """
+        When a coop game ends, we don't expect there to be any game results.
+        """
+        pass
--- a/server/games/coop.py
+++ b/server/games/coop.py
@@ -1,6 +1,7 @@
 from server.abc.base_game import InitMode
 
-from .game import Game, GameType, ValidityState, Victory
+from .game import Game
+from .typedefs import FA, GameType, ValidityState, Victory
 
 
 class CoopGame(Game):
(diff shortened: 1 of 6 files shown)


server/games/coop.py
```diff
@@ -31,6 +33,35 @@ async def validate_game_mode_settings(self):
             "TeamSpawn": ("fixed", ValidityState.SPAWN_NOT_FIXED),
             "RevealedCivilians": ("No", ValidityState.CIVILIANS_REVEALED),
             "Difficulty": (3, ValidityState.WRONG_DIFFICULTY),
-            "Expansion": (1, ValidityState.EXPANSION_DISABLED),
+            "Expansion": ("true", ValidityState.EXPANSION_DISABLED),
```
cleborys, line 36: It would probably be nice if `1`, `"true"`, and `"True"` would all be accepted.

server/games/coop.py
```diff
@@ -31,6 +33,35 @@ async def validate_game_mode_settings(self):
...
         }
         await self._validate_game_options(valid_options)
+
+    async def on_game_end(self):
+        """
+        When a coop game ends, we don't expect there to be any game results.
+
+        TODO: Refactor this duplication
```
cleborys, line 44: I'm guessing the offending lines were `if not self._results: await self.mark_invalid(ValidityState.UNKNOWN_RESULT)` in `Game.on_game_end`?
Maybe that check can be moved into `Game.resolve_game_results` (which can already modify the validity state) and then you override `resolve_game_results` as no-op instead?

Maybe everything in `Game.on_game_end` pertaining to results should be moved into its own method and overridden here...

cleborys (approved):

server/games/coop.py
Askaholic: Done. I made a generic boolean comparison that will handle true/false, yes/no, on/off, 1/0.

server/games/game.py
```diff
@@ -614,11 +618,11 @@ async def validate_game_settings(self):
             await self.mark_invalid(ValidityState.FFA_NOT_RANKED)
             return
         valid_options = {
-            "AIReplacement": ("Off", ValidityState.HAS_AI_PLAYERS),
```
cleborys, line 617: Oh wow I didn't realize that there were even more words for true/false used already 😅

cleborys (approved):

PR #720, opened 2021-02-13
Fix pop timer crash

Askaholic:
It was not directly the fault of the pop timer, but rather an exception that could sometimes occur during random newbie matching. `CombinedSearch` objects would return a list for `failed_matching_attempts` instead of an integer. This has been fixed and I made sure to add an integration test that covers newbie matching with two solo queue-ers who were placed together by the team creation code. I also wrapped the pop timer loop contents in a try block, since this sort of error should really only be failing the current matching attempt, and not crashing the entire timer loop.

### DIFF
--- a/server/matchmaker/search.py
+++ b/server/matchmaker/search.py
@@ -271,9 +271,8 @@ class CombinedSearch(Search):
         return list(itertools.chain(*[s.raw_ratings for s in self.searches]))
 
     @property
-    def failed_matching_attempts(self) -> List[int]:
-        """Used for logging so returning a different type here is fine"""
-        return [search.failed_matching_attempts for search in self.searches]
+    def failed_matching_attempts(self) -> int:
+        return max(search.failed_matching_attempts for search in self.searches)
 
     def register_failed_matching_attempt(self):
         for search in self.searches:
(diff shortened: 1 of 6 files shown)


server/matchmaker/search.py
```diff
@@ -271,9 +271,8 @@ def raw_ratings(self):
         return list(itertools.chain(*[s.raw_ratings for s in self.searches]))
 
     @property
-    def failed_matching_attempts(self) -> List[int]:
-        """Used for logging so returning a different type here is fine"""
-        return [search.failed_matching_attempts for search in self.searches]
+    def failed_matching_attempts(self) -> int:
```
cleborys, line 274: may `nbr_of_failed_matching_attempts` is more descriptive (and would have made this more obvious).

cleborys (approved):

server/matchmaker/search.py
Askaholic: I don't think the naming is a problem, but rather that returning a list here was just a bad idea on my part :P. At the time it wasn't used and I wasn't sure what to return here. I'm gonna add it to the unit tests now.

0x647262 (approved):
LGTM!

PR #722, opened 2021-02-19
Issue/#721 non ascii in matchmaker game names

### LINKED ISSUE #721: Non-ascii clan name can cause tmm match to fail
If both players on a team are members of a clan with non-ascii characters in the clan tag, then the game name that gets created will contain non-ascii characters and will be rejected, causing the match to be cancelled.

```
INFO     Feb 18  21:03:23 Search                         Matched Search(tmm_2v2, [Player(Shinobu, 44898, (2231.63, 96.4996)), Player(FtXCommando, 62213, (2022.73, 121.69))], threshold=0.68, expansion=0.05) with Search(tmm_2v2, [Player(TURBO2, 89695, (2353.8778029868736, 109.35106982969995)), Player(Espeperant, 189718, (2116.247376374932, 88.8876522634832))], threshold=0.74, expansion=0.0)
INFO     Feb 18  21:03:23 Search                         Matched Search(tmm_2v2, [Player(TURBO2, 89695, (2353.8778029868736, 109.35106982969995)), Player(Espeperant, 189718, (2116.247376374932, 88.8876522634832))], threshold=0.74, expansion=0.0) with Search(tmm_2v2, [Player(Shinobu, 44898, (2231.63, 96.4996)), Player(FtXCommando, 62213, (2022.73, 121.69))], threshold=0.68, expansion=0.05)
INFO     Feb 18  21:03:23 PopTimer                       Next tmm2v2 wave happening in 179s
ERROR    Feb 18  21:03:23 LadderService                  Failed to start ladder game!
  File "/code/server/ladder_service.py", line 379, in start_game
    max_players=len(all_players)
  File "/code/server/game_service.py", line 166, in create_game
    game = game_class(**game_args)
  File "/code/server/games/ladder_game.py", line 28, in __init__
    super().__init__(id_, *args, **new_kwargs)
  File "/code/server/games/game.py", line 84, in __init__
    self.name = name
  File "/code/server/games/game.py", line 141, in name
    raise ValueError("Name must be ascii!")
ValueError: Name must be ascii!
```

Ftx and Shinobu are both in a clan with the tag `J¢p`.

I think the choices are:
1. Allow a bypass to the ascii check for internal use.
2. Only use clan tag if the clan tag is ascii
3. Remove the ascii check entirely

I think option 1 actually does make sense. If we're already allowing those characters in the clan tag, there's no harm in allowing them in the game name. The ascii enforcement is mostly there to prevent custom lobbies having names in weird encodings that don't render properly on other systems.

Askaholic:
Allow auto generated game names to include non-ascii characters. This lets us keep the team naming based on clans, since some clan tags include non-ascii characters.

I modified a few of the `start_game` tests to use hypothesis for fuzzing player attributes like clan tag, name, rating and number of games. See the commit descriptions.

Closes #721

### DIFF
--- a/tests/unit_tests/test_ladder_service.py
+++ b/tests/unit_tests/test_ladder_service.py
@@ -6,7 +6,7 @@ from asynctest import CoroutineMock, create_autospec, exhaust_callbacks
 from hypothesis import given, settings
 from hypothesis import strategies as st
 
-from server import LadderService
+from server import LadderService, LobbyConnection
 from server.db.models import matchmaker_queue, matchmaker_queue_map_pool
 from server.games import LadderGame
 from server.ladder_service import game_name
@@ -96,8 +96,8 @@ async def test_load_from_database_new_data(ladder_service, database):
 
 
 @given(
-    player1=st_players("p1", player_id=1),
-    player2=st_players("p2", player_id=2)
+    player1=st_players("p1", player_id=1, lobby_connection_spec="mock"),
+    player2=st_players("p2", player_id=2, lobby_connection_spec="mock")
 )
 @settings(deadline=300)
 @autocontext("ladder_and_game_service_context", "monkeypatch_context")
@@ -114,8 +114,9 @@ async def test_start_game_1v1(
     monkeypatch.setattr(LadderGame, "wait_launched", CoroutineMock())
     monkeypatch.setattr(LadderGame, "timeout_game", CoroutineMock())
     for player in (player1, player2):
-        player.__owned_lobby_connection = mock.Mock(launch_game=CoroutineMock())
-        player.lobby_connection = player.__owned_lobby_connection
+        player.lobby_connection.launch_game = CoroutineMock(
+            spec=LobbyConnection.launch_game
+        )
 
     await ladder_service.start_game([player1], [player2], queue)
 
@@ -137,8 +138,8 @@ async def test_start_game_timeout(
     player_factory,
 ):
     queue = ladder_service.queues["ladder1v1"]
-    p1 = player_factory("Dostya", player_id=1, with_lobby_connection=True)
-    p2 = player_factory("Rhiza", player_id=2, with_lobby_connection=True)
+    p1 = player_factory("Dostya", player_id=1, lobby_connection_spec="auto")
+    p2 = player_factory("Rhiza", player_id=2, lobby_connection_spec="auto")
 
     await ladder_service.start_game([p1], [p2], queue)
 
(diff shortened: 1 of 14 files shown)


tests/unit_tests/test_ladder_service.py
```diff
@@ -139,28 +151,44 @@ async def test_start_game_timeout(
     assert p2.state is PlayerState.IDLE
 
 
+@pytest.mark.filterwarnings("ignore:.*uses the '.*' fixture")
```
cleborys, line 154: What is this about?

tests/unit_tests/test_ladder_service.py
```diff
@@ -93,25 +95,35 @@ async def test_load_from_database_new_data(ladder_service, database):
...
     queue = ladder_service.queues["ladder1v1"]
-    p1 = player_factory("Dostya", player_id=1, with_lobby_connection=True)
-    p2 = player_factory("Rhiza", player_id=2, with_lobby_connection=True)
 
     monkeypatch.setattr(LadderGame, "wait_hosted", CoroutineMock())
     monkeypatch.setattr(LadderGame, "wait_launched", CoroutineMock())
-    await ladder_service.start_game([p1], [p2], queue)
+    monkeypatch.setattr(LadderGame, "timeout_game", CoroutineMock())
+    for player in (player1, player2):
+        player.__owned_lobby_connection = mock.Mock(launch_game=CoroutineMock())
```
cleborys, line 117: This is different from the autospec mock in`test/conftest.py`. Would that one not work? Maybe it would make sense to allow to set `with_lobby_connection` also for players drawn from `st_players`?

tests/unit_tests/test_ladder_service.py
```diff
@@ -139,28 +151,44 @@ async def test_start_game_timeout(
...
-    p2 = player_factory("Rhiza", player_id=2, with_lobby_connection=True)
-    p3 = player_factory("QAI", player_id=3, with_lobby_connection=True)
-    p4 = player_factory("Hall", player_id=4, with_lobby_connection=True)
 
     monkeypatch.setattr(LadderGame, "wait_hosted", CoroutineMock())
     monkeypatch.setattr(LadderGame, "wait_launched", CoroutineMock())
-    await ladder_service.start_game([p1, p3], [p2, p4], queue)
+    monkeypatch.setattr(LadderGame, "timeout_game", CoroutineMock())
+    for player in (player1, player2, player3, player4):
+        player.__owned_lobby_connection = mock.Mock(launch_game=CoroutineMock())
```
cleborys, line 177: See above

cleborys (approved):
Very nice to have fuzzing in more unit tests

tests/unit_tests/test_ladder_service.py
Askaholic: Hypothesis raises a warning when you use function scoped fixtures because they will only be initialized once, and not every time hypothesis queries the test function. You have to make sure that your test function will handle cleaning up its own state changes if it makes any. In our case, it doesn't make any except for mocking certain function calls, which is perfectly fine to do over and over without resetting the fixtures.

tests/unit_tests/test_ladder_service.py
Askaholic: It would work and you could use it already since kwargs are forwarded. The problem is that autospecing is expensive and tends to cause the test to exceed the default execution deadline of 200ms. 

Maybe I should use a spec_set here instead of a full autospec. I think that would be fast and still give most of the benefits of the autospec'd mock.

0x647262 (approved):
LGTM!

PR #606, opened 2020-06-17
Send game_launch to guest only if host hasn't timed out

cleborys:
Counterpart to https://github.com/FAForever/downlords-faf-client/pull/1778

PR #723, opened 2021-02-19
Issue/#606 Rare mysql deadlock

Askaholic:
Can't really test it since it only happens rarely on production and probably only under high load.

Closes #606


p4block:
This may or may not be fixed magically with the move to MariaDB

cleborys (approved):

PR #732, opened 2021-02-28
Add generated map support to ladder_service

Sheikah45:
I went ahead and took a first swing at how supporting generated maps would look like in the server code.
The id for the generated maps is the negative integer of the bytes representing the seed and options. This allows for the map_pool choose maps to still work as intended while removing the possibility of collisions with authored maps in the pool and having a unique id for each generated map.

The current issue is that this would need for the map_version_id to be null in the map_pool_map_version table but that is currently being used as part of a primary key probably need to consult @Brutus5000

### DIFF
--- a/server/db/models.py
+++ b/server/db/models.py
@@ -256,11 +256,11 @@ map_pool = Table(
 
 map_pool_map_version = Table(
     "map_pool_map_version", metadata,
-    Column("id",            Integer,        primary_key=True),
+    Column("id",             Integer,       primary_key=True),
     Column("map_pool_id",    Integer,       ForeignKey("map_pool.id"),      nullable=False),
     Column("map_version_id", Integer,       ForeignKey("map_version.id")),
-    Column("weight",        Integer,        nullable=False),
-    Column("map_params",    Text),
+    Column("weight",         Integer,       nullable=False),
+    Column("map_params",     Text),
 )
 
 map_version = Table(
--- a/server/db/models.py
+++ b/server/db/models.py
@@ -257,7 +257,7 @@ map_pool = Table(
 map_pool_map_version = Table(
     "map_pool_map_version", metadata,
     Column("id",            Integer,        primary_key=True),
-    Column("map_pool_id",   Integer,        ForeignKey("map_pool.id"),      nullable=False),
+    Column("map_pool_id",    Integer,       ForeignKey("map_pool.id"),      nullable=False),
     Column("map_version_id", Integer,       ForeignKey("map_version.id")),
     Column("weight",        Integer,        nullable=False),
     Column("map_params",    Text),
--- a/server/db/models.py
+++ b/server/db/models.py
@@ -256,11 +256,11 @@ map_pool = Table(
 
 map_pool_map_version = Table(
     "map_pool_map_version", metadata,
-    Column("id",                Integer,    primary_key=True),
-    Column("map_pool_id",       Integer,    ForeignKey("map_pool.id"),     nullable=False),
-    Column("map_version_id",    Integer,    ForeignKey("map_version.id"),  nullable=True),
-    Column("weight",            Integer,    nullable=False),
-    Column("map_params",        Text,       nullable=True),
+    Column("id", Integer, primary_key=True),
+    Column("map_pool_id", Integer, ForeignKey("map_pool.id"), nullable=False),
+    Column("map_version_id", Integer, ForeignKey("map_version.id")),
+    Column("weight", Integer, nullable=False),
+    Column("map_params", Text),
 )
 
 map_version = Table(
--- a/server/db/models.py
+++ b/server/db/models.py
@@ -256,11 +256,11 @@ map_pool = Table(
 
 map_pool_map_version = Table(
     "map_pool_map_version", metadata,
-    Column("id", Integer, primary_key=True),
-    Column("map_pool_id", Integer, ForeignKey("map_pool.id"), nullable=False),
-    Column("map_version_id", Integer, ForeignKey("map_version.id")),
-    Column("weight", Integer, nullable=False),
-    Column("map_params", Text),
+    Column("id",            Integer,        primary_key=True),
+    Column("map_pool_id",   Integer,        ForeignKey("map_pool.id"),      nullable=False),
+    Column("map_version_id", Integer,       ForeignKey("map_version.id")),
+    Column("weight",        Integer,        nullable=False),
+    Column("map_params",    Text),
 )
 
 map_version = Table(
--- a/server/db/models.py
+++ b/server/db/models.py
@@ -256,8 +256,11 @@ map_pool = Table(
 
 map_pool_map_version = Table(
     "map_pool_map_version", metadata,
-    Column("map_pool_id",       Integer, ForeignKey("map_pool.id"),     nullable=False),
-    Column("map_version_id",    Integer, ForeignKey("map_version.id"),  nullable=False),
+    Column("id",                Integer,    primary_key=True),
+    Column("map_pool_id",       Integer,    ForeignKey("map_pool.id"),     nullable=False),
+    Column("map_version_id",    Integer,    ForeignKey("map_version.id"),  nullable=True),
+    Column("weight",            Integer,    nullable=False),
+    Column("map_params",        Text,       nullable=True),
 )
 
 map_version = Table(
--- a/server/db/models.py
+++ b/server/db/models.py
@@ -261,6 +261,8 @@ map_pool_map_version = Table(
     Column("map_version_id", Integer,       ForeignKey("map_version.id")),
     Column("weight",         Integer,       nullable=False),
     Column("map_params",     Text),
+    Column("create_time",    TIMESTAMP,     nullable=False),
+    Column("update_time",    TIMESTAMP,     nullable=False)
 )
 
 map_version = Table(
--- a/server/ladder_service.py
+++ b/server/ladder_service.py
@@ -135,8 +135,11 @@ class LadderService(Service):
                     if map_type == "neroxis":
                         map_list.append(NeroxisGeneratedMap.of(params=params, weight=row.weight))
                     else:
-                        self._logger.warning("Unsupported map type %s with parameters %s",
-                                             map_type, row.map_params)
+                        self._logger.warning(
+                            "Unsupported map type %s with parameters %s",
+                             map_type,
+                             row.map_params
+                        )
 
                 except Exception:
                     self._logger.warning(
--- a/server/ladder_service.py
+++ b/server/ladder_service.py
@@ -134,6 +134,9 @@ class LadderService(Service):
                     map_type = params["type"]
                     if map_type == "neroxis":
                         map_list.append(NeroxisGeneratedMap.of(params=params, weight=row.weight))
+                    else:
+                        self._logger.warning("Unsupported map type %s with parameters %s",
+                                             map_type, row.map_params)
 
                 except Exception:
                     self._logger.warning(
--- a/server/ladder_service.py
+++ b/server/ladder_service.py
@@ -1,5 +1,5 @@
 import asyncio
-import itertools
+import json
 import random
 import re
 from collections import defaultdict
@@ -32,8 +32,7 @@ from .game_service import GameService
 from .games import LadderGame
 from .matchmaker import MapPool, MatchmakerQueue, OnMatchedCallback, Search
 from .players import Player, PlayerState
-from .rating import RatingType
-from .types import GameLaunchOptions, Map
+from .types import GameLaunchOptions, Map, NeroxisGeneratedMap
 
 
 @with_logger
@@ -107,6 +106,8 @@ class LadderService(Service):
             select([
                 map_pool.c.id,
                 map_pool.c.name,
+                map_pool_map_version.c.weight,
+                map_pool_map_version.c.map_params,
                 map_version.c.map_id,
                 map_version.c.filename,
                 t_map.c.display_name
@@ -125,8 +126,23 @@ class LadderService(Service):
             _, map_list = map_pool_maps[id_]
             if row.map_id is not None:
                 map_list.append(
-                    Map(row.map_id, row.display_name, row.filename)
+                    Map(row.map_id, row.display_name, row.filename, row.weight)
                 )
+            elif row.map_params is not None:
+                try:
+                    params = json.loads(row.map_params)
+                    map_type = params["type"]
+                    if map_type == "neroxis":
+                        map_list.append(NeroxisGeneratedMap.of(params=params, weight=row.weight))
+
+                except Exception:
+                    self._logger.warning(
+                        "Failed to load map in map pool %d"
+                        "parameters specified as %s",
+                        row.id,
+                        row.map_params,
+                        exc_info=True
+                    )
 
         return map_pool_maps
 
@@ -367,7 +383,7 @@ class LadderService(Service):
             pool = queue.get_map_pool_for_rating(rating)
             if not pool:
                 raise RuntimeError(f"No map pool available for rating {rating}!")
-            map_id, map_name, map_path = pool.choose_map(played_map_ids)
+            _, _, map_path, _ = pool.choose_map(played_map_ids)
 
             game = self.game_service.create_game(
                 game_class=LadderGame,
@@ -410,6 +426,7 @@ class LadderService(Service):
             mapname = re.match("maps/(.+).zip", map_path).group(1)
             # FIXME: Database filenames contain the maps/ prefix and .zip suffix.
             # Really in the future, just send a better description
+
             self._logger.debug("Starting ladder game: %s", game)
             # Options shared by all players
             options = GameLaunchOptions(
@@ -428,7 +445,7 @@ class LadderService(Service):
                 game, is_host=True, options=game_options(host)
             )
             try:
-                await game.wait_hosted(30)
+                await game.wait_hosted(60)
             finally:
                 # TODO: Once the client supports `match_cancelled`, don't
                 # send `launch_game` to the client if the host timed out. Until
@@ -443,7 +460,7 @@ class LadderService(Service):
                     for guest in all_guests
                     if guest.lobby_connection is not None
                 ])
-            await game.wait_launched(30 + 10 * len(all_guests))
+            await game.wait_launched(60 + 10 * len(all_guests))
             self._logger.debug("Ladder game launched successfully")
         except Exception:
             if game:
--- a/server/ladder_service.py
+++ b/server/ladder_service.py
@@ -133,7 +133,7 @@ class LadderService(Service):
                     params = json.loads(row.map_params)
                     map_type = params["type"]
                     if map_type == "neroxis":
-                        map_list.append(NeroxisGeneratedMap.of(params=params, weight=row.weight))
+                        map_list.append(NeroxisGeneratedMap.of(params, row.weight))
                     else:
                         self._logger.warning(
                             "Unsupported map type %s in pool %s",
--- a/server/ladder_service.py
+++ b/server/ladder_service.py
@@ -136,9 +136,9 @@ class LadderService(Service):
                         map_list.append(NeroxisGeneratedMap.of(params=params, weight=row.weight))
                     else:
                         self._logger.warning(
-                            "Unsupported map type %s with parameters %s",
+                            "Unsupported map type %s in pool %s",
                              map_type,
-                             row.map_params
+                             row.id
                         )
 
                 except Exception:
--- a/server/ladder_service.py
+++ b/server/ladder_service.py
@@ -137,7 +137,7 @@ class LadderService(Service):
 
                 except Exception:
                     self._logger.warning(
-                        "Failed to load map in map pool %d"
+                        "Failed to load map in map pool %d "
                         "parameters specified as %s",
                         row.id,
                         row.map_params,
--- a/server/matchmaker/map_pool.py
+++ b/server/matchmaker/map_pool.py
@@ -1,9 +1,9 @@
 import random
 from collections import Counter
-from typing import Iterable
+from typing import Iterable, Union
 
 from ..decorators import with_logger
-from ..types import Map
+from ..types import Map, NeroxisGeneratedMap
 
 
 @with_logger
@@ -12,7 +12,7 @@ class MapPool(object):
         self,
         map_pool_id: int,
         name: str,
-        maps: Iterable[Map] = ()
+        maps: Iterable[Union[Map, NeroxisGeneratedMap]] = ()
     ):
         self.id = map_pool_id
         self.name = name
@@ -39,14 +39,21 @@ class MapPool(object):
         counter.update(id_ for id_ in played_map_ids if id_ in self.maps)
 
         least_common = counter.most_common()[::-1]
-        least_count = least_common[0][1]
+        least_count = 1
+        for map_count in least_common:
+            if isinstance(self.maps[map_count[0]], Map):
+                least_count = map_count[1]
+                break
+
         # Trim off the maps with higher play counts
         for i, (_, count) in enumerate(least_common):
-            if count != least_count:
+            if count > least_count:
                 least_common = least_common[:i]
                 break
 
-        return self.maps[random.choice(least_common)[0]]
+        weights = [self.maps[id_].weight for id_, _ in least_common]
+
+        return self.maps[random.choices(least_common, weights=weights, k=1)[0][0]].get_map()
 
     def __repr__(self):
         return f"MapPool({self.id}, {self.name}, {list(self.maps.values())})"
--- a/server/types.py
+++ b/server/types.py
@@ -45,6 +45,8 @@ class NeroxisGeneratedMap(NamedTuple):
 
     @classmethod
     def of(cls, params: dict, weight: int = 1):
+        assert params["type"] == "neroxis"
+
         map_size_pixels = int(params["size"])
 
         if map_size_pixels <= 0:
--- a/server/types.py
+++ b/server/types.py
@@ -1,3 +1,5 @@
+import base64
+import random
 from typing import NamedTuple, Optional
 
 
@@ -27,3 +29,54 @@ class Map(NamedTuple):
     id: int
     name: str
     path: str
+    weight: int = 1
+
+    def get_map(self) -> "Map":
+        return self
+
+
+# FIXME: Use typing.Protocol in python3.8 to subtype with MAP
+class NeroxisGeneratedMap(NamedTuple):
+    id: int
+    version: str
+    spawns: int
+    map_size_pixels: int
+    weight: int = 1
+
+    @classmethod
+    def of(cls, params: dict, weight: int = 1):
+        map_size_pixels = int(params["size"])
+
+        if map_size_pixels <= 0:
+            raise Exception("Map size is zero or negative")
+
+        if map_size_pixels & (map_size_pixels - 1) != 0:
+            raise Exception("Map size is not a power of 2")
+
+        spawns = int(params["spawns"])
+        if spawns % 2 != 0:
+            raise Exception("spawns is not a multiple of 2")
+
+        version = params["version"]
+        return NeroxisGeneratedMap(
+            -int.from_bytes(bytes(f'{version}_{spawns}_{map_size_pixels}', encoding="ascii"), 'big'),
+            version,
+            spawns,
+            map_size_pixels,
+            weight
+        )
+
+    def get_map(self) -> Map:
+        """
+            Generate a map name based on the version and parameters. If invalid parameters are specified
+            hand back None
+        """
+        seed_bytes = random.getrandbits(64).to_bytes(8, 'big')
+        size_byte = (self.map_size_pixels // 64).to_bytes(1, 'big')
+        spawn_byte = self.spawns.to_bytes(1, 'big')
+        option_bytes = spawn_byte + size_byte
+        seed_str = base64.b32encode(seed_bytes).decode("ascii").replace("=", "").lower()
+        option_str = base64.b32encode(option_bytes).decode("ascii").replace("=", "").lower()
+        map_name = f'neroxis_map_generator_{self.version}_{seed_str}_{option_str}'
+        map_path = f'maps/{map_name}.zip'
+        return Map(self.id, map_name, map_path, self.weight)
--- a/server/types.py
+++ b/server/types.py
@@ -61,7 +61,7 @@ class NeroxisGeneratedMap(NamedTuple):
 
         version = params["version"]
         return NeroxisGeneratedMap(
-            -int.from_bytes(bytes(f'{version}_{spawns}_{map_size_pixels}', encoding="ascii"), 'big'),
+            -int.from_bytes(bytes(f"{version}_{spawns}_{map_size_pixels}", encoding="ascii"), "big"),
             version,
             spawns,
             map_size_pixels,
@@ -73,12 +73,12 @@ class NeroxisGeneratedMap(NamedTuple):
             Generate a map name based on the version and parameters. If invalid parameters are specified
             hand back None
         """
-        seed_bytes = random.getrandbits(64).to_bytes(8, 'big')
-        size_byte = (self.map_size_pixels // 64).to_bytes(1, 'big')
-        spawn_byte = self.spawns.to_bytes(1, 'big')
+        seed_bytes = random.getrandbits(64).to_bytes(8, "big")
+        size_byte = (self.map_size_pixels // 64).to_bytes(1, "big")
+        spawn_byte = self.spawns.to_bytes(1, "big")
         option_bytes = spawn_byte + size_byte
         seed_str = base64.b32encode(seed_bytes).decode("ascii").replace("=", "").lower()
         option_str = base64.b32encode(option_bytes).decode("ascii").replace("=", "").lower()
-        map_name = f'neroxis_map_generator_{self.version}_{seed_str}_{option_str}'
-        map_path = f'maps/{map_name}.zip'
+        map_name = f"neroxis_map_generator_{self.version}_{seed_str}_{option_str}"
+        map_path = f"maps/{map_name}.zip"
         return Map(self.id, map_name, map_path, self.weight)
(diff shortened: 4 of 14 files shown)


Askaholic (commented):
Maybe something like instead of having `MapPool.maps` be a dict of `Map`s, make a new type that returns a `Map` tuple when called.


```python

class Map(NamedTuple):
    # Map fields 

    def __call__(self) -> "Map":
        return self

class NeroxisGeneratedMap(NamedTuple):
    # Generator parameters here

    def __call__(self) -> Map:
        return generated_map(self.params)
```

Then modify the last line of `choose_map` to call the chosen map:
```python
return self.maps[random.choice(least_common)[0]]()
```

So the map pool can consist of both `Map` and `NeroxisGeneratedMap` objects that will resolve to a specific `Map` tuple when chosen.

It could also be a named method like `get_map` instead of using `__call__`.

Sheikah45:
Yeah that would make sense I guess this version tried to be as little invasive as possible which is probably not needed.
Also should the function for generating the map names have its own module in your opinion?

Askaholic:
I’d probably implement it as a method of `NeroxisGeneratedMap` an probably keep that in the same place as `Map`.

I’m also trying to keep it more general than just adding support for generated maps. I’m thinking of it more as a general map source that comes from outside of the database. This could then also be used for coop maps for example since those aren’t in the map_version table either.

Sheikah45:
Ah in that case then it may be better to just have the map_pool_map_version have the json string describing it as that would offer the most flexibility without the need for all the added columns in the future

Askaholic:
Kindof except that json will waste a lot of space storing the same key's and all the quotes and stuff.

Sheikah45:
Somewhat yeah but it will also allow the greatest flexibility especially since the map_pool_map_version table is most likely going to be a pretty dynamic table so I wonder if the json is a good tradeoff for the simplicity of maintaining it

server/types.py
```diff
@@ -27,3 +29,31 @@ class Map(NamedTuple):
...
+    spawns: int
+    map_size_pixels: int
+    weight: int
+
+    def get_map(self) -> Map:
+        """
+            Generate a map name based on the version and parameters. If invalid parameters are specified
+            hand back None
+        """
+        seed_bytes = random.getrandbits(64).to_bytes(8, 'little')
```
Askaholic, line 50: Oof. Sadly `randbytes` isn't added until 3.9.

server/types.py
```diff
@@ -27,3 +29,31 @@ class Map(NamedTuple):
...
+        size_byte = int(self.map_size_pixels / 64).to_bytes(1, 'little')
```
Askaholic, line 51:
```suggestion
        size_byte = (self.map_size_pixels // 64).to_bytes(1, 'little')
```

server/types.py
```diff
@@ -27,3 +29,31 @@ class Map(NamedTuple):
...
+        spawn_byte = self.spawns.to_bytes(1, 'little')
+        option_bytes = spawn_byte + size_byte
+        seed_str = base64.b32encode(seed_bytes).decode("ascii").replace("=", "").lower()
+        option_str = base64.b32encode(spawn_byte + size_byte).decode("ascii").replace("=", "").lower()
```
Askaholic, line 55: Already have `option_bytes` and b32encode is already guaranteed to return lowercase.
```suggestion
        option_str = base64.b32encode(option_bytes).decode("ascii").replace("=", "")
```

tests/unit_tests/test_ladder_service.py
```diff
@@ -52,17 +52,17 @@ async def test_load_from_database(ladder_service, queue_factory):
         assert queue.name == "ladder1v1"
         assert len(queue.map_pools) == 3
         assert list(queue.map_pools[1][0].maps.values()) == [
-            Map(id=15, name="SCMP_015", path="maps/scmp_015.v0003.zip"),
+            Map(id=15, name="SCMP_015", path="maps/scmp_015.v0003.zip", weight=1),
```
Askaholic, line 55: When you add the default you won't need to change it in the tests

server/ladder_service.py
```diff
@@ -125,8 +131,22 @@ async def fetch_map_pools(self, conn) -> Dict[int, Tuple[str, List[Map]]]:
...
+                    if map_type == "neroxis":
+                        if "size" in param_dict and "spawns" in param_dict and "version" in param_dict:
+                            map_size_pixels = param_dict["size"]
+                            spawns = param_dict["spawns"]
+                            version = param_dict["version"]
+                            if isinstance(map_size_pixels, int) and isinstance(spawns, int):
+                                gen_id = version + str(spawns) + str(map_size_pixels)
+                                map_list.append(
+                                    NeroxisGeneratedMap(gen_id, version, spawns, map_size_pixels, row.weight)
+                                )
```
Askaholic, line 149: 10 levels of indentation is a little too much IMO. I would say don't do all this error checking and instead wrap the whole thing in a try block
```suggestion
                try:
                    params = json.loads(row.other_map_params)
                    map_type = params["type"]
                    if map_type != "neroxis":
                        raise Exception("Unsupported map type")
                    map_size_pixels = param_dict["size"]
                    spawns = param_dict["spawns"]
                    version = param_dict["version"]
                    gen_id = version + str(spawns) + str(map_size_pixels)
                    map_list.append(
                        NeroxisGeneratedMap(gen_id, version, spawns, map_size_pixels, row.weight)
                    )
                except Exception:
                    self.logger.warning("Failed to load map [other relevant debug info]", exc_info=True)
```

server/ladder_service.py
```diff
@@ -125,8 +131,22 @@ async def fetch_map_pools(self, conn) -> Dict[int, Tuple[str, List[Map]]]:
...
+                param_dict = json.loads(row.other_map_params)
+                if "type" in param_dict:
+                    map_type = param_dict["type"]
+                    if map_type == "neroxis":
+                        if "size" in param_dict and "spawns" in param_dict and "version" in param_dict:
+                            map_size_pixels = param_dict["size"]
+                            spawns = param_dict["spawns"]
+                            version = param_dict["version"]
+                            if isinstance(map_size_pixels, int) and isinstance(spawns, int):
+                                gen_id = version + str(spawns) + str(map_size_pixels)
```
Askaholic, line 146: Maybe override the constructor to do this automatically. Also a python tip, strings are immutable so it's better to format than add when you have more than 2 strings, otherwise it will lead to excessive copying:
```suggestion
                                gen_id = f"{version}{spawns}{map_size_pixels}"
```

Askaholic (changes requested):
I think the idea is good so now I'm commenting on python/code style things.

Also it will need some tests, but we'll need a new db release before those will be able to run.

server/types.py
Sheikah45: yeah it made me sad too

server/types.py
Sheikah45: good catch on the option bytes. Although when I tested it locally on my machine b32encode returned on uppercase. Is there a configuration somewhere else that forces it lower case?

server/matchmaker/map_pool.py
```diff
@@ -12,7 +12,7 @@ def __init__(
         self,
         map_pool_id: int,
         name: str,
-        maps: Iterable[Map] = ()
+        maps: Iterable[Union[Map, NeroxisGeneratedMap]] = ()
```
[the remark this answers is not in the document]
Sheikah45: Yeah I was wondering if it would be good to define an abstract class with id, weight, and get_map but didnt see it done elsewhere

server/ladder_service.py
Sheikah45: heh yeah that would make it easier

server/ladder_service.py
Sheikah45: ah yeah that was just a derp moment of me forgetting about f strings

server/ladder_service.py
Sheikah45: although I am a little concerned about the type checking for the spawns and map_size_pixels as this way it wouldn't error out until the get_map was called

server/ladder_service.py
Sheikah45: And unfortunately you can't overwrite the __init__ of NamedTuples

server/types.py
Askaholic: Hmm, I guess not. My bad.

server/ladder_service.py
Askaholic: Well you can check it and just raise an exception like I did with the map type.

server/ladder_service.py
Askaholic: I guess you could make a class method and just have it call the constructor :P.

server/types.py
Sheikah45: From what I can tell the default for b32encode and b32decode is uppercase

server/types.py
```diff
@@ -27,3 +29,31 @@ class Map(NamedTuple):
     id: int
     name: str
     path: str
+    weight: int
```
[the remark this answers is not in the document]
Sheikah45: makes sense

server/types.py
Sheikah45: got it

tests/unit_tests/test_ladder_service.py
Sheikah45: yep

server/ladder_service.py
Sheikah45: got it

server/db/models.py
```diff
@@ -256,8 +256,10 @@
 
 map_pool_map_version = Table(
     "map_pool_map_version", metadata,
-    Column("map_pool_id",       Integer, ForeignKey("map_pool.id"),     nullable=False),
-    Column("map_version_id",    Integer, ForeignKey("map_version.id"),  nullable=False),
+    Column("map_pool_id",       Integer,    ForeignKey("map_pool.id"),     nullable=False),
+    Column("map_version_id",    Integer,    ForeignKey("map_version.id"),  nullable=True),
+    Column("weight",            Integer,    nullable=False),
+    Column("other_map_params",  Text,      nullable=True),
```
Askaholic, line 262:
```suggestion
    Column("other_map_params",  Text,       nullable=True),
```

server/ladder_service.py
```diff
@@ -125,8 +128,30 @@ async def fetch_map_pools(self, conn) -> Dict[int, Tuple[str, List[Map]]]:
...
+                    params = json.loads(row.other_map_params)
+                    map_type = params["type"]
+                    if map_type != "neroxis":
+                        raise Exception("Unsupported map type")
+
+                    map_size_pixels = int(params["size"])
+                    spawns = int(params["spawns"])
+                    version = params["version"]
+                    map_list.append(
+                        NeroxisGeneratedMap.of(version, spawns, map_size_pixels, row.weight)
```
Askaholic, line 144:
```suggestion
                        NeroxisGeneratedMap.of(
                            version, 
                            spawns, 
                            map_size_pixels, 
                            row.weight
                        )
```

server/types.py
```diff
@@ -27,3 +29,36 @@ class Map(NamedTuple):
...
+class NeroxisGeneratedMap(NamedTuple):
+    id: str
+    version: str
+    spawns: int
+    map_size_pixels: int
+    weight: int = 1
+
+    @classmethod
+    def of(cls, version: str, spawns: int, map_size_pixels: int, weight: int):
+        return NeroxisGeneratedMap(f'{version}_{spawns}_{map_size_pixels}', version, spawns, map_size_pixels, weight)
```
Askaholic, line 48:
```suggestion
        return NeroxisGeneratedMap(
            f'{version}_{spawns}_{map_size_pixels}', 
            version, 
            spawns, 
            map_size_pixels, 
            weight
        )
```

Askaholic (commented):
Looks fine to me. @Brutus5000 what do you think of the database changes?

Now we just need
1. database PR
2. unit tests
3. sort imports

Askaholic:
Look at GitHub actions “Test / isort”

Sheikah45:
I went ahead and configured the database on my local machine to add and run some tests let me know if they look reasonable to you

server/db/models.py
```diff
@@ -256,8 +256,11 @@
 
 map_pool_map_version = Table(
     "map_pool_map_version", metadata,
-    Column("map_pool_id",       Integer, ForeignKey("map_pool.id"),     nullable=False),
-    Column("map_version_id",    Integer, ForeignKey("map_version.id"),  nullable=False),
+    Column("id",                Integer,    primary_key=True),
+    Column("map_pool_id",       Integer,    ForeignKey("map_pool.id"),     nullable=False),
+    Column("map_version_id",    Integer,    ForeignKey("map_version.id"),  nullable=True),
+    Column("weight",            Integer,    nullable=False),
+    Column("map_params",  Text,       nullable=True),
```
Askaholic, line 263: formatting

server/ladder_service.py
```diff
@@ -125,8 +128,44 @@ async def fetch_map_pools(self, conn) -> Dict[int, Tuple[str, List[Map]]]:
             _, map_list = map_pool_maps[id_]
             if row.map_id is not None:
                 map_list.append(
-                    Map(row.map_id, row.display_name, row.filename)
+                    Map(row.map_id, row.display_name, row.filename, row.weight)
                 )
+            elif row.map_params is not None:
+                try:
+                    params = json.loads(row.map_params)
```
Askaholic, line 135: Maybe you want to extract this to a function. It's getting kinda long.

tests/unit_tests/test_map_pool.py
```diff
@@ -31,7 +33,45 @@ def test_choose_map(map_pool_factory):
...
+        Map(1, "some_map", "maps/some_map.v001.zip", 1),
+        Map(2, "some_map", "maps/some_map.v001.zip", 1),
+        Map(3, "some_map", "maps/some_map.v001.zip", 1),
+        Map(4, "CHOOSE_ME", "maps/choose_me.v001.zip", 1000000),
+    ])
+
+    # Make the probability very low that the test passes because we got lucky
+    for _ in range(20):
+        chosen_map = map_pool.choose_map()
+        assert chosen_map == (4, "CHOOSE_ME", "maps/choose_me.v001.zip", 1000000)
```
Askaholic, line 50: Hm, interesting. Kindof a hard thing to test I guess. I'm just looking forward to that one time I get super unlucky and it fails on a release build :P

Askaholic (commented):
I think I would also like one "integration" style test as well. I think there should be some that you can copy in `tests/integration_tests/test_matchmaker.py`.

tests/unit_tests/test_map_pool.py
Sheikah45: Yeah I was considering if I should even have this or make it that m out of n should be chosen

server/ladder_service.py
```diff
@@ -164,6 +169,42 @@ async def fetch_matchmaker_queues(self, conn):
             ))
         return matchmaker_queues
 
+    def parse_map_params(self, map_list: List[Union[Map, NeroxisGeneratedMap]], map_params: str, weight: int):
+        try:
```
Askaholic, line 173: I would put the `try` block around the call, and I'd just return the `NeroxisGeneratedMap` here, and append it to the list where it's called.

tests/integration_tests/test_matchmaker.py
```diff
@@ -40,6 +42,27 @@ async def test_game_launch_message(lobby_server):
     assert msg2["map_position"] == 2
 
 
+@fast_forward(70)
+async def test_game_launch_message_map_generator(lobby_server):
+    size = 512
+    spawns = 2
+    version = "0.0.0"
```
Askaholic, line 49: I don't think you actually need the version and seed_match variables. They're only used in the regex so might as well put them directly there for readability. And I might even do the same with option_str, just figure out what it is and hardcode that. That's how I like to do it in tests anyways.

tests/unit_tests/test_map_pool.py
Askaholic: I think if the probability of it failing is sufficiently small then it will still be ok. Like 10^-6 or something idk.

I guess the probability that this fails is already pretty close to that:
```
(1 - (1000000/1000003)) * 20 = 6e-05
```

server/ladder_service.py
```diff
@@ -410,6 +422,11 @@ def get_player_mean(player):
             mapname = re.match("maps/(.+).zip", map_path).group(1)
             # FIXME: Database filenames contain the maps/ prefix and .zip suffix.
             # Really in the future, just send a better description
+
+            additional_timeout = 0
+            if "neroxis_map_generator" in mapname:
+                additional_timeout += 30
+
```
Askaholic, line 429: IMO, just increase the base timeout to 60.

server/ladder_service.py
```diff
@@ -501,6 +518,32 @@ async def shutdown(self):
             queue.shutdown()
 
 
+def parse_map_params(map_params: str, weight: int) -> NeroxisGeneratedMap:
```
Askaholic, line 521: Could be a class method of `NeroxisGeneratedMap`?

server/matchmaker/map_pool.py
```diff
@@ -39,14 +39,21 @@ def choose_map(self, played_map_ids: Iterable[int] = ()) -> Map:
         counter.update(id_ for id_ in played_map_ids if id_ in self.maps)
 
         least_common = counter.most_common()[::-1]
-        least_count = least_common[0][1]
+        least_count = 1
+        for map_count in least_common:
+            if not (isinstance(self.maps[map_count[0]], NeroxisGeneratedMap)):
```
Askaholic, line 44: What is the purpose of this? I don't think that generated maps will work with the anti-repetition logic in general because their map-ids will not exist in the database. This function is called only on data loaded from the database.

tests/integration_tests/test_game.py
```diff
@@ -68,29 +68,31 @@ async def open_fa(proto):
...
-        queue_name="ladder1v1",
+        queue_name=queue_name,
         timeout=5
     )
 
     return proto
 
 
-async def queue_players_for_matchmaking(lobby_server):
+async def queue_players_for_matchmaking(lobby_server, queue_name: str = 'ladder1v1'):
```
Askaholic, line 91: Use double quotes for strings.
```suggestion
async def queue_players_for_matchmaking(lobby_server, queue_name: str = "ladder1v1"):
```

server/matchmaker/map_pool.py
Sheikah45: The way it was previously least_common was all the maps with the lowest count of plays based on the list of played_map_ids. Since the generated map ids will never be in there then they would always have a cunt of 1 meaning eventually they would be the only maps in least_common

server/ladder_service.py
Sheikah45: It could be but wasn't sure since the type is in the params itself. and if another type was introduced it wouldn't be part of the neroxis method

server/ladder_service.py
Sheikah45: Aight

server/matchmaker/map_pool.py
Askaholic: So maybe it would be better to filter by the `Map` class instead? Since `Map` represents a map from the database, and any other types would not exist in the database.

server/ladder_service.py
Askaholic: IMO it makes sense to check the type first and then delegate to the appropriate method/function.

Askaholic: I'm just thinking that this function is basically a constructor, so it would make sense for it to be a class method on the class it's constructing.

server/matchmaker/map_pool.py
Sheikah45: Aight

server/ladder_service.py
Sheikah45: Aight

tests/integration_tests/test_game.py
Sheikah45: Aight

tests/data/test-data.sql
```diff
@@ -43,66 +43,100 @@ DELETE FROM ladder_division;
...
-  (103, 'ladder4', 'ladder4@example.com', SHA2('ladder4', 256), null, '2000-01-01 00:00:00'),
-  (104, 'ladder_ban', 'ladder_ban@example.com', SHA2('ladder_ban', 256), null, '2000-01-01 00:00:00'),
-  (105, 'tmm1', 'tmm1@example.com', SHA2('tmm1', 256), null, '2000-01-01 00:00:00'),
-  (106, 'tmm2', 'tmm2@example.com', SHA2('tmm2', 256), null, '2000-01-01 00:00:00'),
-  (200, 'banme', 'banme@example.com', SHA2('banme', 256), null, '2000-01-01 00:00:00'),
+insert into login (id, login, email, password, steamid, create_time)
+values (1, 'test', 'test@example.com', SHA2('test_password', 256), null, '2000-01-01 00:00:00'),
+       (2, 'Dostya', 'dostya@cybran.example.com', SHA2('vodka', 256), null, '2000-01-01 00:00:00'),
+       (3, 'Rhiza', 'rhiza@aeon.example.com', SHA2('puff_the_magic_dragon', 256), null, '2000-01-01 00:00:00'),
+(4, 'No_UID', 'uid@uef.example.com', SHA2('his_pw', 256), null, '2000-01-01 00:00:00'),
```
Askaholic, line 81: Lol wtf is this formatting?

tests/data/test-data.sql
```diff
@@ -301,18 +336,26 @@ insert into matchmaker_queue_game (matchmaker_queue_id, game_stats_id) values
...
-
-insert into map_pool_map_version (map_pool_id, map_version_id) values
-  (1, 15), (1, 16), (1, 17),
-  (2, 11), (2, 14), (2, 15), (2, 16), (2, 17),
-  (3, 1),  (3, 2),  (3, 3);
+  (3, "Large maps"),
+  (4, "Generated Maps with Errors");
+
+insert into map_pool_map_version (map_pool_id, map_version_id, weight, map_params) values
+                                                                                       (1, 15, 1, NULL),(1, 16, 1, NULL),(1, 17, 1, NULL),
```
Askaholic, line 343: ...

tests/unit_tests/test_ladder_service.py
```diff
@@ -67,18 +67,26 @@ async def test_load_from_database(ladder_service, queue_factory):
...
+            NeroxisGeneratedMap.of({"version": "0.0.0", "spawns": 2, "size": 512, "type": "neroxis"}),
+        ]
+
 
 @fast_forward(5)
 async def test_load_from_database_new_data(ladder_service, database):
     async with database.acquire() as conn:
-        result = await conn.execute(matchmaker_queue.insert().values(
+        await conn.execute(matchmaker_queue.insert().values(
+            id=5,
```
Askaholic, line 82: Oof that's annoying. Maybe make the id sufficiently large here that we won't need to worry about updating this test every time a new queue is added.

tests/unit_tests/test_map_pool.py
```diff
@@ -48,6 +88,24 @@ def test_choose_map_all_maps_played(map_pool_factory):
     assert chosen_map in maps
 
 
+def test_choose_map_all_played_but_generated_map_doesnt_dominate(map_pool_factory):
+    maps = [
+        Map(1, "some_map", "maps/some_map.v001.zip", 1000000),
```
Askaholic, line 93: why do theses have big weight?

Askaholic (changes requested):

tests/data/test-data.sql
```diff
@@ -43,66 +43,100 @@ DELETE FROM ladder_division;
...
-DELETE FROM leaderboard;
-DELETE FROM matchmaker_queue;
-DELETE FROM matchmaker_queue_game;
-DELETE FROM matchmaker_queue_map_pool;
-DELETE FROM map_pool;
-DELETE FROM map_pool_map_version;
-
-SET FOREIGN_KEY_CHECKS=1;
+DELETE
+FROM login;
```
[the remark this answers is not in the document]
Sheikah45: damn auto format

tests/data/test-data.sql
Sheikah45: I think it tried to format it as python sigh

tests/unit_tests/test_map_pool.py
Sheikah45: This is to ensure that they are chosen as one of them should be chosen unless they are not included which would happen if the check for Map type in the least common count didnt happen.

server/matchmaker/map_pool.py
```diff
@@ -39,14 +39,21 @@ def choose_map(self, played_map_ids: Iterable[int] = ()) -> Map:
...
         for i, (_, count) in enumerate(least_common):
-            if count != least_count:
+            if count > least_count:
                 least_common = least_common[:i]
                 break
 
-        return self.maps[random.choice(least_common)[0]]
+        weights = [self.maps[id_].weight for id_, _ in least_common]
+
+        return self.maps[random.choices(least_common, weights=weights, k=1)[0][0]].get_map()
```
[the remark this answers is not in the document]
Sheikah45: This isn't a named tuple at this point it is a tuple of (id, count)

server/ladder_service.py
```diff
@@ -125,8 +126,23 @@ async def fetch_map_pools(self, conn) -> Dict[int, Tuple[str, List[Map]]]:
...
-                    Map(row.map_id, row.display_name, row.filename)
+                    Map(row.map_id, row.display_name, row.filename, row.weight)
                 )
+            elif row.map_params is not None:
+                try:
+                    params = json.loads(row.map_params)
+                    map_type = params["type"]
+                    if map_type == "neroxis":
+                        map_list.append(NeroxisGeneratedMap.of(params=params, weight=row.weight))
+
```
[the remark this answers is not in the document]
Sheikah45: Aight

server/ladder_service.py
```diff
@@ -125,8 +126,23 @@ async def fetch_map_pools(self, conn) -> Dict[int, Tuple[str, List[Map]]]:
...
+                except Exception:
+                    self._logger.warning(
+                        "Failed to load map in map pool %d"
```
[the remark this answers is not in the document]
Sheikah45: Done

server/matchmaker/map_pool.py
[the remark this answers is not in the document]
Askaholic: Ah ok. Oh well

tests/unit_tests/test_map_pool.py
Askaholic: Right, ok

server/ladder_service.py
```diff
@@ -125,8 +126,26 @@ async def fetch_map_pools(self, conn) -> Dict[int, Tuple[str, List[Map]]]:
...
                 )
+            elif row.map_params is not None:
+                try:
+                    params = json.loads(row.map_params)
+                    map_type = params["type"]
+                    if map_type == "neroxis":
+                        map_list.append(NeroxisGeneratedMap.of(params=params, weight=row.weight))
+                    else:
+                        self._logger.warning("Unsupported map type %s with parameters %s",
+                                             map_type, row.map_params)
```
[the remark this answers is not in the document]
Askaholic: Also I'm not sure if its actually useful to log the map params here. Maybe the id would be better since if I see this warning I probably just want to be able to find that row in the database and fix it/remove it.

server/types.py
```diff
@@ -27,3 +29,56 @@ class Map(NamedTuple):
...
+            hand back None
+        """
+        seed_bytes = random.getrandbits(64).to_bytes(8, 'big')
+        size_byte = (self.map_size_pixels // 64).to_bytes(1, 'big')
+        spawn_byte = self.spawns.to_bytes(1, 'big')
+        option_bytes = spawn_byte + size_byte
+        seed_str = base64.b32encode(seed_bytes).decode("ascii").replace("=", "").lower()
+        option_str = base64.b32encode(option_bytes).decode("ascii").replace("=", "").lower()
+        map_name = f'neroxis_map_generator_{self.version}_{seed_str}_{option_str}'
+        map_path = f'maps/{map_name}.zip'
```
Askaholic, line 83: Please use double quotes for ALL strings

tests/unit_tests/test_ladder_service.py
```diff
@@ -67,18 +67,26 @@ async def test_load_from_database(ladder_service, queue_factory):
             Map(id=3, name="SCMP_003", path="maps/scmp_003.zip"),
         ]
 
+        queue = ladder_service.queues["neroxis1v1"]
+        assert queue.name == "neroxis1v1"
+        assert len(queue.map_pools) == 1
+        assert list(queue.map_pools[4][0].maps.values()) == [
+            NeroxisGeneratedMap.of({"version": "0.0.0", "spawns": 2, "size": 512, "type": "neroxis"}),
```
Askaholic, line 74: Line too long
```suggestion
            NeroxisGeneratedMap.of({
                "version": "0.0.0", 
                "spawns": 2, 
                "size": 512, 
                "type": "neroxis"
            }),
```

tests/unit_tests/test_map_pool.py
```diff
@@ -31,7 +33,45 @@ def test_choose_map(map_pool_factory):
...
+
+    map_pool = map_pool_factory(maps=[
+        NeroxisGeneratedMap.of({"version": "0.0.0", "spawns": 2, "size": 512, "type": "neroxis"}),
+    ])
+
+    chosen_map = map_pool.choose_map([])
+
+    map_id = -int.from_bytes(bytes(f'{version}_{spawns}_{size}', encoding="ascii"), 'big')
+    size_byte = (size // 64).to_bytes(1, 'big')
+    spawn_byte = spawns.to_bytes(1, 'big')
```
Askaholic, line 66: More single quote strings

tests/unit_tests/test_map_pool.py
```diff
@@ -31,7 +33,45 @@ def test_choose_map(map_pool_factory):
...
+        assert chosen_map == (4, "CHOOSE_ME", "maps/choose_me.v001.zip", 10000000)
+
+
+def test_choose_map_generated_map(map_pool_factory):
+    version = "0.0.0"
+    spawns = 2
+    size = 512
+
+    map_pool = map_pool_factory(maps=[
+        NeroxisGeneratedMap.of({"version": "0.0.0", "spawns": 2, "size": 512, "type": "neroxis"}),
```
Askaholic, line 59: Long line

Askaholic (commented):
Almost there. Now I'm only finding formatting things to complain about :P.

server/ladder_service.py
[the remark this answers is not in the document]
Sheikah45: What format do you prefer?

server/types.py
Sheikah45: yep my habit from another life

server/db/models.py
```diff
@@ -256,8 +256,11 @@
 
 map_pool_map_version = Table(
     "map_pool_map_version", metadata,
-    Column("map_pool_id",       Integer, ForeignKey("map_pool.id"),     nullable=False),
-    Column("map_version_id",    Integer, ForeignKey("map_version.id"),  nullable=False),
+    Column("id", Integer, primary_key=True),
```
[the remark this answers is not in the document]
Sheikah45: yep I should just turn off the autoformat for this project XD

server/ladder_service.py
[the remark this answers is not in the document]
Sheikah45: oh never mind I see it

server/db/models.py
```diff
@@ -256,8 +256,11 @@
 
 map_pool_map_version = Table(
     "map_pool_map_version", metadata,
-    Column("map_pool_id",       Integer, ForeignKey("map_pool.id"),     nullable=False),
-    Column("map_version_id",    Integer, ForeignKey("map_version.id"),  nullable=False),
+    Column("id",            Integer,        primary_key=True),
+    Column("map_pool_id",   Integer,        ForeignKey("map_pool.id"),      nullable=False),
+    Column("map_version_id", Integer,       ForeignKey("map_version.id")),
```
Askaholic, line 261: Can we line these up
```suggestion
    Column("map_pool_id",    Integer,       ForeignKey("map_pool.id"),      nullable=False),
    Column("map_version_id", Integer,       ForeignKey("map_version.id")),
```

Askaholic (approved):
I think it looks good. Tested it and resolved an issue with the timeout

server/db/models.py
```diff
@@ -256,8 +256,11 @@
 
 map_pool_map_version = Table(
     "map_pool_map_version", metadata,
-    Column("map_pool_id",       Integer, ForeignKey("map_pool.id"),     nullable=False),
-    Column("map_version_id",    Integer, ForeignKey("map_version.id"),  nullable=False),
+    Column("id",            Integer,        primary_key=True),
+    Column("map_pool_id",    Integer,       ForeignKey("map_pool.id"),      nullable=False),
+    Column("map_version_id", Integer,       ForeignKey("map_version.id")),
+    Column("weight",        Integer,        nullable=False),
+    Column("map_params",    Text),
```
Askaholic, line 263: xD I meant it should be fixed for all of them
```suggestion
    Column("id",             Integer,       primary_key=True),
    Column("map_pool_id",    Integer,       ForeignKey("map_pool.id"),      nullable=False),
    Column("map_version_id", Integer,       ForeignKey("map_version.id")),
    Column("weight",         Integer,       nullable=False),
    Column("map_params",     Text),
```

server/ladder_service.py
```diff
@@ -125,8 +126,29 @@ async def fetch_map_pools(self, conn) -> Dict[int, Tuple[str, List[Map]]]:
...
                 map_list.append(
-                    Map(row.map_id, row.display_name, row.filename)
+                    Map(row.map_id, row.display_name, row.filename, row.weight)
                 )
+            elif row.map_params is not None:
+                try:
+                    params = json.loads(row.map_params)
+                    map_type = params["type"]
+                    if map_type == "neroxis":
+                        map_list.append(NeroxisGeneratedMap.of(params=params, weight=row.weight))
```
Askaholic, line 136: keyword arguments really necessary here?
```suggestion
                        map_list.append(NeroxisGeneratedMap.of(params, row.weight))
```

PR #734, opened 2021-03-02
Fixup codecov

Askaholic:
I think it wants the full file paths. In coveralls these were just clutter, but codecov needs to map them to files in the repo so it can compare them to the git diff.


Askaholic:
Ok, and it needed the report to be in xml format.

Issue #735, opened 2021-03-03
Czar unit blueprint is wrong


Askaholic:
It should be `UAA0310` according to [unitdb.faforever.com](https://unitdb.faforever.com). This is causing the czar achievement not to work. Maybe someone can go through and audit all the unit blueprints to make sure we have the right constants in there.

I think the prefixes map like this:
- UE - UEF
- UR - Cybran
- UA - Aeon
- XS - Seraphim

MrRowey:
- Fixed Czar Wrong BP Listed
- Moved Salvation under Experimental field ( was move to Experimental tab Years ago)

PR #736, opened 2021-03-04
Fixed Unit Bp & Updated Experimentals

MrRowey:
- #735 Fixed the Czar Unit Bp set as mavor
- Added Salvation Under Experimental


MrRowey:
Not Quite sure why it failed dose not give me an indication to what I am familiar with.

Askaholic:
Awesome, thanks for taking a look at that! 

Looks like another flaky test I need to fix a little bit. It just barely timed out too early so I'm sure when it re-runs it will succeed.

PR #740, opened 2021-03-06
Issue/#724 generalize rating initialization

### LINKED ISSUE #724: Generalize rating initialization based on another rating
For 2v2 we wanted players ratings to be initialized based on their global rating, but with a sigma increase. This was hardcoded into the player rating dictionary here:
https://github.com/FAForever/server/blob/d2265dcdf5164f8f00e55b6b360e46d6ebb81b1e/server/rating.py#L39-L47

For future 3v3+ queues we will want to do a similar thing, so it would be best to fully generalize this process. We should add a field in the database to the `leaderboard` table that will reference another leaderboard id to initialize from. I think we can leave the sigma increase as it is, hardcoded to +150 sigma with a 250 sigma cap. I don't see a reason why those values would need to be configured in the database; they would only complicate the tables.

Askaholic:
Implement reading the initializer from the database.

One question is whether the `global` and `ladder_1v1` ratings need to continue to be eagerly created. Right now even completely new players will have a global and ladder rating in their `ratings` dictionary, however is this really necessary?

Closes #724

### DIFF
--- a/server/rating.py
+++ b/server/rating.py
@@ -3,11 +3,15 @@ Type definitions for player ratings
 """
 
 from dataclasses import dataclass
-from typing import DefaultDict, Optional, Tuple, TypeVar, Union
+from typing import Dict, Optional, Tuple, TypeVar, Union
 
-from trueskill import Rating
+import trueskill
 
+from server.config import config
 from server.weakattr import WeakAttribute
+
+Rating = Tuple[float, float]
+V = TypeVar("V")
 
 
 @dataclass(init=False)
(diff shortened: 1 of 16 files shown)


Askaholic:
I'm going to change how I'm handling dependency cycles. Instead of disallowing cycles, I'm going to rewrite the code so it only ever looks back 1 step, instead of trying to look back recursively. That way if e.g. global is initialized based on 2v2, and 2v2 based on global, then essentially whichever one gets played first will be used as an initializer for the other. 

Also need to do this in the rating service as well, since unfortunately this code is duplicated.

Askaholic:
I decided that the desired behavior is actually to initialize ratings recursively, and use the default rating when a cycle is detected. This would both allow dependency chains like:
- `2v2 -> global -> ladder1v1`. If neither 2v2 nor global exist, but ladder does exist then both global and 2v2 will be initialized based on ladder. The only tricky thing here is if the deviation for ladder is really small we don't actually want to increase it twice.
and cycles like:
- `2v2 -> global -> 2v2`. If one of them exists, the other will be based on the existing one, and if neither exist then a default will be used for whichever one is requested first.

Given one of the recent client bugs we saw with trying to display map pools when no rating existed, I think it is actually still a good idea to eagerly initialize ratings. However, this can cause a weird scenario where for example the following happens:
1. Player who has never played 2v2 logs in, and their 2v2 rating is initialized to their global mean
2. Player plays a whole bunch of global games changing their mean significantly (maybe they're kinda new and dev is high so mean changes fast)
3. Player then queues 2v2, but the current 2v2 rating used by the matchmaker is still based on their old global rating when they logged in
4. Game finishes, their 2v2 rating is re-initialized based on the new global rating and persisted to the database, from now on rating should be as expected.

In this case the first 2v2 game will be balanced based on outdated info. So I think it would be ideal for the `PlayerRatings` dict to keep track of which ratings were set externally (meaning they are more or less tied to the database in some way) and which ratings were automatically set as part of the initialization process. These automatically set ratings should then be re-computed whenever they're queried to ensure they are up to date. They do need to be set on the object so that they can be iterated over, in which case they don't need to be recomputed.

Askaholic:
I think I've implemented everything that I mentioned above.
- Recursive initialization with cycle detection
- Recomputing auto initialized ratings whenever they are queried

I also eliminated the duplication by making the rating service use `PlayerRatings` objects for computing initialization. It now loads all of a players ratings from the database and throws them into a `PlayerRatings` object, then it queries the object for the particular rating that its interested in.

BlackYps (approved):
I thought about how we want the dependency chains to be in practice.
I think ideally we would want `global <--> 4v4 <--> 2v2 <--> 1v1`. However this requires two references and a way to determine which preference should be preferred, if both exist.
With one reference we can build the chain `1v1 -> 2v2 -> 4v4 -> global -> 2v2`. I think this one will work good enough, so we can save the trouble of implementing the double reference version.

server/rating.py
```diff
@@ -2,58 +2,126 @@
...
+        return super().__getitem__(rating_type)
+
+    def _get_initial_rating(
+        self,
+        rating_type: str,
+        history: Set[str],
+    ) -> Rating:
+        """Create an initial rating when no rating exists yet."""
+        leaderboard = self.leaderboards.get(rating_type)
+        if leaderboard is None or leaderboard.initializer is None:
```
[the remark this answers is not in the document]
Askaholic: Because `.get` can return `None` and then `None.initializer` would throw an exception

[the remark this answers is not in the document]
Askaholic: That should only ever happen if somehow a leaderboard key is requested that doesn’t actually exist.

server/rating.py
```diff
@@ -2,58 +2,126 @@
...
+class PlayerRatings(Dict[str, Rating]):
+    def __init__(self, leaderboards: Dict[str, Leaderboard], init: bool = True):
+        self.leaderboards = leaderboards
+        # Rating types which are present but should be recomputed.
+        self.transient: Set[str] = set()
+
+        # Initialize known rating types so the client can display them
+        if init:
+            _ = self[RatingType.GLOBAL]
+            _ = self[RatingType.LADDER_1V1]
```
Sheikah45, line 61: The client no longer makes any assumptions about what ratings exist and will just set them to 0 if they do not exist

Askaholic: Ok cool. I'll mark this as deprecated then.

tests/unit_tests/test_rating_service.py
```diff
@@ -162,6 +162,12 @@ async def test_load_rating_type_ids(uninitialized_service):
         "ladder_1v1": 2,
         "tmm_2v2": 3
     }
+    global_ = Leaderboard(1, "global")
+    assert service.leaderboards == {
+        "global": global_,
+        "ladder_1v1": Leaderboard(2, "ladder_1v1"),
+        "tmm_2v2": Leaderboard(3, "tmm_2v2", global_)
```
BlackYps, line 169: This seems to ignore the initializer leaderboard. No matter what I input in the test data, it always passes.

Askaholic: Good catch, looks like the generated implementation for `__eq__` was ignoring the leaderboard. Adding a type annotation fixes it

PR #749, opened 2021-03-21
Feature/#727 refactor matchmaker for 4v4 party algorithm

### LINKED ISSUE #801: Searches don't register failed matching attempts if too few people are in queue
If there are too few people in a matchmaker queue to fill a game the function exits early [here](https://github.com/FAForever/server/blob/58c5887e386789f406532fba38e7c49d7e73642a/server/matchmaker/matchmaker_queue.py#L153) and misses to register failed matching attemps because the actual [matchmaker](https://github.com/FAForever/server/blob/58c5887e386789f406532fba38e7c49d7e73642a/server/matchmaker/matchmaker_queue.py#L161), that handles all this never gets called.

cleborys:
Refactors `server/matchmaker/algorithm.py` such that the matchmaking algorithm is independent from the queue logic, with the intent that @BlackYps can simply subclass a new `Matchmaker` to propose a new algorithm.

For now I stuck to minimal/minor changes, although I think these files could still benefit a lot from more attention. For example I'm not convinced that these `Matchmaker`s should have instance variables at all but left that as-is for a chance to get feedback first.

Closes #801.

### DIFF
--- a/server/matchmaker/algorithm/matchmaker.py
+++ b/server/matchmaker/algorithm/matchmaker.py
@@ -1,20 +1,8 @@
-import itertools
-import math
-import random
-import statistics as stats
-from collections import OrderedDict
-from typing import (
-    Dict,
-    Iterable,
-    Iterator,
-    List,
-    Optional,
-    Set,
-    Tuple,
-)
+from typing import Iterable, List
 
 from ...decorators import with_logger
-from ..search import CombinedSearch, Match, Search
+from ..search import Match, Search
+
 
 @with_logger
 class Matchmaker(object):
--- a/server/matchmaker/algorithm/matchmaker.py
+++ b/server/matchmaker/algorithm/matchmaker.py
@@ -1,4 +1,4 @@
-from typing import Iterable, List, Tuple
+from typing import Dict, Iterable, List, Set, Tuple
 
 from ...decorators import with_logger
 from ..search import Match, Search
--- a/server/matchmaker/algorithm/matchmaker.py
+++ b/server/matchmaker/algorithm/matchmaker.py
@@ -1,4 +1,4 @@
-from typing import Iterable, List
+from typing import Iterable, List, Tuple
 
 from ...decorators import with_logger
 from ..search import Match, Search
@@ -6,10 +6,38 @@ from ..search import Match, Search
 
 @with_logger
 class Matchmaker(object):
-    def __init__(self, team_size: int):
-        self.team_size = team_size
-
-    def find(self, searches: Iterable[Search]) -> List[Match]:
+    def find(
+        self, searches: Iterable[Search], team_size: int
+    ) -> Tuple[List[Match], List[Search]]:
         raise NotImplementedError(
             "Matchmaker.find should be implemented by concrete subclasses"
         )
+
+
+@with_logger
+class MatchmakingPolicy1v1(object):
+    def __init__(self):
+        self.matches: Dict[Search, Search] = {}
+        self.searches_remaining_unmatched: Set[Search] = set()
+
+    def _match(self, s1: Search, s2: Search):
+        self._logger.debug(
+            "Matching %s and %s (%s)",
+            s1, s2, self.__class__
+        )
+        self.matches[s1] = s2
+        self.matches[s2] = s1
+        self.searches_remaining_unmatched.discard(s1)
+        self.searches_remaining_unmatched.discard(s2)
+
+    def _unmatch(self, s1: Search):
+        s2 = self.matches[s1]
+        self._logger.debug(
+            "Unmatching %s and %s (%s)",
+            s1, s2, self.__class__
+        )
+        assert self.matches[s2] == s1
+        del self.matches[s1]
+        del self.matches[s2]
+        self.searches_remaining_unmatched.add(s1)
+        self.searches_remaining_unmatched.add(s2)
--- a/server/matchmaker/algorithm/matchmaker.py
+++ b/server/matchmaker/algorithm/matchmaker.py
@@ -1,0 +1,27 @@
+import itertools
+import math
+import random
+import statistics as stats
+from collections import OrderedDict
+from typing import (
+    Dict,
+    Iterable,
+    Iterator,
+    List,
+    Optional,
+    Set,
+    Tuple,
+)
+
+from ...decorators import with_logger
+from ..search import CombinedSearch, Match, Search
+
+@with_logger
+class Matchmaker(object):
+    def __init__(self, team_size: int):
+        self.team_size = team_size
+
+    def find(self, searches: Iterable[Search]) -> List[Match]:
+        raise NotImplementedError(
+            "Matchmaker.find should be implemented by concrete subclasses"
+        )
--- a/server/matchmaker/algorithm/matchmaker.py
+++ b/server/matchmaker/algorithm/matchmaker.py
@@ -1,3 +1,4 @@
+from abc import ABC, abstractmethod
 from typing import Dict, Iterable, List, Set, Tuple
 
 from ...decorators import with_logger
@@ -5,13 +6,14 @@ from ..search import Match, Search
 
 
 @with_logger
-class Matchmaker(object):
+class Matchmaker(ABC):
+    @abstractmethod
     def find(
-        self, searches: Iterable[Search], team_size: int
+        self,
+        searches: Iterable[Search],
+        team_size: int
     ) -> Tuple[List[Match], List[Search]]:
-        raise NotImplementedError(
-            "Matchmaker.find should be implemented by concrete subclasses"
-        )
+        pass
 
 
 @with_logger
--- a/server/matchmaker/algorithm/stable_marriage.py
+++ b/server/matchmaker/algorithm/stable_marriage.py
@@ -108,7 +108,8 @@ class StableMarriageMatchmaker(Matchmaker):
 
         return self._remove_duplicates(matches), unmatched_searches
 
-    def _remove_duplicates(self, matches: Dict[Search, Search]) -> List[Match]:
+    @staticmethod
+    def _remove_duplicates(matches: Dict[Search, Search]) -> List[Match]:
         matches_set: Set[Match] = set()
         for s1, s2 in matches.items():
             if (s1, s2) in matches_set or (s2, s1) in matches_set:
@@ -178,7 +179,6 @@ class _MatchingGraph:
 
     @staticmethod
     def is_possible_match(search: Search, other: Search, quality: float) -> bool:
-        log_string = "Quality between %s and %s: %.3f thresholds: [%.3f, %.3f]."
         log_args = (
             search,
             other,
--- a/server/matchmaker/algorithm/stable_marriage.py
+++ b/server/matchmaker/algorithm/stable_marriage.py
@@ -11,26 +11,17 @@ from typing import (
     Optional,
     Set,
     Tuple,
-    TypeVar
 )
 
-from ..decorators import with_logger
-from .search import CombinedSearch, Match, Search
+from ...decorators import with_logger
+from ..search import CombinedSearch, Match, Search
 
-T = TypeVar("T")
 WeightedGraph = Dict[Search, List[Tuple[Search, float]]]
-Buckets = Dict[Search, List[Tuple[Search, float]]]
 
-
-def make_matches(searches: Iterable[Search]) -> List[Match]:
-    """
-    Main entrypoint for the matchmaker algorithm.
-    """
-    return Matchmaker(searches).find()
 
 
 @with_logger
-class MatchmakingPolicy(object):
+class MatchmakingPolicy1v1(object):
     def __init__(self):
         self.matches: Dict[Search, Search] = {}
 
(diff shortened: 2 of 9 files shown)


cleborys:
Github doesn't seem to re-use files for several diffs where a single file has been split into two. To get smaller diffs, you can compare 
 the deleted `test_matchmaker_queue_algorithm.py` to `test_matchmaker_algorithm_bucket_teams.py` and `test_matchmaker_algorithm_stable_marriage.py` and the deleted `matchmaker/algorithm.py` to all of the `matchmaker/algorithm/*.py` files.

server/matchmaker/algorithm/stable_marriage.py
```diff
@@ -0,0 +1,262 @@
...
+WeightedGraph = Dict[Search, List[Tuple[Search, float]]]
+
+
+@with_logger
+class MatchmakingPolicy1v1(object):
+    def __init__(self):
+        self.matches: Dict[Search, Search] = {}
+
+    def _match(self, s1: Search, s2: Search):
+        self._logger.debug(f"Matching %s and %s ({self.__class__})", s1, s2)
```
Askaholic, line 19: Mixing format string and % formatting. Just use the % version for loggers

server/matchmaker/algorithm/stable_marriage.py
```diff
@@ -0,0 +1,262 @@
...
+            search,
+            other,
+            quality,
+            search.match_threshold,
+            other.match_threshold,
+        )
+
+        if search._match_quality_acceptable(other, quality):
+            _MatchingGraph._logger.debug(
+                f"{log_string} Will be considered during stable marriage.", *log_args
```
Askaholic, line 237: Not a fan of this f-string stuff with the logger. Imo the duplication is fine here.

server/matchmaker/matchmaker_queue.py
```diff
@@ -177,23 +184,30 @@ async def find_matches(self) -> None:
...
+        attempt.
+        """
+        searches_set = set(searches)
+        matched_searches = set(
+            search
+            for match in matches
+            for combinedSearch in match
+            for search in combinedSearch.get_original_searches()
+        )
+        unmatched_searches = searches_set - matched_searches
```
Askaholic, line 203: Maybe it would be better to define the matchmaker interface to return both a list of matches and a list of unmatched searches, like how it was with the `make_teams` functions?

Askaholic (approved):
I agree that there is no point in many of the objects used. If they're just holding 1 instance variable then that could really be moved to a function parameter.

tests/unit_tests/test_matchmaker_queue.py
```diff
@@ -436,8 +436,8 @@ def make_matches(*args):
         return []
 
     with mock.patch(
-        "server.matchmaker.matchmaker_queue.make_matches",
-        make_matches
+        "server.matchmaker.matchmaker_queue.StableMarriageMatchmaker.find",
```
Askaholic, line 439: This lookup path is causing the tests to fail. I think this can now be done a lot easier by mocking the `matchmaker` object directly:

```python
queue.matchmaker.find = mock.Mock(side_effect=find)
```
something like that

server/matchmaker/matchmaker_queue.py
cleborys: Good point

server/matchmaker/algorithm.py
```diff
@@ -1,502 +0,0 @@
...
-            search for search in self.searches if search not in self.matches
-        ]
-        self._logger.debug("Matching randomly for remaining newbies...")
-        self.matches.update(RandomlyMatchNewbies().find(remaining_searches))
-
-        self._register_unmatched_searches()
-
-        return self._remove_duplicates()
-
-    def _register_unmatched_searches(self):
```
BlackYps, line 162: This is missing in the new stable marriage file. Is that deliberate?

BlackYps: Should have looked a bit further down. I found it in matchmaker_queue.py

cleborys:
I finally included the suggestion to return a list of unmatched searches from the matchmakers, rather than taking the difference after the fact.
The remaining codacy issues all look like false positives.

server/matchmaker/algorithm/matchmaker.py
```diff
@@ -0,0 +1,43 @@
...
+
+
+@with_logger
+class Matchmaker(object):
+    def find(
+        self, searches: Iterable[Search], team_size: int
+    ) -> Tuple[List[Match], List[Search]]:
+        raise NotImplementedError(
+            "Matchmaker.find should be implemented by concrete subclasses"
+        )
```
Askaholic, line 14: You can use the `abc.ABC` (Abstract Base Class) base class to raise these exceptions for you

server/matchmaker/algorithm/stable_marriage.py
```diff
@@ -0,0 +1,221 @@
...
+    and afterwards adds random matchups for previously unmatched new players.
+    """
+    def find(
+        self, searches: Iterable[Search], team_size: int
+    ) -> Tuple[List[Match], List[Search]]:
+        if team_size != 1:
+            self._logger.error(
+                "Invalid team size %i for stable marriage matchmaker will be ignored",
+                team_size,
+            )
```
Askaholic, line 88: Should there be a return here?

server/matchmaker/matchmaker_queue.py
```diff
@@ -152,14 +154,27 @@ async def find_matches(self) -> None:
         if self.num_players < 2 * self.team_size:
             return
```
Askaholic, line 155: Note to self, @BlackYps pointed out that before returning early here we should be calling `register_failed_matching_attempt`

Askaholic: Or I guess use your function `_register_unmatched_searches`

server/matchmaker/matchmaker_queue.py
```diff
@@ -152,14 +154,27 @@ async def find_matches(self) -> None:
...
+        proposed_matches, unmatched_searches = await loop.run_in_executor(
+            None,
+            self.matchmaker.find,
+            searches,
+            self.team_size,
+        )
+
+        # filter out matches that were cancelled
+        matches: List[Match] = []
+        for match in proposed_matches:
```
Askaholic, line 170: You could unpack here so you don't need to index into `match` later
```suggestion
        for s1, s2 in proposed_matches:
```

tests/unit_tests/test_matchmaker_algorithm_bucket_teams.py
```diff
@@ -0,0 +1,312 @@
...
+    matched, non_matched = _make_teams_from_single(searches, size=2)
+
+    assert matched != []
+    assert non_matched == []
+
+    for search in matched:
+        p1, p2 = search.players
+        p1_mean, _ = p1.ratings[RatingType.LADDER_1V1]
+        p2_mean, _ = p2.ratings[RatingType.LADDER_1V1]
+        #
```
Askaholic, line 78: delete?

tests/unit_tests/test_matchmaker_algorithm_bucket_teams.py
```diff
@@ -0,0 +1,312 @@
...
+
+
+@pytest.mark.parametrize("make_teams_func", (_make_teams, _make_teams_from_single))
+@given(
+    searches=st_searches_list(max_players=1),
+    size=st.integers(min_value=1, max_value=10),
+)
+def test_make_teams_single_correct_size(searches, size, make_teams_func):
+    matched, _ = make_teams_func(searches, size)
+
```
Askaholic, line 47: We could `assume(matched != [])` here to make sure the examples being generated are actually relevant. This will also increase the time it takes for the test to finish since it will be generating more examples.

tests/unit_tests/test_matchmaker_algorithm_bucket_teams.py
```diff
@@ -0,0 +1,312 @@
...
+            p1, p2 = search.players
+            # Order doesn't matter
+            if p1.ratings[RatingType.LADDER_1V1][0] > 900:
+                assert p2.ratings[RatingType.LADDER_1V1][0] < 600
+            else:
+                assert p1.ratings[RatingType.LADDER_1V1][0] < 600
+                assert p2.ratings[RatingType.LADDER_1V1][0] > 900
+
+
+def test_make_buckets_performance(bench, player_factory):
```
Askaholic, line 114: Isn't there a pytest mark for bench tests now? Or is that added by @BlackYps?

tests/unit_tests/test_matchmaker_algorithm_bucket_teams.py
```diff
@@ -0,0 +1,312 @@
...
+
+def test_2v2_count_unmatched_searches(player_factory):
+    players = [
+            player_factory(500, 100, name="lowRating_unmatched_1"),
+            player_factory(500, 100, name="lowRating_unmatched_2"),
+            player_factory(1500, 100, name="midRating_matched_1"),
+            player_factory(1500, 100, name="midRating_matched_2"),
+            player_factory(1500, 100, name="midRating_matched_3"),
+            player_factory(1500, 100, name="midRating_matched_4"),
+            player_factory(2000, 100, name="highRating_unmatched_1"),
```
Askaholic, line 300: dedent 1 level

tests/unit_tests/test_matchmaker_algorithm_stable_marriage.py
```diff
@@ -488,248 +503,74 @@ def test_matchmaker_random_only(player_factory):
...
+def test_unmatched_searches_without_newbies(player_factory):
+    players = [
+            player_factory(100, 10, name="lowRating_unmatched_1"),
+            player_factory(500, 10, name="lowRating_unmatched_2"),
+            player_factory(1500, 10, name="midRating_matched_1"),
+            player_factory(1500, 10, name="midRating_matched_2"),
+            player_factory(1500, 10, name="midRating_matched_3"),
+            player_factory(1500, 10, name="midRating_matched_4"),
+            player_factory(2000, 10, name="highRating_unmatched_1"),
+            player_factory(2500, 10, name="highRating_unmatched_2"),
```
Askaholic, line 537: dedent 1 level

tests/unit_tests/test_matchmaker_algorithm_stable_marriage.py
```diff
@@ -488,248 +503,74 @@ def test_matchmaker_random_only(player_factory):
...
+            player_factory(200, 10, name="newbie2", ladder_games=1),
+            player_factory(300, 10, name="newbie3", ladder_games=1),
+            player_factory(400, 10, name="newbie4", ladder_games=1),
+            player_factory(500, 10, name="newbie5", ladder_games=1),
+            player_factory(1500, 10, name="midRating_matched_1"),
+            player_factory(1500, 10, name="midRating_matched_2"),
+            player_factory(1500, 10, name="midRating_matched_3"),
+            player_factory(1500, 10, name="midRating_matched_4"),
+            player_factory(2000, 10, name="highRating_unmatched_1"),
+            player_factory(2500, 10, name="highRating_unmatched_2"),
```
Askaholic, line 561: dedent 1 level

tests/unit_tests/test_matchmaker_queue.py
```diff
@@ -420,11 +418,32 @@ async def find_matches():
...
+    for search in searches:
+        asyncio.create_task(matchmaker_queue.search(search))
+
+    searches[0].cancel()
+
+    await asyncio.sleep(0.01)
+    await matchmaker_queue.find_matches()
+
+    for search in searches[1:]:
+        assert search.is_matched ^ (search.failed_matching_attempts == 1)
```
Askaholic, line 436: WTF!! Using xor for something that's not encryption???? My brain hurts, I'm not used to thinking about xor as an actual logical operator xD

tests/unit_tests/test_matchmaker_queue.py
```diff
@@ -433,21 +452,37 @@ def make_matches(*args):
...
+        queue._register_unmatched_searches = mock.Mock()
+        queue.matchmaker.find = mock.Mock(side_effect=find)
+
+    await asyncio.gather(*[
+        queue.find_matches() for queue in queues
+    ])
+
+
+@pytest.mark.asyncio
+async def test_queue_pop_communicates_failed_attempts(matchmaker_queue, player_factory):
```
Askaholic, line 474: As @BlackYps pointed out, this would currently fail if there was only 1 player in the queue.

tests/unit_tests/test_matchmaker_algorithm_bucket_teams.py
BlackYps: Yeah, that gets added by my PR

server/matchmaker/algorithm/stable_marriage.py
cleborys: No, I wouldn't say so - it will simply ignore what you passed in and the parameter is only there to comply with the `Matchmaker` interface. By "ignore" I mean it will continue to execute as if you had passed a `team_size` of 1 (and implicitly assume that all `Search`es passed in `searches` correspond to a single player).

tests/unit_tests/test_matchmaker_queue.py
cleborys: xD At least in python you can do xor of booleans here unlike in certain other languages where you have to write `search.is_matched != (search.failed_matching_attempts == 1)` for an xor of two booleans.
I understand that they're the same thing, but they somehow read very differently...

server/matchmaker/algorithm/stable_marriage.py
Askaholic: Oh ok. So really the parameter is always ignored, and here it's just logging that something weird is going on. Wouldn't it be better to just `assert team_size == 1` then to avoid programming errors?

PR #755, opened 2021-03-30
Remove Hardcoded Equlibrium code

MrRowey:


Askaholic:
There’s a bit more to do. I want to get rid of the constant definition as well so we’ll have to do a search through the entire codebase to find every mention of 'equilibrium'. It’ll be in the tests too, and possibly in some other random parts of the game handling code.

PR #760, opened 2021-04-02
Feature/#699 Add queue_pop_time_delta to MatchmakerQueue representation

### LINKED ISSUE #699: Add time delta for queue pop time
As it turns out, it's actually pretty common for the queue pop times to display incorrectly because of weirdly configured system time on the client side. Since knowing the exact pop to the millisecond isn't crucial, we should add a time delta to the matchmaker_info message as well. I suggest we leave the timestamp in so that the client implementation can choose which one to use.

I'd say call it `queue_pop_time_delta` and make it number of seconds as a float.

eoinnoble:
Closes #699.

Not sure if these were the kinds of tests you had in mind, let me know!

### DIFF
--- a/server/matchmaker/matchmaker_queue.py
+++ b/server/matchmaker/matchmaker_queue.py
@@ -2,7 +2,7 @@ import asyncio
 import time
 from collections import OrderedDict
 from concurrent.futures import CancelledError
-from datetime import datetime, timedelta, timezone
+from datetime import datetime, timezone
 from typing import Any, Callable, Dict, Iterable, List, Optional, Tuple
 
 import server.metrics as metrics
--- a/server/matchmaker/matchmaker_queue.py
+++ b/server/matchmaker/matchmaker_queue.py
@@ -247,9 +247,7 @@ class MatchmakerQueue:
             "queue_pop_time": datetime.fromtimestamp(
                 self.timer.next_queue_pop, timezone.utc
             ).isoformat(),
-            "queue_pop_time_delta": timedelta(
-                seconds=self.timer.next_queue_pop - time.time()
-            ).total_seconds(),
+            "queue_pop_time_delta": self.timer.next_queue_pop - time.time(),
             "num_players": self.num_players,
             "boundary_80s": [search.boundary_80 for search in self._queue.keys()],
             "boundary_75s": [search.boundary_75 for search in self._queue.keys()],
--- a/server/matchmaker/matchmaker_queue.py
+++ b/server/matchmaker/matchmaker_queue.py
@@ -2,7 +2,7 @@ import asyncio
 import time
 from collections import OrderedDict
 from concurrent.futures import CancelledError
-from datetime import datetime, timezone
+from datetime import datetime, timedelta, timezone
 from typing import Any, Callable, Dict, Iterable, List, Optional, Tuple
 
 import server.metrics as metrics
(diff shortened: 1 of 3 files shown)


server/matchmaker/matchmaker_queue.py
```diff
@@ -247,6 +247,9 @@ def to_dict(self):
             "queue_pop_time": datetime.fromtimestamp(
                 self.timer.next_queue_pop, timezone.utc
             ).isoformat(),
+            "queue_pop_time_delta": timedelta(
+                seconds=self.timer.next_queue_pop - time.time()
+            ).total_seconds(),
```
Askaholic, line 252: Is there really a point to converting from seconds to `timedelta` and back again here? Just seems funny to use the `seconds` parameter and then call `total_seconds` :P

tests/integration_tests/test_matchmaker.py
```diff
@@ -368,13 +374,20 @@ async def test_matchmaker_info_message(lobby_server, mocker):
         assert "num_players" in queue
 
         assert queue["queue_pop_time"] == "2019-07-01T16:53:21+00:00"
+        assert queue["queue_pop_time_delta"] == math.ceil(
+            config.QUEUE_POP_TIME_MAX / 2
```
Askaholic, line 378: Why is it half here? I guess it's because it's getting the first queue pop? I vaguely remember scheduling the first pop for half the max time.

Askaholic (commented):
I think the tests look good! I wonder if we really need to convert to `timedelta` if we already have the time in seconds.

server/matchmaker/matchmaker_queue.py
eoinnoble: `self.timer.next_queue_pop` would be seconds already, yeah, but seconds since epoch plus some small scheduling window (`QUEUE_POP_TIME_MAX`). I assumed you wanted the seconds remaining until the next queue pop, but maybe you actually wanted to return a `timedelta` object?

tests/integration_tests/test_matchmaker.py
eoinnoble: That's right, this is the first queue pop, so I wanted to protect the test in case `QUEUE_POP_TIME_MAX` ever changed. [This is where it's used](https://github.com/FAForever/server/blob/develop/server/matchmaker/pop_timer.py#L36) in the `PopTimer`.

server/matchmaker/matchmaker_queue.py
Askaholic: What I mean is this code is basically doing this:
```python3
assert timedelta(seconds=30).to_seconds() == 30
```
Converting the number of seconds to a timedelta and then back to the original number. Wouldn't it be simpler just to use
```python3
"queue_pop_time_delta": self.timer.next_queue_pop - time.time()
```

tests/integration_tests/test_matchmaker.py
Askaholic: Using the config is fine, I was just trying to remember why the `/ 2` was needed.

server/matchmaker/matchmaker_queue.py
eoinnoble: Ah gotcha, ok!

server/matchmaker/matchmaker_queue.py
```diff
@@ -2,7 +2,7 @@
 import time
 from collections import OrderedDict
 from concurrent.futures import CancelledError
-from datetime import datetime, timezone
+from datetime import datetime, timedelta, timezone
```
Askaholic, line 5: Import unused now

Askaholic (approved):
Cool, looks fine to me now

server/matchmaker/matchmaker_queue.py
eoinnoble: 🤦

Issue #661, opened 2020-09-15
Better logging for malformed messages


Askaholic:
There are a number of mysterious errors which indicate that sometimes clients are sending malformed protocol messages:

```
ERROR    Sep 03  21:05:08 ServerContext                  Invalid control character at: line 1 column 728 (char 727)
Traceback (most recent call last):
  File "/code/server/servercontext.py", line 80, in client_connected
    message = await protocol.read_message()
  File "/code/server/protocol/qdatastreamprotocol.py", line 119, in read_message
    message = json.loads(action)
  File "/usr/local/lib/python3.7/json/__init__.py", line 348, in loads
    return _default_decoder.decode(s)
  File "/usr/local/lib/python3.7/json/decoder.py", line 337, in decode
    obj, end = self.raw_decode(s, idx=_w(s, 0).end())
  File "/usr/local/lib/python3.7/json/decoder.py", line 353, in raw_decode
    obj, end = self.scan_once(s, idx)
json.decoder.JSONDecodeError: Invalid control character at: line 1 column 728 (char 727)
```

```
ERROR    Sep 02  20:31:37 ServerContext                  Expecting ',' delimiter: line 1 column 10652 (char 10651)
Traceback (most recent call last):
  File "/code/server/servercontext.py", line 80, in client_connected
    message = await protocol.read_message()
  File "/code/server/protocol/qdatastreamprotocol.py", line 119, in read_message
    message = json.loads(action)
  File "/usr/local/lib/python3.7/json/__init__.py", line 348, in loads
    return _default_decoder.decode(s)
  File "/usr/local/lib/python3.7/json/decoder.py", line 337, in decode
    obj, end = self.raw_decode(s, idx=_w(s, 0).end())
  File "/usr/local/lib/python3.7/json/decoder.py", line 353, in raw_decode
    obj, end = self.scan_once(s, idx)
json.decoder.JSONDecodeError: Expecting ',' delimiter: line 1 column 10652 (char 10651)
```

We should probably have some explicit error handling that expects these sorts of errors and logs the entire message. These could easily be caused by non FAF traffic like automated portscans or something, but if they are legitimate messages, then maybe there is a bug in the client somewhere.

PR #761, opened 2021-04-03
- added more verbose exception to invalid JSON

xSchattenfluchx:
- fixes https://github.com/FAForever/server/issues/661#issue-702287230

### DIFF
--- a/server/protocol/qdatastream.py
+++ b/server/protocol/qdatastream.py
@@ -97,9 +97,12 @@ class QDataStreamProtocol(Protocol):
 
         try:
             message = json.loads(action)
-        except json.decoder.JSONDecodeError as err:
+        except json.decoder.JSONDecodeError as e:
             raise json.decoder.JSONDecodeError(
-                "Invalid JSON, full action string was: {}".format(action)) from err
+                msg = f"Invalid JSON, full action string was: \
+                    {action[e.pos-100:e.pos+100]}",
+                doc = e.doc,
+                pos = e.pos) from e
         try:
             for part in self.read_block(block):
                 try:
--- a/server/protocol/qdatastream.py
+++ b/server/protocol/qdatastream.py
@@ -95,7 +95,11 @@ class QDataStreamProtocol(Protocol):
         if action in ("PING", "PONG"):
             return {"command": action.lower()}
 
-        message = json.loads(action)
+        try:
+            message = json.loads(action)
+        except json.decoder.JSONDecodeError as err:
+            raise json.decoder.JSONDecodeError(
+                "Invalid JSON, full action string was: {}".format(action)) from err
         try:
             for part in self.read_block(block):
                 try:


server/protocol/qdatastream.py
```diff
@@ -95,7 +95,11 @@ async def read_message(self):
         if action in ("PING", "PONG"):
             return {"command": action.lower()}
 
-        message = json.loads(action)
+        try:
+            message = json.loads(action)
+        except json.decoder.JSONDecodeError as err:
+            raise json.decoder.JSONDecodeError(
```
Askaholic, line 101: Need to copy all the fields from the error.
```python3
>>> e = JSONDecodeError("foobar")
  File "<stdin>", line 1, in <module>
TypeError: __init__() missing 2 required positional arguments: 'doc' and 'pos'
```

server/protocol/qdatastream.py
```diff
@@ -95,7 +95,11 @@ async def read_message(self):
...
+                "Invalid JSON, full action string was: {}".format(action)) from err
```
Askaholic, line 102: Use f-string please, and I think we use `e` as the error name in other places so it would be good to do that here too.

Askaholic (changes requested):
Hey, thanks for taking a stab at this! 

One other thing that comes to mind is that sometimes the messages that fail can be quite long (like 30,000 characters), so  maybe it would make sense to only display the 100 characters on either side of the position where the error occurred (`e.pos`). Luckily python slicing handles this nicely even if the string is short:
```python3
>>> "foo"[2-100:2+100]
'foo'
```

server/protocol/qdatastream.py
xSchattenfluchx: sorry bout that, new commit in this PR fixes this

server/protocol/qdatastream.py
xSchattenfluchx: done and done

server/protocol/qdatastream.py
```diff
@@ -95,7 +95,14 @@ async def read_message(self):
...
 
-        message = json.loads(action)
+        try:
+            message = json.loads(action)
+        except json.decoder.JSONDecodeError as e:
+            raise json.decoder.JSONDecodeError(
+                msg = f"Invalid JSON, full action string was: \
+                    {action[e.pos-100:e.pos+100]}",
+                doc = e.doc,
+                pos = e.pos) from e
```
Askaholic, line 105: I'm surprised codacy isn't complaining about this but flake8 doesn't like the spaces around `=` here. Can we format it like this?
```suggestion
                msg=f"Invalid JSON, near: {action[e.pos-100:e.pos+100]}",
                doc=e.doc,
                pos=e.pos
            ) from e
```

Askaholic (changes requested):
Would you be able to add a unit test to `test_protocol.py`? It doesn't need to assert very much, maybe just that the function raises a `JSONDecodeError` and not some other type of error. I mostly just want this to be covered by the tests so that for instance forgetting to pass in all 3 parameters would trigger a test failure.

server/protocol/qdatastream.py
xSchattenfluchx: Will do, but the unit test will take some time, having some trouble with the dev environment.

Askaholic: Are you in Windows? If you message me on the faf discord I can try to help you out.

You can also just make a change, push, and look at the github actions output. Then rebase once everything works.

Askaholic:
Any progress on this?

Askaholic:
@xSchattenfluchx are you still working on this? If not I'm gonna close the PR.

PR #764, opened 2021-04-06
Remove references to equilibrium #744

### LINKED ISSUE #744: Remove references to equilibrium
Equilibrium was removed as a featured mod since it stopped being developed. We should remove all hardcoded references to it, there aren't very many.

We should also refactor this:
https://github.com/FAForever/server/blob/4a39a48966bf81fb1617506575de2e4eaa50d399/server/game_service.py#L158-L165

I think it would be best to get rid of the automatic game class detection and pass in the desired class when creating a game.

IanNaughton:
Remove references to equilibrium

I was able do three things as part of this change set.

- Remove Equilibrium from the `FeaturedModType` enum and all references
- Migrate game class detection out of the `game_service` class
- Update `game_service` tests to pass in a game class since the method signature technically changed

Please let me know if you have any questions or feedback for improvement!

Closes #744

### DIFF
--- a/server/game_service.py
+++ b/server/game_service.py
@@ -10,13 +10,10 @@ from .core import Service
 from .db import FAFDatabase
 from .decorators import with_logger
 from .games import (
-    CoopGame,
-    CustomGame,
     FeaturedMod,
     FeaturedModType,
     Game,
     GameState,
-    LadderGame,
     ValidityState,
     VisibilityState
 )
@@ -154,15 +151,6 @@ class GameService(Service):
             "matchmaker_queue_id": matchmaker_queue_id,
         }
         game_args.update(kwargs)
-
-        if not game_class:
-            game_class = {
-                FeaturedModType.LADDER_1V1:   LadderGame,
-                FeaturedModType.COOP:         CoopGame,
-                FeaturedModType.FAF:          CustomGame,
-                FeaturedModType.FAFBETA:      CustomGame,
-                FeaturedModType.EQUILIBRIUM:  CustomGame
-            }.get(game_mode, Game)
         game = game_class(**game_args)
 
         self._games[game_id] = game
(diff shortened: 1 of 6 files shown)


server/lobbyconnection.py
```diff
@@ -895,9 +903,17 @@ async def command_game_host(self, message):
         if rating_max is not None:
             rating_max = float(rating_max)
 
+        game_class = {
+            FeaturedModType.LADDER_1V1:   LadderGame,
+            FeaturedModType.COOP:         CoopGame,
+            FeaturedModType.FAF:          CustomGame,
+            FeaturedModType.FAFBETA:      CustomGame
+        }.get(game_mode, Game)
```
Askaholic, line 911: We only need to check for coop here, and default to `CustomGame`.

tests/unit_tests/test_games_service.py
```diff
@@ -15,6 +15,7 @@ async def test_create_game(players, game_service):
     game = game_service.create_game(
         visibility=VisibilityState.PUBLIC,
         game_mode="faf",
+        game_class=CustomGame,
```
Askaholic, line 18: I think you could make this the default argument.

server/game_service.py
```diff
@@ -154,15 +151,6 @@ def create_game(
             "matchmaker_queue_id": matchmaker_queue_id,
         }
         game_args.update(kwargs)
-
```
Askaholic, line 157: I would definitely either make `game_class` a required argument, or give it a sensible default. Defaulting to `None` just doesn't make any sense anymore.

Askaholic (changes requested):
Nice, looks pretty close!

server/lobbyconnection.py
Askaholic: Just use an if statement

Askaholic (approved):
Looks good!

PR #767, opened 2021-04-08
Issue/#560 Fix co-op leaderboard IntegrityError

### LINKED ISSUE #560: Coop leaderboard results are not resolved appropriately
In `handle_operation_complete` the leaderboard results are taken from the first player who reports them, and any further results throw a mysql error.

```
ERROR    Mar 22  04:33:05 GameConnection                 Something awful happened in a game thread!
  File "/code/server/gameconnection.py", line 192, in handle_action
    await COMMAND_HANDLERS[command](self, *args)
  File "/code/server/gameconnection.py", line 321, in handle_operation_complete
    (mission, self.game.id, secondary, delta, len(self.game.players))
  File "/src/aiomysql/aiomysql/sa/connection.py", line 145, in _execute
    await cursor.execute(query, dp or None)
  File "/src/aiomysql/aiomysql/cursors.py", line 239, in execute
    await self._query(query)
  File "/src/aiomysql/aiomysql/cursors.py", line 457, in _query
    await conn.query(q)
  File "/src/aiomysql/aiomysql/connection.py", line 428, in query
    await self._read_query_result(unbuffered=unbuffered)
  File "/src/aiomysql/aiomysql/connection.py", line 622, in _read_query_result
    await result.read()
  File "/src/aiomysql/aiomysql/connection.py", line 1105, in read
    first_packet = await self.connection._read_packet()
  File "/src/aiomysql/aiomysql/connection.py", line 593, in _read_packet
    packet.check_error()
  File "/usr/local/lib/python3.6/site-packages/pymysql/protocol.py", line 220, in check_error
    err.raise_mysql_exception(self._data)
  File "/usr/local/lib/python3.6/site-packages/pymysql/err.py", line 109, in raise_mysql_exception
    raise errorclass(errno, errval)
pymysql.err.IntegrityError: (1062, "Duplicate entry '11297472' for key 'gameuid'")
```

I'm also not sure what the `int(army) == 1` check is doing there.

eoinnoble:
Closes #560.

There's a `FIXME` in `handle_operation_complete` that I'd be happy to try to address, but I'm not immediately sure where we should be checking that the map file exists?

### DIFF
--- a/server/gameconnection.py
+++ b/server/gameconnection.py
@@ -322,7 +322,7 @@ class GameConnection(GpgNetServerProtocol):
         if self.game.validity != ValidityState.COOP_NOT_RANKED:
             return
 
-        secondary, delta = int(secondary), str(delta)
+        secondary, delta = secondary, str(delta)
         async with self._db.acquire() as conn:
             result = await conn.execute(
                 select([coop_map.c.id]).where(
--- a/server/gameconnection.py
+++ b/server/gameconnection.py
@@ -1,5 +1,6 @@
 import asyncio
 import contextlib
+import typing
 
 from sqlalchemy import select, text
 
@@ -294,14 +295,19 @@ class GameConnection(GpgNetServerProtocol):
         except (KeyError, ValueError):  # pragma: no cover
             self._logger.warning("Invalid result for %s reported: %s", army, result)
 
-    async def handle_operation_complete(self, army, secondary, delta):
-        if not int(army) == 1:
+    async def handle_operation_complete(
+        self,
+        primary_objectives_complete: typing.Union[bool, int],
+        secondary_objectives_complete: typing.Union[bool, int],
+        delta: str,
+    ) -> None:
+        if not primary_objectives_complete:
             return
 
         if self.game.validity != ValidityState.COOP_NOT_RANKED:
             return
 
-        secondary, delta = int(secondary), str(delta)
+        secondary, delta = int(secondary_objectives_complete), str(delta)
         async with self._db.acquire() as conn:
             # FIXME: Resolve used map earlier than this
             result = await conn.execute(
@@ -311,19 +317,29 @@ class GameConnection(GpgNetServerProtocol):
             )
             row = await result.fetchone()
             if not row:
-                self._logger.debug("can't find coop map: %s", self.game.map_file_path)
+                self._logger.debug(
+                    "can't find coop map: %s", self.game.map_file_path
+                )
                 return
             mission = row["id"]
 
-            await conn.execute(
-                coop_leaderboard.insert().values(
-                    mission=mission,
-                    gameuid=self.game.id,
-                    secondary=secondary,
-                    time=delta,
-                    player_count=len(self.game.players),
+            previously_completed_operations = await conn.execute(
+                select([coop_leaderboard]).where(
+                    coop_leaderboard.c.gameuid == self.game.id
                 )
             )
+            if await previously_completed_operations.fetchone() is None:
+                # Each player in a co-op game will send the OperationComplete
+                # message but we only need to perform this insert once
+                await conn.execute(
+                    coop_leaderboard.insert().values(
+                        mission=mission,
+                        gameuid=self.game.id,
+                        secondary=secondary,
+                        time=delta,
+                        player_count=len(self.game.players),
+                    )
+                )
 
     async def handle_json_stats(self, stats):
         self.game.report_army_stats(stats)
--- a/server/gameconnection.py
+++ b/server/gameconnection.py
@@ -304,7 +304,7 @@ class GameConnection(GpgNetServerProtocol):
         """
         :param primary: are primary mission objectives complete?
         :param secondary: are secondary mission objectives complete?
-        :param delta: the time at which the operation occurred
+        :param delta: the time it took to complete the mission
         :return: None
         """
         primary = FA.ENABLED == primary
@@ -320,9 +320,6 @@ class GameConnection(GpgNetServerProtocol):
             return
 
         if self.game.validity != ValidityState.COOP_NOT_RANKED:
-            self._logger.warning(
-                "OperationComplete called for invalid game: %s", self.game.id
-            )
             return
 
         secondary, delta = int(secondary), str(delta)
--- a/server/gameconnection.py
+++ b/server/gameconnection.py
@@ -1,6 +1,6 @@
 import asyncio
 import contextlib
-import typing
+from typing import Any
 
 from sqlalchemy import select, text
 
@@ -11,7 +11,8 @@ from .config import TRACE
 from .db.models import coop_leaderboard, coop_map, teamkills
 from .decorators import with_logger
 from .game_service import GameService
-from .games import Game, GameError, GameState, ValidityState, Victory
+from .games import CoopGame, Game, GameError, GameState, ValidityState, Victory
+from .games.typedefs import _FAFalse, _FATrue
 from .player_service import PlayerService
 from .players import Player, PlayerState
 from .protocol import DisconnectedError, GpgNetServerProtocol, Protocol
@@ -296,25 +297,27 @@ class GameConnection(GpgNetServerProtocol):
             self._logger.warning("Invalid result for %s reported: %s", army, result)
 
     async def handle_operation_complete(
-        self,
-        primary_objectives_complete: typing.Any,
-        secondary_objectives_complete: typing.Any,
-        delta: str,
+        self, primary: Any, secondary: Any, delta: str
     ) -> None:
+        """`primary` and `secondary` denote whether or not the primary and
+        secondary mission objectives have been completed"""
+        primary = _FATrue() == primary
+        secondary = _FATrue() == secondary
 
-        def to_bool(x): return x in [1, "1", True, "True", "true"]
-        primary_objectives_complete = to_bool(primary_objectives_complete)
-        secondary_objectives_complete = to_bool(secondary_objectives_complete)
-
-        if not primary_objectives_complete:
+        if not primary:
             return
 
-        if self.game.validity != ValidityState.COOP_NOT_RANKED:
+        if (
+            not isinstance(self.game, CoopGame)
+            or self.game.validity != ValidityState.COOP_NOT_RANKED
+        ):
+            self._logger.debug(
+                "OperationComplete called for non-coop game: %s", self.game.id
+            )
             return
 
-        secondary, delta = int(secondary_objectives_complete), str(delta)
+        secondary, delta = int(secondary), str(delta)
         async with self._db.acquire() as conn:
-            # FIXME: Resolve used map earlier than this
             result = await conn.execute(
                 select([coop_map.c.id]).where(
                     coop_map.c.filename == self.game.map_file_path
@@ -330,8 +333,7 @@ class GameConnection(GpgNetServerProtocol):
 
             # Each player in a co-op game will send the OperationComplete
             # message but we only need to perform this insert once
-            already_saved = getattr(self.game, '_coop_leaderboard_saved', False)
-            if not already_saved:
+            if not self.game.leaderboard_saved:
                 await conn.execute(
                     coop_leaderboard.insert().values(
                         mission=mission,
@@ -341,7 +343,7 @@ class GameConnection(GpgNetServerProtocol):
                         player_count=len(self.game.players),
                     )
                 )
-                self.game._coop_leaderboard_saved = True
+                self.game.leaderboard_saved = True
 
     async def handle_json_stats(self, stats):
         self.game.report_army_stats(stats)
--- a/server/gameconnection.py
+++ b/server/gameconnection.py
@@ -1,6 +1,6 @@
 import asyncio
 import contextlib
-from typing import Any
+from typing import Any, List
 
 from sqlalchemy import select, text
 
@@ -12,7 +12,7 @@ from .db.models import coop_leaderboard, coop_map, teamkills
 from .decorators import with_logger
 from .game_service import GameService
 from .games import CoopGame, Game, GameError, GameState, ValidityState, Victory
-from .games.typedefs import _FATrue
+from .games.typedefs import FA
 from .player_service import PlayerService
 from .players import Player, PlayerState
 from .protocol import DisconnectedError, GpgNetServerProtocol, Protocol
@@ -191,7 +191,7 @@ class GameConnection(GpgNetServerProtocol):
                     offer=False
                 )
 
-    async def handle_action(self, command, args):
+    async def handle_action(self, command: str, args: List[Any]):
         """
         Handle GpgNetSend messages, wrapped in the JSON protocol
         :param command: command type
@@ -216,7 +216,7 @@ class GameConnection(GpgNetServerProtocol):
     async def handle_desync(self, *_args):  # pragma: no cover
         self.game.desyncs += 1
 
-    async def handle_game_option(self, key, value):
+    async def handle_game_option(self, key: str, value: Any):
         if not self.is_host():
             return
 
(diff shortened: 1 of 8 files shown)


server/gameconnection.py
```diff
@@ -294,14 +295,19 @@ async def handle_game_result(self, army, result):
         except (KeyError, ValueError):  # pragma: no cover
             self._logger.warning("Invalid result for %s reported: %s", army, result)
 
-    async def handle_operation_complete(self, army, secondary, delta):
-        if not int(army) == 1:
+    async def handle_operation_complete(
+        self,
+        primary_objectives_complete: typing.Union[bool, int],
+        secondary_objectives_complete: typing.Union[bool, int],
```
Askaholic, line 301: So, the weird thing here is that the correct type is actually `str` for all parameters because that's how the message is serialized by the game afaik. That's also the reason why we have to call `int`, `bool`, etc on the values before we use them.

server/gameconnection.py
```diff
@@ -294,14 +295,19 @@ async def handle_game_result(self, army, result):
...
+        delta: str,
+    ) -> None:
+        if not primary_objectives_complete:
```
Askaholic, line 304: the var is actually a string, so this will only reject `""` which is not what we want.
```suggestion
        if not bool(primary_objectives_complete):
```

server/gameconnection.py
```diff
@@ -311,19 +317,29 @@ async def handle_operation_complete(self, army, secondary, delta):
...
             mission = row["id"]
 
-            await conn.execute(
-                coop_leaderboard.insert().values(
-                    mission=mission,
-                    gameuid=self.game.id,
-                    secondary=secondary,
-                    time=delta,
-                    player_count=len(self.game.players),
+            previously_completed_operations = await conn.execute(
```
Askaholic, line 326: No reason to do a database query here, just use a boolean value on the `self.game` to keep track of whether or not the entry has been written yet.

tests/unit_tests/test_gameconnection.py
```diff
@@ -438,56 +442,119 @@ async def test_handle_action_GameEnded_ends_game(
...
     ugame.map_file_path = "maps/prothyon16.v0005.zip"
     ugame.validity = ValidityState.COOP_NOT_RANKED
     game_connection.game = ugame
-
-    secondary = 1
     time_taken = "09:08:07.654321"
-    await game_connection.handle_action("OperationComplete", ["1", secondary, time_taken])
+
+    await game_connection.handle_action(
+        "OperationComplete", [True, True, time_taken]
```
Askaholic, line 456: the args here should be strings as sent by the game. The lua serializer that the game uses is notoriously weird about serializing booleans. We should be able to handle any of the following inputs to this function for booleans:
1. `True`/`False`
2. `"true"`/`"false"`
3. `1`/`0`

tests/unit_tests/test_gameconnection.py
```diff
@@ -438,56 +442,119 @@ async def test_handle_action_GameEnded_ends_game(
...
+        caplog.clear()
+        await game_connection.handle_action(
+            "OperationComplete", [True, True, time_taken]
+        )
+
+        assert not any(
+            record.exc_info
+            and isinstance(record.exc_info[1], pymysql.err.IntegrityError)
+            for record in caplog.records
+        )
```
Askaholic, line 557: Cool, haven't seen this before.

Askaholic (changes requested):
Huh, I've never seen that error before in the tests. Maybe something weird going on with GitHub actions. Maybe it would be possible to have the control server bind to `localhost` during tests instead of the public interface. Not sure if that would make a difference though. Probably they will succeed on the next run.

Honestly I think that comment with the `FIXME` can be deleted. That has been there since before I started working on the server, but I think it's actually fine to resolve the file path there. I think custom games do it in a similar place (at game end). The alternative would be to do it at game start time, but as far as I know there are no problems that would be solved by moving the map resolution there. So I think it's best not to mess with it and leave it how it is.

server/gameconnection.py
eoinnoble: Ah thanks, I wasn't sure how this was being serialized.

server/gameconnection.py
eoinnoble: Oh nice idea, thank you!

tests/unit_tests/test_gameconnection.py
eoinnoble: Ok, I'll try to test that.

server/gameconnection.py
eoinnoble: `bool` seems like a bad idea if we can get boolean strings too?

```python
>>> bool("false") is True
True
```

tests/unit_tests/test_gameconnection.py
eoinnoble: If you mean `caplog`, neither had I!

server/gameconnection.py
Askaholic: You could do something similar to this, or maybe just use this class: 
https://github.com/FAForever/server/blob/aa535a92b676525f67223be8035fb132ca106f60/server/games/typedefs.py#L214-L218

tests/unit_tests/test_gameconnection.py
Askaholic: Well I knew about caplog, but I didn't know you could use it to check the `exc_info` and get the exception type and stuff.

server/gameconnection.py
```diff
@@ -311,19 +322,26 @@ async def handle_operation_complete(self, army, secondary, delta):
...
-            await conn.execute(
-                coop_leaderboard.insert().values(
-                    mission=mission,
-                    gameuid=self.game.id,
-                    secondary=secondary,
-                    time=delta,
-                    player_count=len(self.game.players),
+            # Each player in a co-op game will send the OperationComplete
+            # message but we only need to perform this insert once
+            already_saved = getattr(self.game, '_coop_leaderboard_saved', False)
```
Askaholic, line 333: IMO, just make this an instance variable on `CoopGame`. Which brings up a point: what if we see an `OperationComplete` message on a game that is not a `CoopGame`? I think that either means there's a bug and the server didn't use the right game class, or the game is doing something really weird probably due to mods. I think either way though, probably in that case no leaderboard entry should be created and a warning should be logged.

server/gameconnection.py
```diff
@@ -294,14 +295,24 @@ async def handle_game_result(self, army, result):
         except (KeyError, ValueError):  # pragma: no cover
             self._logger.warning("Invalid result for %s reported: %s", army, result)
 
-    async def handle_operation_complete(self, army, secondary, delta):
-        if not int(army) == 1:
+    async def handle_operation_complete(
+        self,
+        primary_objectives_complete: typing.Any,
+        secondary_objectives_complete: typing.Any,
```
Askaholic, line 301: Is there really a point in using `Any` here? Omitting the type signature would have the same effect I think. Also the rest of the codebase imports the needed types from `typing` and uses them without the `typing.` prefix, so lets at least keep that consistent.

Askaholic, line 301: These variable names seem a bit excessively long to me. Maybe we can call them `primary` and `secondary` and add a description of what they mean in a docstring? I know theres a point about self documenting code, but using 34% of available line length for 1 variable name seems pretty hefty to me:
```python3
>>> len("primary_objectives_complete")
27
>>> 27/79
0.34177215189873417

>>> len("secondary_objectives_complete")/79
0.3670886075949367
```

tests/unit_tests/test_gameconnection.py
eoinnoble: Ah yes, good old `LogRecord`

server/gameconnection.py
eoinnoble: Isn't that what `if self.game.validity != ValidityState.COOP_NOT_RANKED:` is checking?

server/gameconnection.py
eoinnoble: Given that the function could end up getting a variety of inputs here I personally think it's a useful hint to future developers that they can't assume uniform input types?

server/gameconnection.py
Askaholic: Hm, I guess that would do it yea. Although it's more of a happy accident :P.

server/gameconnection.py
Askaholic: Maybe. If you're gonna do it for one `handle_*` function you should probably do it for all of them though for consistency.

server/gameconnection.py
```diff
@@ -294,36 +296,54 @@ async def handle_game_result(self, army, result):
         except (KeyError, ValueError):  # pragma: no cover
             self._logger.warning("Invalid result for %s reported: %s", army, result)
 
-    async def handle_operation_complete(self, army, secondary, delta):
-        if not int(army) == 1:
+    async def handle_operation_complete(
+        self, primary: Any, secondary: Any, delta: str
+    ) -> None:
+        """`primary` and `secondary` denote whether or not the primary and
+        secondary mission objectives have been completed"""
```
Askaholic, line 303: Use the `:param:` notation (see handle_teamkill_happened). In my documentation PR I'm redoing the docstrings to use markdown instead of rst, but just keep it consistent for now. I'll fix it later.

server/gameconnection.py
```diff
@@ -294,36 +296,54 @@ async def handle_game_result(self, army, result):
         except (KeyError, ValueError):  # pragma: no cover
             self._logger.warning("Invalid result for %s reported: %s", army, result)
 
-    async def handle_operation_complete(self, army, secondary, delta):
-        if not int(army) == 1:
+    async def handle_operation_complete(
+        self, primary: Any, secondary: Any, delta: str
```
Askaholic, line 300: I think I said these would be strings before, but after looking at the code again, I realized they can be any json value. (Have a look at `handle_action` and `LobbyConnection.on_message_received`). The GPGNet messages look like this:

```json
{
    "command": "OperationComplete",
    "target": "game",
    "args": [1, 1, "12:34"]
}
```
The parameters to these `handle_*` functions are just the args array from the message, so the type can be anything produced by `json.loads`. Which is also the reason why you'll see lots of calls to `str` and `int` on those arguments in handler functions.

eoinnoble: As someone coming from the world of django-rest-framework it might be nice to pass these messages through some kind of standard serializer so that the data is always ready for the database rather than each handler having to do this?

server/gameconnection.py
```diff
@@ -294,36 +296,54 @@ async def handle_game_result(self, army, result):
...
+    ) -> None:
+        """`primary` and `secondary` denote whether or not the primary and
+        secondary mission objectives have been completed"""
+        primary = _FATrue() == primary
+        secondary = _FATrue() == secondary
```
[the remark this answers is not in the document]
eoinnoble: I like that!

server/gameconnection.py
```diff
@@ -294,36 +296,54 @@ async def handle_game_result(self, army, result):
...
             return
 
-        if self.game.validity != ValidityState.COOP_NOT_RANKED:
+        if (
+            not isinstance(self.game, CoopGame)
+            or self.game.validity != ValidityState.COOP_NOT_RANKED
+        ):
+            self._logger.debug(
+                "OperationComplete called for non-coop game: %s", self.game.id
+            )
```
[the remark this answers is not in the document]
eoinnoble: Maybe separate log statements for each case then.

server/gameconnection.py
Askaholic: I'm not entirely sure what you mean. The messages are already deserialized by the json parser. But the message format is essentially a remote function call, you can see it just sends an arg list. The most standard way to automatically convert types would probably be to deserialize to named tuples instead and define the types that way, but then you have to define a named tuple for every handler function which seems bad.

I suppose we could write some decorator or something that inspects the type arguments of the handler function and calls appropriate parsing functions. That would work pretty well for primitives I think, in which case you could just call the parameter type like: `parameter.type(arg)` e.g. `str(arg)` but if the params are tuples or something, then it would be much more annoying logic to write. You'd have to handle each case separately I think.

Seems like a lot of work, but it would be pretty cool.

server/gameconnection.py
```diff
@@ -294,44 +298,76 @@ async def handle_game_result(self, army, result):
...
 
-    async def handle_operation_complete(self, army, secondary, delta):
-        if not int(army) == 1:
+    async def handle_operation_complete(
+        self, primary: Any, secondary: Any, delta: str
+    ):
+        """
+        :param primary: are primary mission objectives complete?
+        :param secondary: are secondary mission objectives complete?
+        :param delta: the time at which the operation occurred
```
Askaholic, line 307: This would be how long it took to complete the mission. That's why whoever named the parameter called it `delta`, because it's an amount of time, not a timestamp.

server/gameconnection.py
```diff
@@ -294,44 +298,76 @@ async def handle_game_result(self, army, result):
...
+            return
+
+        if not isinstance(self.game, CoopGame):
+            self._logger.warning(
+                "OperationComplete called for non-coop game: %s", self.game.id
+            )
             return
 
         if self.game.validity != ValidityState.COOP_NOT_RANKED:
+            self._logger.warning(
```
Askaholic, line 323: I wouldn't log at all here. It's perfectly normal for people to play unranked games. For coop this could be as simple as setting the difficulty to normal or easy. I consider warnings to be more like configuration errors, or like above, code states that shouldn't be possible but aren't fatal.

Askaholic (approved):
Minor things. Code wise it looks pretty good. I'll put it up on the test server and we can try it out if you have some time.

tests/integration_tests/test_matchmaker.py
```diff
@@ -264,7 +264,7 @@ async def test_game_matchmaking_disconnect(lobby_server):
     assert msg == {"command": "match_cancelled"}
 
 
-@fast_forward(130)
+@fast_forward(131)
```
Askaholic, line 267: What's the point of this change?

eoinnoble: This test was timing out so I increased the timeout to see if that would help it pass and it did.

Askaholic: This isn't a timeout. It just fast forwards the eventloop by the given number of seconds. So the for example the tests don't need to wait 30 real seconds on a `sleep(30)`. They will wait 30 event loop seconds which will be ticking by really quickly. If something was timing out before and now it's not then that sounds like a flaky test to me. These numbers should have no effect on the test outcome.

Askaholic: IMO we should leave revert this change.

server/gameconnection.py
```diff
@@ -294,44 +298,73 @@ async def handle_game_result(self, army, result):
...
+        if not isinstance(self.game, CoopGame):
+            self._logger.warning(
+                "OperationComplete called for non-coop game: %s", self.game.id
+            )
             return
 
         if self.game.validity != ValidityState.COOP_NOT_RANKED:
             return
 
         secondary, delta = int(secondary), str(delta)
```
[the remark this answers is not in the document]
eoinnoble: Will the db coerce a boolean into an integer for us?

tests/unit_tests/test_gameconnection.py
```diff
@@ -438,56 +443,143 @@ async def test_handle_action_GameEnded_ends_game(
...
+                [coop_leaderboard.c.secondary, coop_leaderboard.c.gameuid]
+            ).where(coop_leaderboard.c.gameuid == coop_game.id),
+        )
 
-    secondary = 1
+        row = await result.fetchone()
+        assert (row[0], row[1]) == (1, coop_game.id)
+
+
+@pytest.mark.parametrize("primary", [0, False, "False", "false"])
```
Askaholic, line 477: Don't wanna parametrize secondary here?

eoinnoble: The message handler bails before it does anything with `secondary` if `primary` is falsy, so there didn't seem much point?

PR #775, opened 2021-04-17
Issue/#655 add mq to ci

Askaholic:
Moves the rabbitmq tests back with the normal "integration_tests" and deselects them when running through pipenv by registering them under the `rabbitmq` mark.

Adds rabbitmq service to github actions and runs rabbitmq tests


.github/workflows/test.yml
```diff
@@ -48,6 +48,22 @@ jobs:
           MYSQL_USER: faf-python-server
           MYSQL_PASSWORD: banana
 
+      faf-rabbitmq:
+        image: rabbitmq:3-management-alpine
```
Askaholic, line 52: Actually not sure if this needs the management plugin. I think probably not since that's just a web interface.

.github/workflows/scripts/init-rabbitmq.sh
```diff
@@ -0,0 +1,15 @@
+#!/usr/bin/env bash
+
+MAX_WAIT=60 # max. 1 minute waiting time in loop before timeout
```
cleborys, line 3: or `MAX_WAIT_SECONDS` and then it doesn't need a comment (I know it's just copied 😅 )

cleborys (approved):
Looks good to me, very good idea to mark the tests!

.github/workflows/scripts/init-rabbitmq.sh
Askaholic: And I think the original code was copied from faf stack lol.

PR #753, opened 2021-03-30
Issue/#482 improve documentation

Askaholic:
Not quite done with this.

- Rewrite param docs to use markdown instead of rst
- Add doctests to pytest
- Remove `server.abc`

#482

PR #776, opened 2021-04-20
[WIP] Add game_type field to game_launch command message

### LINKED ISSUE #685: Use game_type instead of init_mode in game_launch message

wrharding:
Closes #685

I've been getting these note's all night while running tests, looks like I still have some network configuration to fix but the DB is online:

.faf-db                   | 2021-04-20T00:57:20.848527Z 633 [Note] Aborted connection 633 to db: 'faf' user: 'root' host: '172.21.0.1' (Got an error reading communication packets)

### DIFF
--- a/server/lobbyconnection.py
+++ b/server/lobbyconnection.py
@@ -39,6 +39,7 @@ from .games import (
     GameState,
     VisibilityState
 )
+from .games.typedefs import GameType
 from .geoip_service import GeoIpService
 from .ice_servers.coturn import CoturnHMAC
 from .ice_servers.nts import TwilioNTS
@@ -953,6 +954,8 @@ class LobbyConnection:
             # options are. Currently, options for ladder are hardcoded into the
             # client.
             "name": game.name,
+            "game_type": GameType.to_string(game.game_type),
+            # Use game_type instead, init_mode is depreciated
             "init_mode": game.init_mode.value,
             "rating_type": game.rating_type,
             **options._asdict()
--- a/tests/unit_tests/test_lobbyconnection.py
+++ b/tests/unit_tests/test_lobbyconnection.py
@@ -389,7 +389,6 @@ async def test_should_reply_with_game_type_field_on_game_lauch(database,
     game.id = 42
     game.name = "Test Game Name"
     game_service._games[42] = game
-    lobbyconnection.player = players.hosting
     test_game_info["uid"] = 42
 
     await lobbyconnection.on_message_received({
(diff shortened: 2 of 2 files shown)


server/lobbyconnection.py
```diff
@@ -953,6 +954,8 @@ async def launch_game(
             # options are. Currently, options for ladder are hardcoded into the
             # client.
             "name": game.name,
+            "game_type": GameType.to_string(game.game_type),
```
Askaholic, line 957: Could we take the opportunity to refactor this `to_string` stuff? `Enum`s have some nice support for this already with the `__members__` dictionary and being able to simply call the constructor with the enum value. See the `Faction` enum in `server/factions.py`

```python3
>>> from enum import Enum
>>> class Foo(Enum):
...     bar = 1
...     baz = 2
... 
>>> Foo(1)
<Foo.bar: 1>
>>> Foo(1).name
'bar'
>>> Foo(1).value
1
>>> Foo.__members__
mappingproxy({'bar': <Foo.bar: 1>, 'baz': <Foo.baz: 2>})
```

tests/unit_tests/test_lobbyconnection.py
```diff
@@ -369,6 +370,40 @@ async def test_command_game_join_uid_as_str(
         "mod": "faf",
         "uid": 42,
         "name": "Test Game Name",
+        "game_type": "custom",
+        "init_mode": InitMode.NORMAL_LOBBY.value,
+        "rating_type": "global",
+    }
+    lobbyconnection.send.assert_called_with(expected_reply)
+
+async def test_should_reply_with_game_type_field_on_game_lauch(database,
```
Askaholic, line 379: Can we format like this please:
```python3
def long_function_name(
    arg1,
    arg2,
    ...
):
    pass
```

Askaholic: Isn't this testing exactly the same thing as `test_command_game_join_calls_join_game`? I think it would be more useful to add some tests for the other game types: coop and matchmaker. Maybe it would also be good to add a test of the "integration test" variety in `tests/integration_tests/test_game.py`. I don't think we have one yet that actually checks the contents of the game_launch message. I'd also modify `tests/integration_tests/test_matchmaker.py::test_game_launch_message` to check the game type.

Askaholic (commented):
So you are getting that error, but the tests are still able to run and access the database? @Sheikah45 Do you know anything about that error?

Also, you'll need to rebase the PR to get the latest changes from develop. There are no conflicts so this should be easy:
```
$ git remote add faf https://github.com/FAForever/server.git
$ git checkout develop
$ git pull faf develop
$ git checkout -
$ git rebase develop
$ git push --force
```

Sheikah45:
I unfortunately have not run across that error before

server/lobbyconnection.py
Askaholic: Actually it turns out I did this already in #753. So just rebase again and the changes will be there.

Askaholic:
@wrharding any progress on this?

Askaholic:
Are you going to have a chance to look at this again? If not I'll close the PR

PR #779, opened 2021-04-21
Issue/#771 send broadcast info to rabbitmq

### LINKED ISSUE #771: Add broadcasted info updates to RabbitMQ
There are currently 3 types of messages sent to every client as a broadcasted state update:

- player_info
- game_info
- matchmaker_info

This is handled here: 
https://github.com/FAForever/server/blob/aa535a92b676525f67223be8035fb132ca106f60/server/__init__.py#L141-L142

We should post these updates to a RabbitMQ exchange so that other services can subscribe to them as well. Additionally each `ServerContext` should get a boolean attribute signaling whether it will broadcast updates or not. This would allow us to keep backwards compatible behavior while also handling broadcasts differently for newer clients that support it.

### Use cases
1. Lobby bridge
The lobby bridge idea is to break out some of the network handling into a separate more performant application. This application would proxy connections to the lobby server and would handle the broadcasting of updates. This is desirable because the update broadcasts are very slow in python and are the number 1 cause of CPU load for the lobby server.
2. Galactic War backend
I guess the gw-backend server might want to subscribe to game information.
3. API endpoint
Maybe this could replace or reduce queries to the control server currently used by the API for displaying number of online players and number of active games on the website.

Askaholic:
I refactored the broadcasting logic to its own service which now also posts the broadcast messages to rabbitmq. I stuck with posting them to the same exchange as the `gameResults` for now, but I'd be interested to hear other peoples opinions on whether that makes sense or it should get its own exchange. The 3 part routing key is a little weird since there will never be any success states other than "success" and the CRUD operation will always be "update" and there is no real clear way to specify that the message is a broadcast.

I also decided to aggregate all of the game info into a single message just like it is sent during login. Maybe we should take the opportunity to differentiate between `"game_info"` messages that represent a single game and `"game_info"` messages that have a `"games"` sub-list. Maybe call it `"game_list_info"`, or just get rid of the single game version entirely. I think getting rid of the single game version makes a lot of sense as it would maintain consistency with how the other broadcasted info messages look.

Closes #771

### DIFF
--- a/server/broadcast_service.py
+++ b/server/broadcast_service.py
@@ -48,6 +48,12 @@ class BroadcastService(Service):
         self._logger.debug("Broadcast service initialized")
 
     async def report_dirties(self):
+        """
+        Send updates about any dirty (changed) entities to connected players.
+        This function is called at a fixed interval, which guarantees that any
+        given object will not be written out to the clients more than once per
+        interval.
+        """
         self.game_service.update_active_game_metrics()
         dirty_games = self.game_service.pop_dirty_games()
         dirty_queues = self.game_service.pop_dirty_queues()
--- a/server/broadcast_service.py
+++ b/server/broadcast_service.py
@@ -65,10 +65,7 @@ class BroadcastService(Service):
                 "command": "player_info",
                 "players": [player.to_dict() for player in dirty_players]
             }
-            self.server.write_broadcast(
-                player_info,
-                lambda lobby_conn: lobby_conn.authenticated
-            )
+            self.server.write_broadcast(player_info)
 
         game_info = {
             "command": "game_info",
--- a/server/broadcast_service.py
+++ b/server/broadcast_service.py
@@ -49,11 +49,9 @@ class BroadcastService(Service):
 
     async def report_dirties(self):
         self.game_service.update_active_game_metrics()
-        dirty_games = self.game_service.dirty_games
-        dirty_queues = self.game_service.dirty_queues
-        dirty_players = self.player_service.dirty_players
-        self.game_service.clear_dirty()
-        self.player_service.clear_dirty()
+        dirty_games = self.game_service.pop_dirty_games()
+        dirty_queues = self.game_service.pop_dirty_queues()
+        dirty_players = self.player_service.pop_dirty_players()
 
         if dirty_queues:
             matchmaker_info = {
--- a/server/broadcast_service.py
+++ b/server/broadcast_service.py
@@ -1,3 +1,5 @@
+from aio_pika import DeliveryMode
+
 from .config import config
 from .core import Service
 from .decorators import with_logger
@@ -45,7 +47,7 @@ class BroadcastService(Service):
         )
         self._logger.debug("Broadcast service initialized")
 
-    def report_dirties(self):
+    async def report_dirties(self):
         self.game_service.update_active_game_metrics()
         dirty_games = self.game_service.dirty_games
         dirty_queues = self.game_service.dirty_queues
@@ -54,20 +56,26 @@ class BroadcastService(Service):
         self.player_service.clear_dirty()
 
         if dirty_queues:
-            self.server.write_broadcast({
+            matchmaker_info = {
                 "command": "matchmaker_info",
                 "queues": [queue.to_dict() for queue in dirty_queues]
-            })
+            }
+            self.server.write_broadcast(matchmaker_info)
 
         if dirty_players:
+            player_info = {
+                "command": "player_info",
+                "players": [player.to_dict() for player in dirty_players]
+            }
             self.server.write_broadcast(
-                {
-                    "command": "player_info",
-                    "players": [player.to_dict() for player in dirty_players]
-                },
+                player_info,
                 lambda lobby_conn: lobby_conn.authenticated
             )
 
+        game_info = {
+            "command": "game_info",
+            "games": []
+        }
         # TODO: This spams squillions of messages: we should implement per-
         # connection message aggregation at the next abstraction layer down :P
         for game in dirty_games:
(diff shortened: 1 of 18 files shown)


Brutus5000 (commented):
I don't understand much about the structure of the Python server.
So my first question is: Why do you need a "BroadcastService" and does it really do what the name implies?

It looks like the broadcast service is not only responsible for broadcasting but also responsible for fetching the dirty games and also clearing them. For me this looks like a violation of separation of concerns.

If the broadcast service follows it's name it should be responsible for broadcasting messages and the services calling it are responsible for delivering the data. If that's not the case, maybe a different name would be better.

In any scenario the gameService (?) should be responsible for clearing it's own dirtylist.
Whether it pushes the list before cleaning it to the broadcast service or whether it offers a "transactional" method for querying **and** cleaning the dirty games list to be called by the broadcast service doesn't matter.

Askaholic:
> Why do you need a "BroadcastService"

Mostly this is a refactor of the broadcast loop that used to be part of `ServerInstance`. It didn't really feel right to keep adding to the code where it was.

> whether it offers a "transactional" method for querying **and** cleaning the dirty games list to be called by the broadcast service

I could do that. That's probably nicer than accessing the member variables directly. There's also this bit of code which IMO doesn't belong there either
```python3
if game.state == GameState.ENDED:
    self.game_service.remove_game(game)
```
but I'd rather not mess with refactoring that ATM since it's out of the scope of this PR and getting it wrong could cause all sorts of horrible stuff like ghost games.

server/broadcast_service.py
```diff
@@ -0,0 +1,120 @@
...
+            start=True
+        )
+        self._broadcast_ping_timer = LazyIntervalTimer(
+            lambda: config.PING_INTERVAL,
+            self.broadcast_ping,
+            start=True
+        )
+        self._logger.debug("Broadcast service initialized")
+
+    async def report_dirties(self):
```
eoinnoble, line 50: As someone new to the code base I'd love a brief docstring here.

server/broadcast_service.py
```diff
@@ -0,0 +1,120 @@
...
+            self.server.write_broadcast(matchmaker_info)
+
+        if dirty_players:
+            player_info = {
+                "command": "player_info",
+                "players": [player.to_dict() for player in dirty_players]
+            }
+            self.server.write_broadcast(
+                player_info,
+                lambda lobby_conn: lobby_conn.authenticated
```
eoinnoble, line 70:
```suggestion
                lambda conn: conn.authenticated
```
More consistent with how it's used later in this function.

server/game_service.py
```diff
@@ -105,24 +104,24 @@ async def update_data(self):
...
-        return self._dirty_queues
-
     def mark_dirty(self, obj: Union[Game, MatchmakerQueue]):
         if isinstance(obj, Game):
             self._dirty_games.add(obj)
         elif isinstance(obj, MatchmakerQueue):
             self._dirty_queues.add(obj)
 
-    def clear_dirty(self):
+    def get_new_dirty_games(self) -> Set[Game]:
```
eoinnoble, line 113: `get` here doesn't necessarily communicate the side effect of resetting `self._dirty_games`, which might end up surprising someone? I'm not sure what would be better, `dump`, `pop`, `get_and_reset_dirty_*`?

server/timing/timer.py
```diff
@@ -87,11 +87,35 @@ def __call__(self, func):
...
     def __str__(self):
-        return f"{self.interval} {self.func}"
+        return f"{self.get_delay()} {self.func}"
 
     def __repr__(self):
         return f"<Timer {str(self)}>"
 
 
+class LazyIntervalTimer(Timer):
+    """A timer that calls a funciton to get the next interval"""
```
eoinnoble, line 97:
```suggestion
    """A timer that calls a function to get the next interval"""
```

tests/integration_tests/conftest.py
```diff
@@ -45,8 +50,19 @@ async def party_service(game_service):
...
+    game_service,
+    player_service,
+):
+    # The reference to the ServerInstance needs to be established later
+    service = BroadcastService(
+        None,
+        message_queue_service,
+        game_service,
+        player_service,
+    )
```
eoinnoble, line 65:
```suggestion
    service = BroadcastService(
        None,  # Reference to the Server instance needs to be established later
        message_queue_service,
        game_service,
        player_service,
    )
```

tests/integration_tests/conftest.py
```diff
@@ -260,3 +285,72 @@ async def connect_and_sign_in(
...
+    return proto
+
+
+class AioQueueProtocol(Protocol):
+    """
+    A wrapper around an asyncio `Queue` that exposes the `Protocol` interface.
+    """
+
+    def __init__(self, queue):
+        super().__init__(mock.Mock(), mock.Mock())
```
eoinnoble, line 332: Is it work autospeccing anything for either of these mocks?

eoinnoble (approved):
This is way over my head in most respects but certainly nothing glaringly wrong from what I can see.

server/game_service.py
Askaholic: I wasn't really sure either. That's why I called it `get_new` instead of just `get`. Not really a fan of `get_and` though. Maybe `pop` is better.

tests/integration_tests/conftest.py
Askaholic: I don't think so. Now that I think of it though. I don't think I need to call `super` here at all, I think the super call is totally optional in python, and I don't need any of the base class behavior here.

eoinnoble: It is optional, yep.

0x647262 (approved):
LGTM!

server/broadcast_service.py
Askaholic: How about that?

PR #780, opened 2021-04-21
Issue/#777 remove games that shouldn't be visible, from the game list and add tests

### LINKED ISSUE #777: Game list sent on login sends all games, not just visible ones
# Problem
When a player logs in they are sent a list of all open games here:
https://github.com/FAForever/server/blob/4dd4c92867f8630497eb305287681b1c2473ca7b/server/lobbyconnection.py#L229-L233

However, there are certain visibility rules that are followed in consequent broadcasted game updates here:
https://github.com/FAForever/server/blob/4dd4c92867f8630497eb305287681b1c2473ca7b/server/__init__.py#L174-L180

This causes 2 problems:
1. When a player logs in they can see friend only games for people who don't have them as friends, and they can see games of people who have foed them.
2. They will not receive any updates about those games causing them to appear as "ghost lobbies" that never go away.

# Solution
Fixing it should be as simple as adding a filter condition that calls `is_visible_to_player` to the list comprehension in `send_game_list`. 

## Tests
We should add at least the following tests:
1. A test in `tests/integration_tests/test_server.py` that verifies that players can't see foes games and can't see games outside of their rating range when logging in.
2. A test in `tests/integration_tests/test_matchmaker.py` that verifies that players can still launch matchmaker games against foes.
3. A test in `tests/unit_tests/test_lobbyconnection.py` that verifies the behavior of `send_game_list`.

CallMeMisterOwl:
Adda simple check to remove games on login that shouldn't be visible, from the game list and also add tests like mentioned here #777.
This should close #777.


server/lobbyconnection.py
```diff
@@ -229,7 +229,8 @@ async def command_matchmaker_info(self, message):
     async def send_game_list(self):
         await self.send({
             "command": "game_info",
-            "games": [game.to_dict() for game in self.game_service.open_games]
+            "games": [game.to_dict() if game.is_visible_to_player(self.player) else None
+                      for game in self.game_service.open_games]
```
Askaholic, line 233: The filter condition should come after the `for x in y` part, then you don't need to include `None` objects:

```python3
>>> my_list = [1, 123, 5, 20, 3, 10, 11, 7]

>>> [item for item in my_list if item < 10]
[1, 5, 3, 7]

>>> [item if item < 10 else None for item in my_list]
[1, None, 5, None, 3, None, None, 7]
```

Askaholic (changes requested):
Hey, thanks for taking a stab at this!

The real work that needs to be done here is in writing some good tests. Currently the tests pass, yet there is a bug, therefore the tests are insufficient. What we really need are some tests that would fail if you were to undo the fix. Something that if it had existed a year ago, we would have known about the bug then. I already suggested several tests on the issue ticket.

Also can name the PR with a short description of the bug its fixing? See the other open PR's for examples.

CallMeMisterOwl:
> The real work that needs to be done here is in writing some good tests. Currently the tests pass, yet there is a bug, therefore the tests are insufficient.

Right. I misread the issue. I'll work on that ASAP.

> Also can name the PR with a short description of the bug its fixing? See the other open PR's for examples.

Sure

server/lobbyconnection.py
CallMeMisterOwl: Alright, that makes sense, though the python syntax is a bit confusing here.

Askaholic:
If you're not working on this anymore I'm going to close the PR

CallMeMisterOwl:
> If you're not working on this anymore I'm going to close the PR

Yeah, sorry you can close it. Completely forgot about this and I'm busy rn, so this won't get done anytime soon.

PR #781, opened 2021-04-23
Issue/#769 nightly test run

### LINKED ISSUE #769: Add nightly test run to GitHub Actions
According to the hypothesis docs, hypothesis works best when you can run tests with a large number of examples (thousands or tens of thousands). Of course this takes a long time so it isn't really viable for a local test run or for CI on pull requests. 

We should set up a github action that runs on the develop branch once every 24 hours with a hypothesis profile that runs a large number of examples. Using profiles is described here: https://hypothesis.readthedocs.io/en/latest/settings.html#hypothesis.settings.load_profile. I would say we set `max_examples` to be at least 1000.

Askaholic:
Run hypothesis tests with 10,000 examples once per day. Caches the hypothesis example database so that falsifying examples found on one test run will be tried on subsequent runs. 

This means the nightly test run should essentially be gathering examples of edge cases that will then be tried on normal PR's, making the normal PR runs more effective.

Closes #769

### DIFF
--- a/.github/workflows/test.yml
+++ b/.github/workflows/test.yml
@@ -69,6 +69,12 @@ jobs:
     steps:
       - uses: actions/checkout@v2
 
+      - name: Cache hypothesis examples
+        uses: actions/cache@v2
+        with:
+          path: .hypothesis
+          key: .hypothesis
+
       - name: Setup Python
         uses: actions/setup-python@v1
         with:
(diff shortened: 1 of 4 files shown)


Askaholic:
I could also make the nightly test run a separate action, but then I'd have to copy over a lot of setup code so I wasn't sure if that was the right approach. On the other hand using the `if` conditions feels kindof dirty too.

0x647262 (approved):
LGTM!

PR #782, opened 2021-04-27
Feature/#747 save additional metadata in game results

### LINKED ISSUE #747: Save additional metadata in game results
# Motivation
There are some use cases in modded games where the game results may be more nuanced than simply `victory`, `defeat`, or `draw`. For example in phantom games, the results will also include whether or not the player was a phantom. See #663.

For galactic war an army may become defeated by recalling, and it is important that a distinction can be made between a recall defeat and an acu death, as gw includes perma death.

Therefore it would be nice if we could save additional metadata with game result reports and include it in the post game rabbitmq messages. We should just extend the current format similar to how phantomx has already done it by parsing game results as a space-separated list of values where the last value is an integer score and the second to last value is either `defeat`, `victory` or `draw` indicating the result for rating purposes.

# Format specification 
Pseudo-ebnf:
```
{<text> }defeat|victory|draw <integer>
```
Where `<text>` is some lowercase ascii text without whitespace and `<integer>` is an ascii representation of an integer.

Regex:
```regex
([a-z0-9]+ )*(defeat|victory|draw) -?[0-9]+
```

# Handling
The server should not do any special handling of metadata, however it should pass the metadata along in post-game rabbitmq messages so that other services may be able to implement special handling for certain meta tags.

eoinnoble:
Some third parties, and in future Galactic War, will send additional information with game result messages like
```
"phantom score -5"
"recall victory 5"
```
This PR captures those additional tags, tries to resolve them across multiple messages, and finally adds them to the `TeamRatingSummary` object(s) that we broadcast at the end of a game.

I'm not 100% happy with the naming of things, it feels like there are a lot of objects or enumerators named like `*Result`, `*Score` and `*Info` and I'm not sure if there are or should be clear distinctions about when to use each noun. I've also erred on the side of caution when resolving the metadata tags: if there's any discrepancy in the tags sent for an army then we'll only return the tags that exist in _all_ messages. We may want to be more lenient than that.

Closes #747.


Askaholic (commented):
This is a great start! I think some things can actually be simplified a little bit.

After reading this PR, I think the feature boils down to 3 chunks of work that we should tackle in the following order:
1. Save the metadata on the GameResultReport object instead of ignoring them
2. Include the metadata in the EndedGameInfo object
3. Add conflict resolution for metadata

So I'll probably try to focus my reviews in that order as well.

We'll definitely need some integration tests, but it may be easier to write those with the changes from my broadcast PR.

server/games/game.py
```diff
@@ -822,6 +844,12 @@ def get_player_outcome(self, player: Player) -> ArmyOutcome:
 
         return self._results.outcome(army)
 
+    def get_army_results(self, player: Player) -> ArmyResult:
+        army = self.get_player_option(player.id, "Army")
+        return ArmyResult(
+            player.id, army, ArmyOutcome.UNKNOWN, self._results.metadata(army)
```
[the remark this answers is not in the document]
eoinnoble: I initialize it as `UNKNOWN` and then set it to whatever the outcome is for each team in `EndedGameInfo.from_basic`.

server/games/game.py
```diff
@@ -466,11 +474,21 @@ async def resolve_game_results(self) -> EndedGameInfo:
         team_outcomes = [GameOutcome.UNKNOWN for _ in basic_info.teams]
 
         if self.validity is ValidityState.VALID:
+            team_player_partial_outcomes = []
```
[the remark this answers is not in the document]
eoinnoble: Ah ok, I'll look into it and see if I can manage it.

server/games/game_results.py
```diff
@@ -149,6 +161,42 @@ def _compute_outcome(self, army: int) -> ArmyOutcome:
         )
         return decision
 
+    def metadata(self, army: int) -> List[str]:
```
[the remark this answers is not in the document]
eoinnoble: Ok, check it out now, I think it's more in line with what you had in mind.

server/games/typedefs.py
```diff
@@ -154,13 +155,25 @@ def from_basic(
...
                 "same number of teams in the same order."
             )
 
+        # team_partial_army_results were initialized with ArmyOutcome.UNKNOWN so
+        # we need to get the resolved outcome from team_outcomes
+        team_complete_army_results = []
+        for outcome, army_results in zip(team_outcomes, team_partial_army_results):
+            complete_team_results = []
+            for player in army_results:
+                resolved_data = {**player._asdict(), "army_result": outcome.name}
```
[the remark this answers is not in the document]
eoinnoble: Ahh, I completely misunderstood that, thank you for catching it!

server/games/game.py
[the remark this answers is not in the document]
eoinnoble: Have taken a stab at it, let me know what you think?

server/games/typedefs.py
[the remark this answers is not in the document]
Askaholic: The result reporting/resolution is a bit confusing since the words victory/defeat/draw have different meanings in different contexts. That was the main reason for creating 3 different types for them so it would be easier to keep track of which context you're dealing with.

tests/unit_tests/test_game.py
```diff
@@ -951,16 +951,76 @@ async def test_game_results(game: Game, players):
...
+        ]
     assert result_dict["game_id"] == game.id
     assert result_dict["map_id"] == game.map_id
     assert result_dict["featured_mod"] == "faf"
     assert result_dict["sim_mod_ids"] == []
 
 
+async def test_team_game_results(game: Game, game_add_players):
+    game.state = GameState.LOBBY
+    players = (*game_add_players(game, 2, 0), *game_add_players(game, 2, 2))
```
Askaholic, line 972: Probably doesn't matter, but there is no team with id 0. Team 1 is the free for all team displayed like "-" in lobby, and the rest are all positive integers greater than 1.
```suggestion
    players = (*game_add_players(game, 2, 3), *game_add_players(game, 2, 2))
```

tests/unit_tests/test_game.py
```diff
@@ -951,16 +951,76 @@ async def test_game_results(game: Game, players):
...
+        for _ in players:
+            if i == 0:  # This player was the last one standing
+                await game.add_result(player.id, i, "victory", 1)
+            else:
+                await game.add_result(player.id, i, "defeat", 1)
+
+    game_results = await game.resolve_game_results()
+    result_dict = game_results.to_dict()
+
+    for i, team in enumerate(result_dict["teams"]):
```
Askaholic, line 985: If it's not too overly inconvenient I would prefer the asserts to be as "non-programatic" or as "hard coded" as possible. That just prevents the test from using a functions behavior to test its behavior. For instance this test would pass if `result_dict["teams"]` returned an empty list. I think it's perfectly OK to have a bit of duplicate code in unit tests.

tests/unit_tests/test_game.py
```diff
@@ -951,16 +951,76 @@ async def test_game_results(game: Game, players):
...
+        if i == 0:  # This player was the last one standing
+            await game.add_result(player.id, i, "victory", 1)
+        else:
+            await game.add_result(player.id, i, "defeat", 1)
+
+    game_results = await game.resolve_game_results()
+    result_dict = game_results.to_dict()
+
+    for i, team in enumerate(result_dict["teams"]):
+        assert team["outcome"] == "UNKNOWN"
```
Askaholic, line 1013: Here's what I was trying to say about gw games. They will most likely be unrated for some reason, but we still want to know what the team outcomes resolve to anyways. Like, even if the game is unrated because it was 3v1, the gw server still needs to know which team won. It would be possible for it to decide based on army results, but then we're just duplicating team outcome resolution across multiple services which isn't great.

Askaholic (commented):
Yes! This looks really good now; much simpler than it was at first.

Lastly I'd like to see some integration tests, which means I guess I had better get a move on with my rabbitmq PR. You could also copy the `AioQueueProtocol` helper class and then rebase again once it's merged. Or maybe this will be merged first and then I can rebase my PR, doesn't really matter.

server/games/game.py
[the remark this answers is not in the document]
Askaholic: For some reason I thought the messages wouldn't be sent to rabbitmq if the game wasn't valid, but looking over the code again I don't think that's the case. `GameService.publish_game_results` always publishes to rmq, and only sends to the *rating service* if the game wasn't valid, which is the behavior we want.

So I think you can just remove that `if self.validity is VALID` check and everything will work as expected. Definitely want a test or two to confirm tho.

tests/unit_tests/test_game.py
eoinnoble: Should I just remove the `if self.validity is ValidityState.VALID` check entirely then, so the team outcomes are always calculated?

tests/unit_tests/test_game.py
eoinnoble: More like this?

```python
assert result_dict["teams"][0]["outcome"] == "UNKNOWN"
assert result_dict["teams"][0]team["army_results"] == [
    ArmyResult(1, 0, "VICTORY", []),
    ...
    ArmyResult(4, 3, "DEFEAT", [])
]
```

Askaholic: Yea or even
```python3
assert result_dict["teams"][0] == {
    "outcome": "VICTORY",
    "army_results": [
        ...
    ]
}
```

tests/unit_tests/test_game.py
Askaholic: I think that would be fine

server/games/game.py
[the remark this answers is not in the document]
eoinnoble: Hmm, ok, will take a look at this.

tests/integration_tests/test_game.py
```diff
@@ -532,3 +537,109 @@ async def test_gamestate_ended_clears_references(
...
+        ("test", "test_password"), lobby_server
+    )
+    guest_id, _, guest_proto = await connect_and_sign_in(
+        ("Rhiza", "puff_the_magic_dragon"), lobby_server
+    )
+    await read_until_command(guest_proto, "game_info")
+    ratings = await get_player_ratings(host_proto, "test", "Rhiza")
+
+    # Set up the game
+    game_id = await host_game(host_proto)
```
Askaholic, line 561: Looks like this helper function already has a parameter for featured mod. You'll just have to add the mod to the `test-data.sql` file.
```suggestion
    game_id = await host_game(host_proto, mod="gw")
```

tests/integration_tests/test_game.py
```diff
@@ -532,3 +537,109 @@ async def test_gamestate_ended_clears_references(
...
+
+    # Now disconnect both players
+    for proto in (host_proto, guest_proto):
+        await proto.send_message({
+            "target": "game",
+            "command": "GameState",
+            "args": ["Ended"]
+        })
+
+    expected_army_results = [
```
Askaholic, line 630: Could you just assert the contents of the entire message, instead of only army results?

tests/integration_tests/test_game.py
```diff
@@ -532,3 +537,109 @@ async def test_gamestate_ended_clears_references(
...
+        [1, 1, "VICTORY", []],
+        [3, 2, "DEFEAT", ["recall"]]
+    ]
+    messages = []
+
+    await read_until(
+        mq_proto_all, lambda message: "teams" in message and not messages.append(message)
+    )
+
```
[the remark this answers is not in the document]
eoinnoble: That's much better, thanks, `not messages.append` was making me unhappy.

tests/integration_tests/test_game.py
eoinnoble: Sure, I can do that.

tests/data/test-data.sql
```diff
@@ -229,6 +229,7 @@ INSERT INTO `coop_map` (`type`, `name`, `description`, `version`, `filename`) VA
 insert into game_featuredMods (id, gamemod, name, description, publish, git_url, git_branch, file_extension, allow_override) values
   (1, 'faf', 'FAF', 'Forged Alliance Forever', 1, 'https://github.com/FAForever/fa.git', 'deploy/faf', 'nx2', FALSE),
   (6, 'ladder1v1', 'FAF', 'Ladder games', 1, 'https://github.com/FAForever/fa.git', 'deploy/faf', 'nx2', TRUE),
+  (24, 'gw', '', 'Galactic War', 1, NULL, NULL, NULL, NULL),
```
Askaholic, line 232: `publish` is set to false for GW because you're not supposed to be able to start GW games from the "Host Game" menu. It doesn't mean that the mod is disabled or anything. I think it just affects whether the mod is returned from the API.

tests/integration_tests/test_game.py
```diff
@@ -532,3 +537,112 @@ async def test_gamestate_ended_clears_references(
...
+    for proto in (host_proto, guest_proto):
+        await proto.send_message({
+            "target": "game",
+            "command": "GameState",
+            "args": ["Ended"]
+        })
+
+    message = await asyncio.wait_for(mq_proto_all.read_message(), timeout=5)
+
+    assert message == {
```
Askaholic, line 624: I like this way of asserting a lot better.

tests/integration_tests/test_game.py
```diff
@@ -532,3 +537,112 @@ async def test_gamestate_ended_clears_references(
 
     assert test.lobby_connection.game_connection is None
     assert rhiza.lobby_connection.game_connection is None
+
+
+@pytest.mark.rabbitmq
+@fast_forward(30)
+async def test_game_ended_broadcasts_army_results(lobby_server, channel):
```
Askaholic, line 544: We should definitely have a test for a game with uneven teams, like 2v1 for example. Dunno if you wanna do 2 separate tests or just make this one use uneven teams.

server/games/game.py
```diff
@@ -463,14 +471,21 @@ async def resolve_game_results(self) -> EndedGameInfo:
...
+
+        team_army_results = [
+            [self.get_army_results(player) for player in team]
+            for team in basic_info.teams
+        ]
+
         team_outcomes = [GameOutcome.UNKNOWN for _ in basic_info.teams]
 
-        if self.validity is ValidityState.VALID:
+        if self.validity == ValidityState.VALID:
```
Askaholic, line 482: So looks like you are gonna keep this in for now?

tests/integration_tests/test_matchmaker.py
```diff
@@ -306,7 +306,7 @@ async def test_game_matchmaking_close_fa_and_requeue(lobby_server):
 async def test_anti_map_repetition(lobby_server):
     proto1, proto2 = await queue_players_for_matchmaking(lobby_server)
 
-    # Play one game so that it exists in the players history
+    # Play one game so that it exists in the players' history
```
Askaholic, line 309: Nice, lol

server/games/game.py
eoinnoble: Yeah I thought this might be better as a separate PR as it looked like I might have to alter the logic of `game_results.resolve_game` as well and we might want to be able to roll that back easily without undoing the changes in this PR?

tests/integration_tests/test_game.py
```diff
@@ -532,3 +537,241 @@ async def test_gamestate_ended_clears_references(
...
+        "map_id": 7,
+        "featured_mod": "gw",
+        "sim_mod_ids": [],
+        "commander_kills": {},
+        "validity": "VALID",
+        "teams": [
+            {
+                "outcome": "DEFEAT",
+                "player_ids": [3],
+                "army_results": [[3, 2, "DEFEAT", ["recall"]]]
```
Askaholic, line 636: I've been thinking maybe it would be better to have these be a dictionary than a list.
```suggestion
                "army_results": [
                    {
                        "player_id": 3, 
                        "team": 2, 
                        "outcome": "DEFEAT", 
                        "metadata": ["recall"]
                    }
                ]
```

eoinnoble: That's a good idea 👍

eoinnoble: Is it weird to you that in my 2v1 scenario there are three items in `message['teams']`? I'd have expected two of those to be combined into a single entry?

tests/integration_tests/test_game.py
```diff
@@ -532,3 +537,241 @@ async def test_gamestate_ended_clears_references(
...
+
+    # Set up the game
+    game_id = await host_game(host_proto, mod="gw")
+    await join_game(guest_proto, game_id)
+    await join_game(other_guest_proto, game_id)
+    # Set player options
+    await send_player_options(
+        host_proto,
+        [host_id, "Army", 1],
+        [host_id, "Team", 1],
```
Askaholic, line 680: This is a 1v1v1 because 2 people are in the ffa team.

tests/integration_tests/test_game.py
Askaholic: Looks like your team setup isn’t quite right.

tests/integration_tests/test_game.py
eoinnoble: Ah ok 🤦

tests/integration_tests/test_game.py
```diff
@@ -532,3 +537,270 @@ async def test_gamestate_ended_clears_references(
...
+        "validity": "VALID",
+        "teams": [
+            {
+                "outcome": "DEFEAT",
+                "player_ids": [3],
+                "army_results": [
+                    {
+                        "player_id": 3,
+                        "army": 2,
+                        "army_result": "DEFEAT",
```
Askaholic, line 640:
```suggestion
                        "army_outcome": "DEFEAT",
```

tests/integration_tests/test_game.py
```diff
@@ -532,3 +537,270 @@ async def test_gamestate_ended_clears_references(
...
+    # Set up the game
+    game_id = await host_game(host_proto, mod="gw")
+    await join_game(guest_proto, game_id)
+    await join_game(other_guest_proto, game_id)
+    # Set player options
+    await send_player_options(
+        host_proto,
+        [host_id, "Army", 1],
+        [host_id, "Team", 0],
+        [host_id, "StartSpot", 0],
```
Askaholic, line 695: I wouldn't use 0 for anything unless you're explicitly testing for it.

tests/integration_tests/test_game.py
```diff
@@ -532,3 +537,270 @@ async def test_gamestate_ended_clears_references(
...
+
+    # Set up the game
+    game_id = await host_game(host_proto, mod="gw")
+    await join_game(guest_proto, game_id)
+    # Set player options
+    await send_player_options(
+        host_proto,
+        [host_id, "Army", 1],
+        [host_id, "Team", 1],
+        [host_id, "StartSpot", 0],
```
Askaholic, line 567: Again, idk about the 0 here just cuz the game will never send it. Team 1 is probably fine but maybe a little weird.

tests/data/test-data.sql
```diff
@@ -229,6 +229,7 @@ INSERT INTO `coop_map` (`type`, `name`, `description`, `version`, `filename`) VA
 insert into game_featuredMods (id, gamemod, name, description, publish, git_url, git_branch, file_extension, allow_override) values
   (1, 'faf', 'FAF', 'Forged Alliance Forever', 1, 'https://github.com/FAForever/fa.git', 'deploy/faf', 'nx2', FALSE),
   (6, 'ladder1v1', 'FAF', 'Ladder games', 1, 'https://github.com/FAForever/fa.git', 'deploy/faf', 'nx2', TRUE),
+  (24, 'gw', '', 'Galactic War', 0, NULL, NULL, NULL, NULL), -- GW unpublished because you can't start them from "Host game"
```
Askaholic, line 232: Well this comment is obviously wrong because you're doing it in the unit tests xD. But it shouldn't show up in the featured mod list in the client.

server/games/typedefs.py
```diff
@@ -154,6 +155,7 @@ def from_basic(
         validity: ValidityState,
         team_outcomes: List[GameOutcome],
         commander_kills: Dict[str, int],
+        team_army_results: List[List[OrderedDict]],
```
Askaholic, line 158: Can we make this a `List[ArmyResult]` and only convert the inner ones to dictionaries in the `to_dict` method? I think it would work fine if you add a `to_dict` method for `TeamRatingSummary`.

eoinnoble: Do you mean `List[List[ArmyResult]]`? Each team could have more than one.

Askaholic: Yea, my bad

Askaholic:
Alright, I'm merging this. Here's something quirky I noticed, but it's only because we're using python3.7.

When the end game result is logged, it prints out the object before serialization (no problem this is actually nice for printing enums) and in python3.7 the return value of `NamedTuple._asdict` is an `OrderedDict` which prints out kinda ugly:
```
TRACE    May 29  22:23:45 MessageQueueService            Published message {'game_id': 13757642, 'rating_type': 'global', 'map_id': 558, 'featured_mod': 'faf', 'sim_mod_ids': [], 'commander_kills': {}, 'validity': 'VALID', 'teams': [{'outcome': 'DEFEAT', 'player_ids': [59632], 'army_results': [OrderedDict([('player_id', 59632), ('army', 1), ('army_outcome', 'DEFEAT'), ('metadata', ['recall'])])]}, {'outcome': 'VICTORY', 'player_ids': [273862], 'army_results': [OrderedDict([('player_id', 273862), ('army', 2), ('army_outcome', 'VICTORY'), ('metadata', [])])]}]} to faf-lobby/success.gameResults.create
```

However, in python3.8 the return value is just a normal dict so lets just leave it. (I'll have to give upgrading another shot at some point)

PR #783, opened 2021-05-01
Feature/#727 party matchmaking algorithm

BlackYps:
This builds upon the changes made by @cleborys in #749 
A clean diff is available here: https://github.com/cleborys/server/pull/6

This adds the actual matchmaking algorithm to enable matchmaking for teams with more than two players.

### DIFF
--- a/server/matchmaker/search.py
+++ b/server/matchmaker/search.py
@@ -249,6 +249,13 @@ class Search:
         """For debugging"""
         return f"Search({[p.login for p in self.players]}, {self.average_rating})"
 
+    def get_original_searches(self) -> List["Search"]:
+        """
+        Returns the searches of which this Search is comprised,
+        as if it were a CombinedSearch of one
+        """
+        return [self]
+
 
 class CombinedSearch(Search):
     def __init__(self, *searches: Search):
@@ -329,3 +336,9 @@ class CombinedSearch(Search):
 
     def __str__(self):
         return "CombinedSearch({})".format(",".join(str(s) for s in self.searches))
+
+    def get_original_searches(self) -> List[Search]:
+        """
+        Returns the searches of which this CombinedSearch is comprised
+        """
+        return list(self.searches)
--- a/server/matchmaker/search.py
+++ b/server/matchmaker/search.py
@@ -247,7 +247,13 @@ class Search:
 
     def __repr__(self) -> str:
         """For debugging"""
-        return f"Search({[p.login for p in self.players]}, {self.average_rating})"
+        string = f"({len(self.players)} players, {self.average_rating}"
+        if self.failed_matching_attempts > 0:
+            string += f", FMA: {self.failed_matching_attempts}"
+        if self.has_newbie():
+            string += ", has_newbie"
+        string += ")"
+        return string
 
     def get_original_searches(self) -> List["Search"]:
         """
--- a/server/matchmaker/search.py
+++ b/server/matchmaker/search.py
@@ -247,7 +247,7 @@ class Search:
 
     def __repr__(self) -> str:
         """For debugging"""
-        string = f"({len(self.players)} players, {self.average_rating}"
+        string = f"Search({[p.login for p in self.players]}, {self.average_rating}"
         if self.failed_matching_attempts > 0:
             string += f", FMA: {self.failed_matching_attempts}"
         if self.has_newbie():
--- a/server/matchmaker/search.py
+++ b/server/matchmaker/search.py
@@ -30,6 +30,15 @@ class Game:
         self.quality = match_quality
 
 
+def get_average_rating(searches):
+    ratings = []
+    for s in searches:
+        for mean, dev in s.raw_ratings:
+            rating = mean - 3 * dev
+            ratings.append(rating)
+    return statistics.mean(ratings)
+
+
 @with_logger
 class Search:
     """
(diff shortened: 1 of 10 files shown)


tests/unit_tests/test_matchmaker_algorithm_team_matchmaker.py
```diff
@@ -0,0 +1,320 @@
...
+    searches = searches_with_player_size[0]
+    team_size = searches_with_player_size[1]
+    assume(sum(len(search.players) for search in searches) == 2 * team_size)
+
+    try:
+        teams = TeamMatchMaker().make_teams(searches, team_size)
+        assert len(teams[0].players) == team_size
+        assert len(teams[1].players) == team_size
+    except UnevenTeamsException:
+        assert True
```
Askaholic, line 128: Why are we asserting true? This will never fail. If you want a noop use `pass`

tests/unit_tests/test_matchmaker_algorithm_team_matchmaker.py
```diff
@@ -0,0 +1,320 @@
...
+    c1 = CombinedSearch(*[s.pop(), s.pop(), s.pop()])
+    c2 = CombinedSearch(*[s.pop(), s.pop(), s.pop()])
+    c3 = CombinedSearch(*[s.pop(), s.pop()])
+    s.append(c1)
+    s.append(c2)
+    s.append(c3)
+
+    matches, unmatched = TeamMatchMaker().find(s, 4)
+
+    assert len(matches) == 0
```
Askaholic, line 211: If it's a list I usually do `assert matches == []`

tests/unit_tests/test_matchmaker_algorithm_team_matchmaker.py
```diff
@@ -0,0 +1,320 @@
...
+def test_pick_noncolliding_games_no_games():
+    assert TeamMatchMaker().pick_noncolliding_games([]) == []
+
+
+@given(st_game_candidates_list(min_size=1))
+def test_pick_noncolliding_games(games):
+    max_quality_game = games[0]
+    for game in games:
+        if game.quality > max_quality_game.quality:
+            max_quality_game = game
```
Askaholic, line 309:
```python
max_quality_game = max(games, key=lambda game: game.quality)
```

tests/unit_tests/test_matchmaker_algorithm_team_matchmaker.py
```diff
@@ -0,0 +1,317 @@
...
+    quality_before = matchmaker.assign_game_quality((team_a, team_b)).quality
+
+    team_a.register_failed_matching_attempt()
+    team_b.register_failed_matching_attempt()
+    quality_after = matchmaker.assign_game_quality((team_a, team_b)).quality
+
+    num_newbies = sum(search.has_newbie() for search in team_a.get_original_searches())
+    num_newbies += sum(search.has_newbie() for search in team_b.get_original_searches())
+
+    assert abs(quality_before + 6 * config.TIME_BONUS + num_newbies * config.NEWBIE_TIME_BONUS - quality_after) < 0.0000001
```
Askaholic, line 265: Could use `pytest.approx` and compare to 0?

server/matchmaker/search.py
```diff
@@ -234,7 +247,20 @@ def _players_repr(self) -> str:
 
     def __repr__(self) -> str:
         """For debugging"""
-        return f"Search({[p.login for p in self.players]})"
+        string = f"Search({[p.login for p in self.players]}, {self.average_rating}"
+        if self.failed_matching_attempts > 0:
+            string += f", FMA: {self.failed_matching_attempts}"
+        if self.has_newbie():
+            string += ", has_newbie"
+        string += ")"
```
Askaholic, line 255: Pretty low hanging fruit but rewriting this so there's no `+=` on strings might make it better

tests/unit_tests/test_matchmaker_algorithm_team_matchmaker.py
```diff
@@ -0,0 +1,317 @@
...
+    player_id_counter = 0
+
+    def make(
+        mean: int = 1500,
+        deviation: int = 500,
+        ladder_games: int = config.NEWBIE_MIN_GAMES+1,
+        name=None
+    ):
+        nonlocal player_id_counter
+        """Make a player with the given ratings"""
```
Askaholic, line 35: I think this codacy issue could be fixed, just swap these two lines

server/matchmaker/search.py
```diff
@@ -247,13 +247,9 @@ def _players_repr(self) -> str:
...
     def __repr__(self) -> str:
         """For debugging"""
-        string = f"Search({[p.login for p in self.players]}, {self.average_rating}"
-        if self.failed_matching_attempts > 0:
-            string += f", FMA: {self.failed_matching_attempts}"
-        if self.has_newbie():
-            string += ", has_newbie"
-        string += ")"
-        return string
+        return f"Search({[p.login for p in self.players]}, {self.average_rating}" \
```
Askaholic, line 250: Don't need the backslash at the end of the line

BlackYps: Without the backslash the following code becomes unreachable because of the return statement.

Askaholic: Oh, then put it in parenthesis. Like this:

```python
return (
    f"some string {stuff}. "
    r"This part is a raw string, "
    "and this part is a normal string"
)
```
That would become
```python
"some string value_of_stuff. This part is a raw string, and this part is a normal string"
```

PR #792, opened 2021-05-17
Use maxminddb directly instead of geoip2

Askaholic:
The `geoip2` package depends on `requests` and provides no way to opt out of it. Since we're not using the downloading feature anyways, we can pretty easily change the db lookup to use `maxminddb` directly instead of using the wrapper provided by `geoip2`.

Now the only package requiring `requests` is `twilio`. Since we aren't even using Twilio and it is causing other dependency inconveniences in the OAuth branch, I would be inclined to remove it as well.


server/geoip_service.py
```diff
@@ -193,11 +193,16 @@ def load_db(self) -> None:
             # Set the time first, if the file is corrupted we don't need to try
             # loading it again anyways
             self.db_update_time = datetime.now()
-            self.db = geoip2.database.Reader(self.file_path)
+            new_db = maxminddb.open_database(self.file_path)
```
eoinnoble, line 196: If it were me I would wrap only this line in `try`/`except`, as this is the line that looks like it raises? Do we want to update `self.db_update_time` if opening the database might fail immediately after? I'd probably put the `open_database` call in `try`/`except` and then the rest in an `else`.

server/geoip_service.py
```diff
@@ -193,11 +193,16 @@ def load_db(self) -> None:
...
+
+            if self.db is not None:
+                self.db.close()
```
eoinnoble, line 199: Is it worth awaiting your `shutdown` function here and making this function `async`?

eoinnoble (approved):
This looks good, just a question about order of operations in `load_db`.

server/geoip_service.py
Askaholic: Nah, `shutdown` is a lifecycle hook for services that is called when the server is being stopped. It would be bad style to call it from elsewhere. I mean I could make a `self.close` method and call if from shutdown and here, but that seems a bit overkill for refactoring 2 lines :P.

server/geoip_service.py
Askaholic: In the grand scheme of things it doesn't matter if the `db_update_time` is updated before or after. If we update it after then we might generate more errors in the log tho if the database file is corrupted, as the service will keep trying to load it every 10 minutes.

server/geoip_service.py
```diff
@@ -189,18 +195,24 @@ def load_db(self) -> None:
...
                 "Failed to load maxmind db! Maybe the download was interrupted"
             )
+        else:
+            if self.db is not None:
+                self.db.close()
+
+            self.db = new_db
+            self._logger.info(
+                "File loaded successfully from %s", self.file_path
+            )
```
Wesmania, line 215: We can move this to the try block after the `open_database` line, it's equivalent and perhaps more readable.

Wesmania (commented):
Otherwise seems fine

server/geoip_service.py
Wesmania: Ah, nevermind, there's been talk about this earlier. I prefer early return in the 'except branch', but it's a nitpick, do what you think is best.

Askaholic: Haha. Yea I really don't know what's better. Usually I put it inside the try-except, but Eoin suggested this way so I thought I'd give it a try.

eoinnoble: In my opinion this way we catch what we expect can fail, and anything else that starts failing with similar errors in the future will be a surprise to us instead of being quietly caught 🤷

PR #798, opened 2021-06-09
Add timeout to base game class

### LINKED ISSUE #797: Memory leak when ladder games fail to start
When ladder games time out here:
https://github.com/FAForever/server/blob/23955b8635b28aeeea04f4063416bb673c7dafba/server/ladder_service.py#L452-L454
`on_game_end` will not be called to clean the game up because the `try/finally` block suppresses all exceptions. Normally, there would be another timeout going in the background which would be launched from the constructor such as here in `CoopGame`
https://github.com/FAForever/server/blob/23955b8635b28aeeea04f4063416bb673c7dafba/server/games/coop.py#L25
However, that is not present in `LadderGame`.

## Solutions
We could either
1. Modify the `try/finally` block to propagate `asyncio.TimeoutError`s, or
2. Add a second timeout to the constructor of `LadderGame` like it is done in `CoopGame`, or
3. Refactor the second timeout to the base class `Game` and add a constructor parameter for setting the timeout.

I think options 1 and 3 are probably the best. 1 is nice because it avoids having to define the timeout in 2 places so that the two timeouts don't interfere with eachother. It's just a little more error prone as there is no guaranteed timeout generated by the constructor, and relies on the caller to ensure that the game is cleaned up correctly.

Sheikah45:
Fixes #797

### DIFF
--- a/server/games/game.py
+++ b/server/games/game.py
@@ -70,6 +70,7 @@ class Game():
         displayed_rating_range: Optional[InclusiveRange] = None,
         enforce_rating_range: bool = False,
         max_players: int = 12,
+        setup_timeout: int = 60,
     ):
         self._db = database
         self._results = GameResultReports(id_)
@@ -120,8 +121,9 @@ class Game():
         self._launch_fut = asyncio.Future()
 
         self._logger.debug("%s created", self)
+        asyncio.get_event_loop().create_task(self.timeout_game(setup_timeout))
 
-    async def timeout_game(self, timeout: int = 30):
+    async def timeout_game(self, timeout: int = 60):
         await asyncio.sleep(timeout)
         if self.state is GameState.INITIALIZING:
             self._is_hosted.set_exception(
--- a/tests/unit_tests/test_ladder_service.py
+++ b/tests/unit_tests/test_ladder_service.py
@@ -149,17 +149,23 @@ async def test_start_game_1v1(
     LadderGame.wait_launched.assert_called_once()
 
 
-@fast_forward(35)
+@fast_forward(65)
 async def test_start_game_timeout(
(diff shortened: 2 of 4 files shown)


server/games/game.py
```diff
@@ -120,8 +121,9 @@ def __init__(
         self._launch_fut = asyncio.Future()
 
         self._logger.debug("%s created", self)
+        asyncio.get_event_loop().create_task(self.timeout_game(setup_timeout))
 
-    async def timeout_game(self, timeout: int = 30):
+    async def timeout_game(self, timeout: int = 60):
```
Askaholic, line 126: So this does change the timeout for custom games. I think that might cause 0 player games to hang around longer. You could fix this by adding the `setup_timeout` parameter to the `new_kwargs` dict in the custom game constructor

Askaholic (commented):
Can we get a simple regression test in `unit_tests/test_ladder_service.py`. It just needs to 
1. Call `LadderService.start_game`
2. Wait 60+ seconds
3. Assert that `Game.on_game_ended` was called

Make sure it's an async test and use the `fast_forward` decorator. Also make sure that it actually fails without these changes :P

server/games/game.py
Sheikah45: Done

tests/unit_tests/test_ladder_service.py
```diff
@@ -149,17 +149,23 @@ async def test_start_game_1v1(
     LadderGame.wait_launched.assert_called_once()
 
 
-@fast_forward(35)
+@fast_forward(65)
```
Askaholic, line 152: Oh nice there was already a test for this

PR #804, opened 2021-06-29
Let rating service publish rating changes to RabbitMQ

### LINKED ISSUE #803: Publish rating changes to RabbitMQ for LeagueService to use
Right now only game results are published to RabbitMQ, but since the league service doesn't want to duplicate the rater logic, what it really wants is an event published on rating change.
The expected format of these messages is e.g. available [in these tests](https://github.com/FAForever/faf-league-service/blob/develop/tests/integration_tests/test_league_request_rating.py).
When this is done the [routing keys](https://github.com/FAForever/faf-league-service/blob/develop/service/config.py) in the league service should be updated accordingly.

(Right now the league service actually listens for game results using the "success.gameResult.create" routing key, but it then tries to parse the message as if it was an event as described above and hence fails.)

cleborys:
Closes #803.

### DIFF
--- a/server/rating_service/rating_service.py
+++ b/server/rating_service/rating_service.py
@@ -19,6 +19,7 @@ from server.db.models import (
 )
 from server.decorators import with_logger
 from server.games.game_results import GameOutcome
+from server.message_queue_service import MessageQueueService
 from server.metrics import rating_service_backlog
 from server.player_service import PlayerService
 from server.rating import RatingType, RatingTypeMap
@@ -50,13 +51,19 @@ class RatingService(Service):
     atomic.
     """
 
-    def __init__(self, database: FAFDatabase, player_service: PlayerService):
+    def __init__(
+        self,
+        database: FAFDatabase,
+        player_service: PlayerService,
+        message_queue_service: MessageQueueService
+    ):
         self._db = database
         self._player_service_callback = player_service.signal_player_rating_change
         self._accept_input = False
         self._queue = asyncio.Queue()
         self._task = None
         self._rating_type_ids = None
+        self._message_queue_service = message_queue_service
 
     async def initialize(self) -> None:
         if self._task is not None:
@@ -137,6 +144,9 @@ class RatingService(Service):
         }
         await self._persist_rating_changes(
             summary.game_id, summary.rating_type, old_ratings, new_ratings, outcome_map
+        )
+        await self._publish_rating_changes(
+            summary.rating_type, new_ratings, outcome_map
         )
 
     async def _get_rating_data(self, summary: GameRatingSummary) -> GameRatingData:
@@ -380,6 +390,31 @@ class RatingService(Service):
             "Sending player rating update for player with id %i", player_id
         )
         self._player_service_callback(player_id, rating_type, new_rating)
+
+    async def _publish_rating_changes(
+        self,
+        rating_type: str,
+        new_ratings: Dict[PlayerID, Rating],
+        outcomes: Dict[PlayerID, GameOutcome],
+    ):
+        for player_id, new_rating in new_ratings.items():
+            if player_id not in outcomes:
+                self._logger.error("Missing outcome for player %i", player_id)
+                continue
+
+            rating_change_dict = {
+                "player_id": player_id,
+                "rating_type": rating_type,
+                "new_rating_mean": new_rating.mu,
+                "new_rating_deviation": new_rating.sigma,
+                "outcome": outcomes[player_id]
+            }
+
+            await self._message_queue_service.publish(
+                config.MQ_EXCHANGE_NAME,
+                "success.ratingChange.create",
+                rating_change_dict,
+            )
 
     async def _join_rating_queue(self) -> None:
         """
--- a/server/rating_service/rating_service.py
+++ b/server/rating_service/rating_service.py
@@ -146,7 +146,7 @@ class RatingService(Service):
             summary.game_id, summary.rating_type, old_ratings, new_ratings, outcome_map
         )
         await self._publish_rating_changes(
-            summary.rating_type, new_ratings, outcome_map
+            summary.rating_type, old_ratings, new_ratings, outcome_map
         )
 
     async def _get_rating_data(self, summary: GameRatingSummary) -> GameRatingData:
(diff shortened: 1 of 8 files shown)


cleborys:
Running the tests locally, I see `tests/integration_tests/test_matchmaker.py` fail from time to time, both on this branch and on `develop`. It's not always the same test that fails and running in isolation they all seem to pass.

Askaholic (approved):
Maybe you want to add an integration test? I think I added some helpers for getting rabbitmq messages in my broadcast PR.

server/rating_service/rating_service.py
```diff
@@ -381,6 +391,31 @@ def _update_player_object(
...
+            if player_id not in outcomes:
+                self._logger.error("Missing outcome for player %i", player_id)
+                continue
+
+            rating_change_dict = {
+                "player_id": player_id,
+                "rating_type": rating_type,
+                "new_rating_mean": new_rating.mu,
+                "new_rating_deviation": new_rating.sigma,
+                "outcome": outcomes[player_id]
```
Askaholic, line 410: Oh, from the unit tests. It's failing to JSON serialize outcomes
```suggestion
                "outcome": outcomes[player_id].value
```

server/rating_service/rating_service.py
```diff
@@ -381,6 +391,31 @@ def _update_player_object(
...
+        for player_id, new_rating in new_ratings.items():
+            if player_id not in outcomes:
+                self._logger.error("Missing outcome for player %i", player_id)
+                continue
+
+            rating_change_dict = {
+                "player_id": player_id,
+                "rating_type": rating_type,
+                "new_rating_mean": new_rating.mu,
+                "new_rating_deviation": new_rating.sigma,
```
BlackYps, line 409: Maybe we should also publish the old rating. The league service doesn't need it, but it could be useful for other things in the future.

BlackYps (approved):

server/rating_service/rating_service.py
BlackYps: I can't reproduce the failing JSON serialization on my machine.

server/rating_service/rating_service.py
cleborys: Good point

server/rating_service/rating_service.py
cleborys: Interesting, they pass locally for me too. Also after a `pipenv sync --dev`, I don't know what's different.

cleborys:
I added an integration test - turns out there are some floating point issues somewhere in the interaction between the database and trueskill that lead to results like this:
```
E       AssertionError: assert {'new_rating_...00000002, ...} == {'new_rating_...: 2000.0, ...}
E         Omitting 6 identical items, use -vv to show
E         Differing items:
E         {'old_rating_mean': 2000.0000000000002} != {'old_rating_mean': 2000.0}
```

tests/integration_tests/test_game.py
```diff
@@ -231,6 +232,147 @@ async def test_game_ended_rates_game(lobby_server):
         await read_until_command(host_proto, "player_info", timeout=10)
 
 
+#@pytest.mark.rabbitmq
```
Askaholic, line 235: If you need to run this test locally use `-m rabbitmq` or `-m ""` to run all tests

tests/integration_tests/test_game.py
```diff
@@ -231,6 +232,147 @@ async def test_game_ended_rates_game(lobby_server):
...
+    expected_message = expected_messages_by_id[first_id]
+    for key in ["rating_type", "outcome"]:
+        assert first_message[key] == expected_message[key]
+    for key in [
+        "new_rating_mean",
+        "new_rating_deviation",
+        "old_rating_mean",
+        "old_rating_deviation",
+    ]:
+        assert math.isclose(first_message[key], expected_message[key])
```
Askaholic, line 348: Maybe try https://docs.pytest.org/en/6.2.x/reference.html#pytest.approx

I think this would let you write:
```python3
assert first_message == approx(expected_message)
```

tests/unit_tests/test_game_rating.py
```diff
@@ -189,7 +201,41 @@ async def test_on_game_end_global_ratings(custom_game, players):
...
+    await rating_service._join_rating_queue()
+
+    results = get_published_results_by_player_id(
+        custom_game.game_service._rating_service
+    )
+    assert players.hosting.id in results
+    assert players.joining.id in results
+
+    host_result = results[players.hosting.id]
+    assert host_result["rating_type"] is RatingType.GLOBAL
```
Askaholic, line 226: I don't think you want `is` here as `RatingType.GLOBAL` is just a string constant (not an enum)

tests/unit_tests/test_game_rating.py
```diff
@@ -189,7 +201,41 @@ async def test_on_game_end_global_ratings(custom_game, players):
...
+    assert "new_rating_mean" in host_result
+    assert "new_rating_deviation" in host_result
+    assert host_result["outcome"] is GameOutcome.VICTORY.value
```
Askaholic, line 229: Similar thing here, I think you want `==` instead of `is`

tests/unit_tests/test_game_rating.py
```diff
@@ -212,6 +258,67 @@ async def test_on_game_end_ladder_ratings_(ladder_game, players):
...
+
+    await ladder_game.on_game_end()
+    await rating_service._join_rating_queue()
+
+    results = get_published_results_by_player_id(rating_service)
+    assert players.hosting.id in results
+    assert players.joining.id in results
+
+    host_result = results[players.hosting.id]
+    assert host_result["rating_type"] is RatingType.LADDER_1V1
```
Askaholic, line 281: Same in this test

tests/integration_tests/test_game.py
cleborys: Argh. I had forgotten how to do that so I did the obvious stupid thing and of course it came back to bite me.

tests/unit_tests/test_game_rating.py
cleborys: Ah yes, I forgot that that changed a while ago...

tests/unit_tests/test_game_rating.py
cleborys: `GameOutcome` is still an enum though.

PR #805, opened 2021-07-01
Fix tests trying to contact the real hydra server

Askaholic:
This was the cause of some flaky tests when the server went down.

### DIFF
--- a/tests/integration_tests/conftest.py
+++ b/tests/integration_tests/conftest.py
@@ -17,6 +17,7 @@ from server import (
     BroadcastService,
     GameService,
     LadderService,
+    OAuthService,
     PartyService,
     ServerInstance
 )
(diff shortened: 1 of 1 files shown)


Sheikah45 (approved):
Ah okay I see there was a race condition in terms of the oauth service initialization vs when the first update of the public keys happens.

Although I am confused as to how that happened since the initialization should have only happened after the jwk path was mocked. And I would think that would cause all the tests to fail even if the server wasn't done as it would have different jwks than the one the test cases are signed with.

But overall changes look good, but confused as to why it was flaky in the first place.

Askaholic:
It's only flaky in that one test `test_multiple_contexts` because that one doesn't use the `lobby_server` fixture.

PR #809, opened 2021-07-22
Don't serialize team if it is None

### LINKED ISSUE #808: String "null" sent as team in game_info
If the server has received info that a player has joined the game it could potentially send "null" since it might not have received the Team info for the player yet

Sheikah45:
Fixes #808

### DIFF
--- a/server/games/game.py
+++ b/server/games/game.py
@@ -908,7 +908,7 @@ class Game():
                     player.login for player in self.players
                     if self.get_player_option(player.id, "Team") == team
                 ]
-                for team in self.teams if team
+                for team in self.teams if team is None
             }
         }
 
--- a/server/games/game.py
+++ b/server/games/game.py
@@ -908,7 +908,7 @@ class Game():
                     player.login for player in self.players
                     if self.get_player_option(player.id, "Team") == team
                 ]
-                for team in self.teams if team is None
+                for team in self.teams if team is not None
             }
         }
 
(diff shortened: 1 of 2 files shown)


server/games/game.py
```diff
@@ -908,7 +908,7 @@ def to_dict(self):
                     player.login for player in self.players
                     if self.get_player_option(player.id, "Team") == team
                 ]
-                for team in self.teams
+                for team in self.teams if team
```
Askaholic, line 911: Maybe do `is None` since this will also ignore the team if it’s set to `0`

Sheikah45: Yeah I can do that but as far as I know 0 would be an invalid value as well. But still would be good.

Askaholic: Yea it shouldn’t come up in SupCom, but still nice to be robust. Also there is TAF which forked from here and occasionally merges our new changes in, and in TA maybe `0` is valid idk.

server/games/game.py
```diff
@@ -908,7 +908,7 @@ def to_dict(self):
                     player.login for player in self.players
                     if self.get_player_option(player.id, "Team") == team
                 ]
-                for team in self.teams
+                for team in self.teams if team is None
```
Askaholic, line 911: My bad `is not None` :P

Sheikah45: Yeah I should have realized ha

Askaholic (approved):

PR #810, opened 2021-07-23
Prepare dropping table `lobby_admin`

Brutus5000:


Sheikah45:
There is a bit more work to be done as currently the server is only tested against dbV112

Brutus5000:
This is a preventive measure to avoid breaking when the db is updated to a version where this table is missing.

PR #811, opened 2021-07-23
update to version v119

Sheikah45:


Sheikah45:
Could also add all the new columns if we think that is needed. They don't affect the server.

PR #820, opened 2021-08-08
Add dummy implementation for client command `match_ready` (#607)

Katharsas:
Moves #607 forward (see https://github.com/FAForever/downlords-faf-client/issues/1783#issuecomment-894844963).

Sadly untested because if currently don't have a local environment for running the python server, so i would be glad if somebody else could test that. Sorry.

### DIFF
--- a/tests/unit_tests/test_lobbyconnection.py
+++ b/tests/unit_tests/test_lobbyconnection.py
@@ -963,6 +963,7 @@ async def test_command_match_ready(lobbyconnection):
         "command": "match_ready"
     })
 
+    
 async def test_command_matchmaker_info(
     lobbyconnection,
     ladder_service,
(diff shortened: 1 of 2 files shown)


Askaholic:
Can you also add a unit test to `unit_tests/test_lobbyconnection.py`? It doesn’t need to do much just trigger the command to make sure it doesn’t raise any exceptions (see other examples in that file)

Katharsas:
Of couse, this shall present no problem for me!

tests/unit_tests/test_lobbyconnection.py
```diff
@@ -957,6 +957,10 @@ async def test_command_game_matchmaking_not_party_owner(
 
     lobbyconnection.ladder_service.cancel_search.assert_called_once()
 
+async def test_command_match_ready(lobbyconnection):
```
Askaholic, line 960: Please put 2 blank lines between functions, otherwise my linter will yell at me.

Askaholic: *Top level functions only. If they're methods of a class then only 1 line

PR #825, opened 2021-08-29
Issue/#824 support for GameOptions in matchmaker queues

### LINKED ISSUE #824: Send game options in `game_launch` message for matchmaker games
As we expand the capabilities of the team matchmaker we will need the ability to configure game options from the server. For example in larger matchmaker games (3v3, 4v4+) we may want to disable fullshare (which is currently hardcoded to always be enabled because we wanted it for 2v2).

We should add a new game launch option called `game_options` that is a dictionary mapping option names to values. For example:
```json
{
    "command": "game_launch",
    "game_options": {
        "Share": "FullShare",
        "Victory": "demoralization"
    },
    ...
}
```

Ideally we would pull these from the queue definition in the database, but for now it might be OK to hardcode them in based on team size.

The game launch options are defined here:
https://github.com/FAForever/server/blob/716717f330bf1945e4e1ca2e5e9678257985b9ad/server/types.py#L22-L30
And they are set here:
https://github.com/FAForever/server/blob/716717f330bf1945e4e1ca2e5e9678257985b9ad/server/ladder_service.py#L437-L440

Askaholic:
Loads additional params from the database and parses out game options to send in the game_launch message. My goal is for the `params` column to closely resemble the `gameInfo` table in the lua code. So something like this:

```json
{
    "GameOptions": {
        "FogOfWar": "explored",
        "UnitCap": 1000,
        "Share": "FullShare",
    },
    "GameMods": {}
}
```

But it could always be expanded in the future.

Requires https://github.com/FAForever/db/pull/285
Closes #824

### DIFF
--- a/server/ladder_service.py
+++ b/server/ladder_service.py
@@ -3,7 +3,6 @@ Manages interactions between players and matchmakers
 """
 
 import asyncio
-import json
 import random
 import re
 from collections import defaultdict
@@ -133,10 +132,12 @@ class LadderService(Service):
                 )
             elif row.map_params is not None:
                 try:
-                    params = json.loads(row.map_params)
+                    params = row.map_params
                     map_type = params["type"]
                     if map_type == "neroxis":
-                        map_list.append(NeroxisGeneratedMap.of(params, row.weight))
+                        map_list.append(
+                            NeroxisGeneratedMap.of(params, row.weight)
+                        )
                     else:
                         self._logger.warning(
                             "Unsupported map type %s in pool %s",
--- a/server/ladder_service.py
+++ b/server/ladder_service.py
@@ -446,6 +446,10 @@ class LadderService(Service):
                 game.set_player_option(player.id, "Army", slot)
                 game.set_player_option(player.id, "Color", slot)
 
+            game_options = queue.get_game_options()
+            if game_options:
+                game.gameOptions.update(game_options)
+
             mapname = re.match("maps/(.+).zip", map_path).group(1)
             # FIXME: Database filenames contain the maps/ prefix and .zip suffix.
             # Really in the future, just send a better description
@@ -455,6 +459,7 @@ class LadderService(Service):
             options = GameLaunchOptions(
                 mapname=mapname,
                 expected_players=len(all_players),
+                game_options=game_options
             )
 
             def game_options(player: Player) -> GameLaunchOptions:
--- a/server/ladder_service.py
+++ b/server/ladder_service.py
@@ -74,6 +74,7 @@ class LadderService(Service):
                     featured_mod=info["mod"],
                     rating_type=info["rating_type"],
                     team_size=info["team_size"],
+                    params=info.get("params")
                 )
                 self.queues[name] = queue
                 queue.initialize()
@@ -162,6 +163,7 @@ class LadderService(Service):
                 matchmaker_queue.c.id,
                 matchmaker_queue.c.technical_name,
                 matchmaker_queue.c.team_size,
+                matchmaker_queue.c.params,
                 matchmaker_queue_map_pool.c.map_pool_id,
                 matchmaker_queue_map_pool.c.min_rating,
                 matchmaker_queue_map_pool.c.max_rating,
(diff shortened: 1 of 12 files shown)


BlackYps:
I'm having a hard time figuring out how the server actually sends the game launch options to the client. i think mainly because my IDE can't find the declaration of this call: https://github.com/FAForever/server/blob/2b84b87e51a87dd4243db7b24756d6825368daf7/server/ladder_service.py#L456-L458

Askaholic:
Your reliance on IDE’s has been your downfall! Just grep for `def launch_game` and you’ll find it:

https://github.com/FAForever/server/blob/152242a57994fe0ac1adb8ed77503ddccf70b971/server/lobbyconnection.py#L1043

tests/integration_tests/test_game.py
```diff
@@ -121,6 +121,34 @@ async def queue_players_for_matchmaking(lobby_server, queue_name: str = "ladder1
...
+        for _ in range(num_players)
+    ])
+    protos = await asyncio.gather(*[
+        queue_player_for_matchmaking(user, lobby_server, queue_name)
+        for user in users
+    ])
+
+    # If the players did not match, this will fail due to a timeout error
+    await asyncio.gather(*[
+        read_until_command(proto, "match_found", timeout=30)
```
BlackYps, line 145: This integration test fails sometimes on my machine. I saw in the log that it took many matchmaker iterations until the logged in players finally were queued up for searching. They basically matched immediatly, but that was in the teardown log already, so I believe it was after the timeout aborted the test. Maybe increasing the timeout limit would help?
```suggestion
        read_until_command(proto, "match_found", timeout=60)
```
On the other hand I couldn't get it to reliably fail with a 10s timeout either, so I don't really know what exactly is happening. Other Integration tests also fail sometimes, but that is unrelated to this PR. It seems like it sometimes takes a lot of time for a command to take effect, for no apparent reason.

server/matchmaker/matchmaker_queue.py
```diff
@@ -49,6 +49,7 @@ def __init__(
         featured_mod: str,
         rating_type: str,
         team_size: int = 1,
+        params: Optional[Dict[str, Any]] = None,
```
BlackYps, line 52: Does it really make sense to use Any here? Ultimately these will all become strings for the command line anyway.

Askaholic: The type annotations are just for internal typechecking. They should reflect the type of the variable accurately, although I'm just a bit too lazy to do `Union[str, int, bool]` here so I just do `Any`. All this typing stuff will get a bit nicer in python 3.9

BlackYps: I actually meant why do you want to have any other types than str anyway?

tests/integration_tests/test_matchmaker.py
```diff
@@ -32,27 +33,66 @@ async def test_game_launch_message(lobby_server):
...
     msg2 = await read_until_command(proto2, "game_launch")
 
-    assert msg1["uid"] == msg2["uid"]
-    assert msg1["mod"] == msg2["mod"] == "ladder1v1"
-    assert msg1["mapname"] == msg2["mapname"]
-    assert msg1["team"] == 2
+    assert msg2["uid"] == msg1["uid"]
+    assert msg2["mod"] == msg1["mod"]
+    assert msg2["mapname"] == msg1["mapname"]
     assert msg2["team"] == 3
```
BlackYps, line 39: Why are they team 2 and 3 instead of 1 and 2?

Askaholic: Team 1 is the ffa team (shown as "-" in the lobby)

BlackYps (approved):

server/matchmaker/matchmaker_queue.py
```diff
@@ -83,6 +85,9 @@ def get_map_pool_for_rating(self, rating: int) -> Optional[MapPool]:
                 continue
             return map_pool
 
+    def get_game_options(self) -> Dict[str, Any]:
+        return self.params.get("GameOptions") or None
```
Wesmania, line 89: `get` already returns `None` when a key is missing

server/ladder_service.py
```diff
@@ -182,6 +184,7 @@ async def fetch_matchmaker_queues(self, conn):
             info["mod"] = row.gamemod
             info["rating_type"] = row.rating_type
             info["team_size"] = row.team_size
+            info["params"] = json.loads(row.params) if row.params else None
```
Wesmania, line 187: Catch json parse errors

server/matchmaker/matchmaker_queue.py
Askaholic: Heh, yea but not if GameOptions is `{}` 😉

server/ladder_service.py
Askaholic: Good call

server/matchmaker/matchmaker_queue.py
Askaholic: I guess because I still expect the params object to contain any valid json value

server/ladder_service.py
```diff
@@ -133,10 +134,13 @@ async def fetch_map_pools(self, conn) -> Dict[int, Tuple[str, List[Map]]]:
...
                 try:
-                    params = json.loads(row.map_params)
+                    params = row.map_params
                     map_type = params["type"]
                     if map_type == "neroxis":
-                        map_list.append(NeroxisGeneratedMap.of(params, row.weight))
+                        map_list.append(NeroxisGeneratedMap.of(
+                            params,
+                            row.weight
+                        ))
```
Askaholic, line 143: This might be better
```suggestion
                        map_list.append(
                            NeroxisGeneratedMap.of(params, row.weight)
                        )
```

tests/integration_tests/test_game.py
Askaholic: Yea I think it has to do with the fast forwarding somehow. Is it really timing out here though? This should only start waiting once all of the players have successfully queued, because `queue_player_for_matchmaking` waits for the `search_info` start message before returning.

Askaholic: Well, I haven't gotten these to fail, so I'm just going to leave it as is for now.

BlackYps: I'll just keep an eye out on this. Maybe it has to do with the local configuration on my machine...

PR #830, opened 2021-09-03
Fix misc errors that show up in the logs

Askaholic:
- Better logging of errors with reported JsonStats
- Error handling if hydra jwk endpoint is unreachable
- Better handling of game timeout
- Fix error in `deadlock_retry_execute` exception handler

### DIFF
--- a/server/db/__init__.py
+++ b/server/db/__init__.py
@@ -57,7 +57,8 @@ async def deadlock_retry_execute(conn, *args, max_attempts=3):
         try:
             return await conn.execute(*args)
         except OperationalError as e:
-            if any(msg in e.message for msg in (
+            error_text = str(e)
+            if any(msg in error_text for msg in (
                 "Deadlock found",
                 "Lock wait timeout exceeded"
             )):
--- a/server/gameconnection.py
+++ b/server/gameconnection.py
@@ -4,6 +4,7 @@ Game communication over GpgNet
 
 import asyncio
 import contextlib
+import json
 from typing import Any, List
 
 from sqlalchemy import select, text
@@ -362,7 +363,14 @@ class GameConnection(GpgNetServerProtocol):
                 self.game.leaderboard_saved = True
 
     async def handle_json_stats(self, stats: str):
-        self.game.report_army_stats(stats)
+        try:
+            self.game.report_army_stats(stats)
+        except json.JSONDecodeError:
+            self._logger.warning(
+                "Malformed game stats reported by %s: '...%s'",
+                self._player.login,
+                stats[-20:]
+            )
 
     async def handle_enforce_rating(self):
         self.game.enforce_rating = True
--- a/server/gameconnection.py
+++ b/server/gameconnection.py
@@ -479,11 +479,15 @@ class GameConnection(GpgNetServerProtocol):
             if self.player.state != PlayerState.HOSTING:
                 return
 
-            self._logger.info(
-                "Launching game %s in state %s",
-                self.game,
-                self.game.state
-            )
+            if self.game.state is not GameState.LOBBY:
+                self._logger.warning(
+                    "Trying to launch game %s in invalid state %s",
+                    self.game,
+                    self.game.state
+                )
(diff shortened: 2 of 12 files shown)


server/games/game.py
```diff
@@ -269,17 +265,16 @@ def get_team_sets(self) -> List[Set[Player]]:
 
     async def wait_hosted(self, timeout: float):
         return await asyncio.wait_for(
-            asyncio.shield(self._is_hosted),
```
Askaholic, line 272: Might want to keep the call to `shield`

server/db/__init__.py
```diff
@@ -57,7 +57,8 @@ async def deadlock_retry_execute(conn, *args, max_attempts=3):
         try:
             return await conn.execute(*args)
         except OperationalError as e:
-            if any(msg in e.message for msg in (
+            error_text = str(e)
+            if any(msg in error_text for msg in (
```
Cheyans, line 61: I'm curious, what's more advantageous about using `str(e)` over referencing the error message directly?

server/gameconnection.py
```diff
@@ -362,7 +363,14 @@ async def handle_operation_complete(
...
 
     async def handle_json_stats(self, stats: str):
-        self.game.report_army_stats(stats)
+        try:
+            self.game.report_army_stats(stats)
+        except json.JSONDecodeError:
+            self._logger.warning(
+                "Malformed game stats reported by %s: '...%s'",
+                self._player.login,
+                stats[-20:]
```
Cheyans, line 372: I'm obviously not familiar with this codebase, but `stats[-20:]` seem quite ambiguous. What is special about the last 20 elements of this object? Maybe some comments would be useful here.

server/gameconnection.py
```diff
@@ -471,11 +479,15 @@ async def handle_game_state(self, state: str):
...
-            )
+            if self.game.state is not GameState.LOBBY:
+                self._logger.warning(
+                    "Trying to launch game %s in invalid state %s",
+                    self.game,
+                    self.game.state
+                )
+                return
+
+            self._logger.info("Launching game %s", self.game)
```
Cheyans, line 490: I think including which player is launching the game could be useful.

server/ladder_service.py
```diff
@@ -467,10 +467,16 @@ def game_options(player: Player) -> GameLaunchOptions:
...
             await game.wait_launched(60 + 10 * len(all_guests))
             self._logger.debug("Ladder game launched successfully")
-        except Exception:
+        except Exception as e:
+            if isinstance(e, asyncio.TimeoutError):
+                self._logger.info(
+                    "Failed to start ladder game! Game setup timed out"
+                )
+            else:
+                self._logger.exception("Failed to start ladder game!")
```
Cheyans, line 476: Similar comment as above, I think in including the player ids/names could be useful here.

server/db/__init__.py
Askaholic: It doesn't throw an attribute error xD. I think `str(e)` is the only standard way to get the error message from an exception, exceptions don't have a `message` attribute.

server/gameconnection.py
Askaholic: Just that when these errors happens, it seems like it's always due to truncated data. Having the last few characters just lets you confirm that by seeing if the json structure is terminated or not.

server/gameconnection.py
Askaholic: It's always the host, and the host is included in the string representation of `self.game`

server/ladder_service.py
Askaholic: Yea that's probably true

Wesmania (approved):
Looks alright to me.

server/games/game.py
Askaholic: I tested it and it doesn't seem to matter. I think I added shield before because it prevented some exception from showing up in the logs when the server was shutting down, but that doesn't seem to happen with events.

PR #831, opened 2021-09-06
Issue/#703 update game info when player disconnects

### LINKED ISSUE #703: Implement a way for the party leader to determine whether party members are still in-game
I just had @BlackYps fill me in on the current state of TMM and he brought this up as an open issue but we didn't find a ticket:

It seems that when a player exits a game (or maybe just a TMM game) early, they cannot queue up for TMM again before the game has finished since the party leader will believe they are in an active game. That seems to be because the party leader checks the player lists of active games in `game_info` to determine whether a party member is still in-game, but this list shows all players present at game start and not players currently in the game.

As far as I'm aware there is currently no way for the server to know that a player has left a game before it ended. BlackYps opened https://github.com/FAForever/downlords-faf-client/issues/2125 for a message to be sent by the client in that case.
To resolve this it seems likely that either the player lists in `game_info` need to be updated on such messages or a player in-game status that is broadcasted should be added.

Askaholic:
Changes the `game_info` message to use connected players instead of players frozen at game launch. This should allow the client to tell when a player has left the game even before the game has completely ended.

Closes #703

### DIFF
--- a/server/games/game.py
+++ b/server/games/game.py
@@ -93,7 +93,7 @@ class Game():
         self.map_file_path = f"maps/{map_}.zip"
         self.map_scenario_path = None
         self.password = None
-        self._players = []
+        self._players_at_launch: List[Player] = []
         self.AIs = {}
         self.desyncs = 0
         self.validity = ValidityState.VALID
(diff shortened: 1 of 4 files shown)


Askaholic:
Tested on the test server and it looks like it works as expected.

PR #833, opened 2021-09-11
Change game quality calculations

BlackYps:
Here we go again...

When looking at some real games I noticed two problems with the current implementation:
- If the uniformity and the fairness are both very bad (close to 0) and at least one is negative, we get a better score (slightly below 0) than if one is negative and the other one is really good (close to 1). Then we get a score that is way more negative.
- We don't care so much if the rating disparity is exactly zero, but we want to discourage games where uniformity or fairness are way off.

We can fix the first one by not multiplying but adding uniformity and fairness. That way one getting worse will always worsen the game quality.
For the second one we can simply square the penalty for deviating from the ideal values. This makes small errors smaller and increases big errors.
With these changes I have to tune the config values again, I will update this PR when I made some more tests.

### DIFF
--- a/server/matchmaker/algorithm/team_matchmaker.py
+++ b/server/matchmaker/algorithm/team_matchmaker.py
@@ -1,6 +1,7 @@
 import logging
 import statistics
 from collections import defaultdict
+from math import sqrt
 from typing import Dict, Iterable, List, NamedTuple, Set, Tuple
 
 from sortedcontainers import SortedList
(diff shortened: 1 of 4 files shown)


BlackYps:
I opted for a distance calculation instead. This one gives rather nice results. I will make a forum post with some comparison graphs soon.

server/matchmaker/algorithm/team_matchmaker.py
```diff
@@ -281,17 +282,17 @@ def assign_game_quality(self, match: Match, team_size: int) -> GameCandidate:
...
+        unfairness = rating_disparity / config.MAXIMUM_RATING_IMBALANCE
         deviation = statistics.pstdev(ratings)
-        uniformity = 1 - (deviation / config.MAXIMUM_RATING_DEVIATION)
+        rating_variety = deviation / config.MAXIMUM_RATING_DEVIATION
 
-        quality = fairness * uniformity
-        if fairness < 0 and uniformity < 0:
-            quality *= -1
-        quality += newbie_bonus + time_bonus
+        # Visually this creates a cone in the unfairness-rating_variety plane that slowly raises with the time boni.
```
Askaholic, line 289: Can we cut this line off at 79 chars? Also do you really mean 'boni'?

Askaholic: I mean, like continue the comment on the next line after 79 chars

tests/data/test-data.sql
```diff
@@ -132,7 +132,7 @@ insert into leaderboard_rating (login_id, mean, deviation, total_games, leaderbo
   (102, 1500, 500, 0, 1),
   (102, 1500, 500, 0, 2),
   (105, 1400, 150, 20, 3),
-  (106, 1500, 75, 20, 3)
+  (106, 900, 75, 20, 3)
```
Askaholic, line 135: Why is this change needed again?

BlackYps: The new algorithm doesn't match newbies completely randomly. If the game is too unbalanced the players don't get matched at all. The integration test placed two 0 rated newbies against a 950 and a 1275 rated player. From the feedback I gathered the players don't find matches like these desireable. 
I lowered the rating to 950 and 675. This is still bad and at the edge of matching at all, but matches after three rounds when the maximum newbie time bonus is reached.
It passed before, because many really bad matchups would unintentionally get a quality rating of just below 0 instead of a way negative one.

Askaholic (approved):

PR #835, opened 2021-09-13
Serialize faction as string in game launch message

### LINKED ISSUE #834: Use string for faction in game launch message
It would be nice to consistently use the string version of faction in all the messages from the server. All clients going back to 1.4.6 are able to handle the faction as a string.

BlackYps:
Closes #834

### DIFF
--- a/server/ladder_service.py
+++ b/server/ladder_service.py
@@ -442,7 +442,7 @@ class LadderService(Service):
             def game_options(player: Player) -> GameLaunchOptions:
                 return options._replace(
                     team=game.get_player_option(player.id, "Team"),
-                    faction=player.faction,
+                    faction=player.faction.name,
                     map_position=game.get_player_option(player.id, "StartSpot")
                 )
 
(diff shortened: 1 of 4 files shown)


BlackYps:
One integration test is broken, but I don't know why, because it seems totally unrelated to me.
test_game_ratings in test_teammatchmaker.py fails to finish. I found this in the log:
` WARNING  RatingService:rating_service.py:293 gps_update_sql resultset is empty for game_id 41956`
It seems like the game_player_stats table fails to update correctly.

Sheikah45:
You need to make sure the integer value is still used for faction to update the game_stats entries

BlackYps:
What are we going to do with this, given that we want to reduce the number of open PRs, but v2 is probably a long time away?

Askaholic:
Well, since I’m not actively working on the server much anymore we should probably release a less ambitious version of the v2 changes that just includes what we have so far. We can always do a v3 later if more stuff gets done.

I have merged some breaking change stuff onto the `v2-develop` branch which I then occasionally go and rebase onto `develop`. We could merge this PR there too and release any time since it sounds like the official client can handle this already. So if you rebase this onto `v2-develop` and change the base branch to that, we can merge.

PR #836, opened 2021-09-20
Add pre-commit configuration to the project

Gatsik:
Hopefully this will help the maintainers spend less time on reviewing some trivial things like typos and good code style.

At the moment. some basic hooks have been added, although [there are a lot more of them](https://pre-commit.com/hooks.html). Perhaps you'd like to add something.

Also not sure about max-line-lenght (`E501` in `flake8`'s black list). But for now probably it is better to keep it as there are a lot of changes to be done to fix this.

There is the [pre-commit GitHub Action](https://github.com/pre-commit/action) that also could be added to the checks.

### DIFF
--- a/Pipfile.lock
+++ b/Pipfile.lock
@@ -848,6 +848,14 @@
             "markers": "python_version >= '2.7' and python_version not in '3.0, 3.1, 3.2, 3.3'",
             "version": "==0.13.1"
         },
+        "pre-commit": {
+            "hashes": [
+                "sha256:a4ed01000afcb484d9eb8d504272e642c4c4099bbad3a6b27e519bd6a3e928a6",
+                "sha256:3c25add78dbdfb6a28a651780d5c311ac40dd17f160eb3954a0c59da40a505a7"
+            ],
+            "index": "pypi",
+            "version": "==2.15.0"
+        },
         "py": {
             "hashes": [
                 "sha256:21b81bda15b66ef5e1a777a21c4dcd9c20ad3efd0b3f817e7a809035269e1bd3",
--- a/Pipfile.lock
+++ b/Pipfile.lock
@@ -848,14 +848,6 @@
             "markers": "python_version >= '2.7' and python_version not in '3.0, 3.1, 3.2, 3.3'",
             "version": "==0.13.1"
         },
-        "pre-commit": {
-            "hashes": [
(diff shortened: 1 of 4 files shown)


Askaholic:
Nice idea! I haven’t used pre-commit before, but it certainly can’t hurt.

I’ve been kindof lenient on the line lengths because there are arguments for allowing long lines (like most monitors are huge) but personally I develop on a 13" laptop and I’ve turned the screen resolution down so I don’t have to squint in order to see things, so typically I can only see about 80 characters before I need to scroll.

Pipfile.lock
```diff
@@ -848,6 +848,14 @@
             "markers": "python_version >= '2.7' and python_version not in '3.0, 3.1, 3.2, 3.3'",
             "version": "==0.13.1"
         },
+        "pre-commit": {
```
Askaholic, line 851: I’m not sure this should be in the lock file… Did this get added automatically?

Gatsik: no, it was added manually. thought it was kind of `requirements.txt` equivalent, but probably you are right, it is not required

Askaholic: The `Pipfile` is where you put in dependencies, like requirements.txt, `Pipfile.lock` is autogenerated by running `pipenv lock`. But I wouldn't put pre-commit in there since it's not required for the build.

Gatsik:
Not sure how to modify README.md (is it needed?)
`pipenv install pre-commit --skip-lock` will modify Pipfile, and i haven't found any way to avoid it. The only way to install the package without changing the Pipfile I came up with:
```
$ pipenv shell
$ python3 -m pip install pre-commit
```

Askaholic:
Isn’t `pre-commit` something you would install system wide? Pipenv is just for managing the virtual environment for this project. I would think you’d just install it with pip outside of the virtualenv.

```
python3 -m pip install pre-commit
```

README.md
```diff
@@ -111,6 +111,17 @@ You can check for possible unused code with `vulture` by running:
 It tends to produce a lot of false positives, but it can provide a good place
 to start.
 
+To identify simple issues before submission to code review using `pre-commit`,
+install it and set up the git hook scripts by running:
```
Askaholic, line 115: Can we change the wording to

> There are some `pre-commit` hooks that can fix basic formatting issues for you to make the review process go smoother. You can install them by running:

I would also put the section about `pre-commit` before the section about vulture, since vulture is a kindof niche thing.

PR #837, opened 2021-09-21
Fix some code style issues

Gatsik:
There are left some `# noqa:`s, and currently I have no idea how to properly fix them


Gatsik:
Irony: by fixing flake8's style issues (F821 undefined name) I added Codacy Static Code Analysis issues. FeelsBadMan

server/games/custom_game.py
```diff
@@ -1,4 +1,3 @@
-import asyncio
```
Gatsik, line 1: and why this was not marked as unused import before?

server/games/coop.py
```diff
@@ -1,5 +1,3 @@
-import asyncio
```
Gatsik, line 1: and this

server/broadcast_service.py
```diff
@@ -9,6 +13,9 @@
 from .player_service import PlayerService
 from .timing import LazyIntervalTimer
 
+if TYPE_CHECKING:
```
Askaholic, line 16: I don't think the if check is needed here. At least flake8 doesn't generate unused import warnings here. Hopefully its the same on codacy

server/gameconnection.py
```diff
@@ -616,25 +616,25 @@ def __str__(self):
...
-    "Bottleneck":           GameConnection.handle_bottleneck,
-    "BottleneckCleared":    GameConnection.handle_bottleneck_cleared,
-    "Disconnected":         GameConnection.handle_disconnected,
-    "IceMsg":               GameConnection.handle_ice_message,
-    "Chat":                 GameConnection.handle_chat,
-    "GameFull":             GameConnection.handle_game_full
+    "Desync": GameConnection.handle_desync,
+    "GameState": GameConnection.handle_game_state,
+    "GameOption": GameConnection.handle_game_option,
+    "GameMods": GameConnection.handle_game_mods,
```
Askaholic, line 622: Nooo. This is so much worse xD

server/games/custom_game.py
Askaholic: Actually I am seeing an unused import warning for this using flake8. It might have been in there for a long time though, like since before we added codacy?

server/games/game.py
```diff
@@ -1,9 +1,21 @@
+from __future__ import annotations
```
Askaholic, line 1: Oh does this add forward references? That's pretty nice, although I'm hoping to migrate to python 3.9 soon.

server/lobbyconnection.py
```diff
@@ -1065,7 +1066,7 @@ async def command_modvault(self, message):
                                    downloads=downloads, date=int(date.timestamp()), uid=uid, name=name, version=version, author=author,
                                    ui=ui)
                         await self.send(out)
-                    except:
+                    except BaseException:
```
Askaholic, line 1069: Can we just do `Exception` instead?

tests/integration_tests/test_oauth_session.py
```diff
@@ -34,7 +34,7 @@ def oauth2_server(event_loop):
     @routes.post("/token")
     async def token(request):
         data = await request.post()
-        return await {
+        return await {  # noqa: W606
```
Askaholic, line 37: Oh yea this is a funny one

Askaholic: TBH we could probably define the dict as a class attribute

tests/unit_tests/test_profiler.py
```diff
@@ -145,7 +145,7 @@ async def test_profiler_refresh_cancels():
     profiler.refresh()
     await asyncio.sleep(10)
 
-    assert profiler._running == False
+    assert profiler._running is False
```
Askaholic, line 148: Wow, I didn't think we had any of these in the codebase still :(

Askaholic (commented):
Awesome work, these are some great fixes! Looking forward to merging this.

The only ones I'm not a fan of are the conditional imports. I don't think they're needed and they're going to mess with isort as well as code coverage.

server/games/game.py
Gatsik: I believe this will become the default behavior only in python 3.10

server/broadcast_service.py
Gatsik: The check is needed because of import cycle. If you remove import, then the undefined name error will appear.

Askaholic: Huh, then how is it possible for the type checker to import it?

Gatsik: Type checkers assume this constant to be `True`.

Askaholic: Yea I know, so wouldn’t the type checker still experience an import cycle?

Gatsik: No, type checker wouldn't experience an import cycle

Askaholic: Why?

Gatsik: idk, maybe because type checkers don't run the code? ([stack overflow answer](https://stackoverflow.com/questions/61545580/how-does-mypy-use-typing-type-checking-to-resolve-the-circular-import-annotation))

server/games/game.py
Gatsik: Or even 3.11: [PEP 563: Correct references to 3.10 default adoption](https://github.com/python/peps/pull/2086)

Askaholic:
Superceeded by #881

PR #839, opened 2021-09-23
Optimize rating initialization

Askaholic:
I had a look at the server profile generated after the last update, and it looks like the rating initialization was being called an excessive amount. This is mostly because it gets invoked both when generating `game_info` as well as `player_info` messages. 

In the `game_info` messages it would be called even if the rating data wasn't actually needed, just because it had been easier to write that ways. Now, it checks whether rating enforcement is enabled first and only queries the rating if it is. I also added some caching to the rating computation itself so that the queries generated by the `player_info` message creation will now short circuit the majority of the time.

### DIFF
--- a/server/games/game.py
+++ b/server/games/game.py
@@ -861,17 +861,22 @@ class Game():
         self._process_pending_army_stats()
 
     def is_visible_to_player(self, player: Player) -> bool:
+        """
+        Determine if a player should see this game in their games list.
+
+        Note: This is a *hot* function, it can have significant impacts on
+        performance.
+        """
         if self.host is None:
             return False
 
         if player == self.host or player in self._connections:
             return True
 
-        mean, dev = player.ratings[self.rating_type]
-        displayed_rating = mean - 3 * dev
         if (
             self.enforce_rating_range
-            and displayed_rating not in self.displayed_rating_range
+            and player.get_displayed_rating(self.rating_type)
+            not in self.displayed_rating_range
         ):
             return False
 
--- a/tests/unit_tests/test_rating.py
+++ b/tests/unit_tests/test_rating.py
@@ -1,3 +1,5 @@
+from unittest import mock
+
 import pytest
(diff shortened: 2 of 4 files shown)


server/games/game.py
```diff
@@ -861,17 +861,22 @@ def report_army_stats(self, stats_json):
         self._process_pending_army_stats()
 
     def is_visible_to_player(self, player: Player) -> bool:
+        """
+        Determine if a player should see this game in their games list.
+
+        Note: This is a *hot* function, it can have significant impacts on
+        performance.
```
BlackYps, line 868: Is it because of the rating initialization? It doesn't become clear, why this comment on performance is here.

tests/unit_tests/test_rating.py
```diff
@@ -135,6 +137,16 @@ def test_initialization_transient(chained_ratings):
     assert chained_ratings["tmm_2v2"] == (300, 200)
 
 
+def test_initialization_transient_caching(chained_ratings):
+    # Return a new mock on each call. By default it will return the same mock
+    chained_ratings._get_initial_rating = mock.Mock(side_effect=mock.Mock)
+
+    chained_ratings["global"] = (1000, 50)
+
+    assert chained_ratings["tmm_2v2"] == chained_ratings["tmm_2v2"]
```
BlackYps, line 146: Is this a typo? How will this ever fail?

server/games/game.py
Askaholic: It’s there because this is now the second time I’ve had to optimize something that was added to that function because it showed up in the profiler.

tests/unit_tests/test_rating.py
Askaholic: This checks that the rating is cached and not recomputed. If it were recomputed we would get different mock objects for the two calls.

BlackYps (approved):

PR #840, opened 2021-09-23
Check for lobby scope in token during login

Sheikah45:
This asserts that the oauth token contains the lobby scope before allowing a player to login. This allows for players to be banned from just the game service. All oauth tokens from the clients currently have the lobby scope

### DIFF
--- a/server/oauth_service.py
+++ b/server/oauth_service.py
@@ -87,7 +87,10 @@ class OAuthService(Service, name="oauth_service"):
             )
 
             if "lobby" not in decoded["scp"]:
-                raise AuthenticationError("Token does not have permission to login to the lobby server", "token")
+                raise AuthenticationError(
+                    "Token does not have permission to login to the lobby server",
+                    "token"
+                )
 
             return int(decoded["sub"])
         except (InvalidTokenError, KeyError, ValueError):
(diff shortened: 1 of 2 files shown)


Sheikah45:
Hmmm not sure why that test is failing because it uses the old password login

Sheikah45:
When run by itself it passes just fine

Askaholic:
Just a flakey test

Brutus5000:
Also this is super important for enabling Gog link :rocket: wheeeeee

Brutus5000:
Oh wait. This PR should instead remove the check for Steam link as this moves to the user service.

Askaholic:
The steam link check only happens when you login with password

server/oauth_service.py
```diff
@@ -85,6 +85,10 @@ async def get_player_id_from_token(self, token: str) -> int:
                 algorithms="RS256",
                 options={"verify_aud": False}
             )
+
+            if "lobby" not in decoded["scp"]:
+                raise AuthenticationError("Token does not have permission to login to the lobby server", "token")
```
Askaholic, line 90:
```suggestion
                raise AuthenticationError(
                    "Token does not have permission to login to the lobby server", 
                    "token"
                )
```

Askaholic:
So I'm assuming when you try to log in via the API and you're banned, you see the ban reason on the webpage?

Brutus5000:
Correct

Askaholic (approved):

PR #844, opened 2021-09-27
Allow map size of the generator params to be non power of 2

Sheikah45:
Only enforce that the param size needs to be a multiple of 64 as specified by the map name encoding of the generator

### DIFF
--- a/server/types.py
+++ b/server/types.py
@@ -57,7 +57,7 @@ class NeroxisGeneratedMap(NamedTuple):
         if map_size_pixels <= 0:
             raise Exception("Map size is zero or negative")
 
-        if map_size_pixels // 64 != 0:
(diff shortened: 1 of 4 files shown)


server/types.py
```diff
@@ -57,8 +57,8 @@ def of(cls, params: dict, weight: int = 1):
         if map_size_pixels <= 0:
             raise Exception("Map size is zero or negative")
 
-        if map_size_pixels & (map_size_pixels - 1) != 0:
-            raise Exception("Map size is not a power of 2")
+        if map_size_pixels // 64 != 0:
```
Askaholic, line 60: Don't you mean 
```suggestion
        if map_size_pixels % 64 != 0:
```
Also add a test?

Askaholic (approved):

PR #816, opened 2021-08-01
v2 - Remove Deprecations

Askaholic:
Remove all the deprecated stuff

Closes #606

PR #848, opened 2021-10-13
Remove search boundaries

BlackYps:
I guess they were used by the python client for the "A person in your rating range is searching for a game. Click here to join them." message. However, the ranges are unreliable, because it just gives a hardcoded range and the trueskill quality also depends on the deviation. Also we alter the acceptable quality for every unsucessful queue pop. For team queues we use a different system entirely.
The java client doesn't use them for anything at the moment.

### DIFF
--- a/server/matchmaker/matchmaker_queue.py
+++ b/server/matchmaker/matchmaker_queue.py
@@ -257,9 +257,6 @@ class MatchmakerQueue:
         self._is_running = False
 
     def to_dict(self):
-        """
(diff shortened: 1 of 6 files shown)


Askaholic:
Btw there is a `v2-develop` branch where I’m going to merge breaking changes to. It’s a bit behind though since we had some more releases after I created it, so when I find the time it going to fast forward it again. But, any v2 prs should be made against that branch.

Askaholic:
Superseded by #816

PR #849, opened 2021-10-17
Issue/#845 Make matchmaker games also affect global rating

### LINKED ISSUE #845: Adjust global rating with ranked matchmaker games
Now that we're adding a number of different matchmakers, we want to make it so that new players can use the matchmaker to establish a global rating for themselves. To do this, we want to apply rating changes from the matchmaker back to global, but only if the rating changes are positive (to prevent rating manipulation and other issues).

The way it would work is:
<details>
<summary>Proposal 1</summary>
Whenever a game is rated and the rating used is not global and not hidden, then compute the normalized adjustment made by trueskill to the rating i.e. `(after_mean - before_mean) / before_mean` and `(after_dev - before_dev) / before_dev)`. If the new displayed rating is higher than the displayed global rating, compute the adjusted global rating by applying the normalized changes to the global rating, scaled by the global rating i.e. `global_mean += global_mean * normalized_mean_change` and `global_dev += global_dev * normalized_dev_change`. If the displayed adjusted global rating is higher than the displayed old global rating, save the rating adjustments, otherwise leave the global rating unchanged.

### Example 1
Suppose a player's global rating and matchmaker ratings are both `(1500, 500)` and they **LOSE** a matchmaker game where their new rating returned by trueskill is `(1400, 400)`. Since the new displayed rating `(1400 - 3 * 400) = 200` is higher than the displayed global rating `(1500 - 3 * 500) = 0`, we perform the adjustment as follows: 
- Normalized mean change is `(1400 - 1500) / 1500 = -0.067`
- Normalized deviation change is `(400 - 500) / 500 = -0.200`. 
- Global mean adjustment is `1500 * -0.067 = -100`
- Global deviation adjustment is  `500 * -0.200 = -100`

The adjusted global mean is then computed to be `(1500 - 100, 500 - 100) = (1400, 400)` which is displayed as `200`. Since this is higher than the original displayed global rating `0`, the global rating adjustments are saved.

### Example 2
Suppose a player's global rating is `(1000, 150)` and their matchmaker rating is `(1200, 100)` and they **WIN** a matchmaker game where their new rating returned by trueskill is `(1205, 95)`. Since the new displayed rating `(1205- 3 * 95) = 920` is higher than the displayed global rating `(1000 - 3 * 150) = 550`, we perform the adjustment as follows: 
- Normalized mean change is `(1205 - 1200) / 1200 = 0.0042`
- Normalized deviation change is `(95 - 100) / 100 = -0.0500`. 
- Global mean adjustment is `1000 * 0.0042 = 4.2`
- Global deviation adjustment is  `150 * -0.0500 = -7.500`

The adjusted global mean is then computed to be `(1000 + 4.2, 150 - 7.500) = (1004.2, 142.500)` which is displayed as `567.2`. Since this is higher than the original displayed global rating `550`, the global rating adjustments are saved.

### Example 3
Suppose a player's global rating is `(1000, 150)` and their matchmaker rating is `(1200, 100)` and they **LOSE** a matchmaker game where their new rating returned by trueskill is `(1195, 105)`. Since the new displayed rating `(1195 - 3 * 105) = 880` is higher than the displayed global rating `(1000 - 3 * 150) = 550`, we perform the adjustment as follows: 
- Normalized mean change is `(1195 - 1200) / 1200 = -0.0042`
- Normalized deviation change is `(105 - 100) / 100 = 0.0500`. 
- Global mean adjustment is `1000 *- 0.0042 = -4.2`
- Global deviation adjustment is  `150 * 0.0500 = 7.500`

The adjusted global mean is then computed to be `(1000 - 4.2, 150 + 7.500) = (995.8, 157.500)` which is displayed as `523.3`. Since this is lower than the original displayed global rating `550`, the global rating adjustments are discarded.
</details>

<details>
<summary>Proposal 2</summary>
Whenever a game is rated and the rating used is not global and not hidden, then compute the adjusted global mean by replacing the player's non-global rating with the global mean in the call to the trueskill rating adjustments.

### Scenario 1
Suppose a player's global rating and matchmaker ratings are the same, then the new global rating will exactly equal the matchmaker rating, as the call to trueskill is unchanged.

### Scenario 2 (1vs1)
- MM Rating: `(800, 150)` displayed `350`
- Global rating: `(1000, 150)` displayed `550`
- MM Opponent Rating: `(800, 150)` displayed `350`

Player **WINS**:
- New MM rating: `(848.918, 143.182)` displayed `419`
- New Global rating: `(1031.924, 144.475)` displayed `598`

Player **LOSES**:
- New MM rating: `(751.082, 143.182)` displayed `322`
- New Global rating: `(931.189, 142.189)` displayed `505`
</details>

Askaholic:
I implemented proposal 2 with the following restrictions: 
- The rating adjustment will only ever increase the displayed rating.
- The rating adjustment only happens when global rating is lower than matchmaker rating.
- The rating adjustment maxes out at a certain displayed rating as set in the configuration. 

This should allow players to build up a rating by playing matchmaker but also prevent the total abuse of the system by established players.

Notes:
- Need to skip the game_player_stats update for the secondary rating.
- Eventually we should get rid of the game_player_stat update completely and drop those columns.

Closes #845

### DIFF
--- a/tests/unit_tests/test_game_rating.py
+++ b/tests/unit_tests/test_game_rating.py
@@ -1098,6 +1098,11 @@ async def do_test_rating_adjustment(
         assert result.outcomes == {**team1_outcomes, **team2_outcomes}
 
 
+# These ratings show up a lot because our tests have a lot of new players
+NEWBIE_1V1_WINNER = (1766, 429)
+NEWBIE_1V1_LOSER = (1235, 429)
+
+
 async def test_rating_adjustment_1v1_newbies(rating_service, player_factory):
     IGNORED = object()
 
@@ -1110,14 +1115,14 @@ async def test_rating_adjustment_1v1_newbies(rating_service, player_factory):
             PersistedResults(
                 rating_type=RatingType.LADDER_1V1,
                 ratings={
-                    1: pytest.approx((1766, 429), abs=1),
-                    2: pytest.approx((1235, 429), abs=1),
+                    1: pytest.approx(NEWBIE_1V1_WINNER, abs=1),
+                    2: pytest.approx(NEWBIE_1V1_LOSER, abs=1),
                 },
                 outcomes=IGNORED
             ),
             PersistedResults(
                 rating_type=RatingType.GLOBAL,
-                ratings={1: pytest.approx((1766, 429), abs=1)},
+                ratings={1: pytest.approx(NEWBIE_1V1_WINNER, abs=1)},
                 outcomes=IGNORED
             )
         ]
@@ -1141,8 +1146,8 @@ async def test_rating_adjustment_1v1_ladder_newbies_global_pros(
             PersistedResults(
                 rating_type=RatingType.LADDER_1V1,
                 ratings={
-                    1: pytest.approx((1766, 429), abs=1),
-                    2: pytest.approx((1235, 429), abs=1),
+                    1: pytest.approx(NEWBIE_1V1_WINNER, abs=1),
+                    2: pytest.approx(NEWBIE_1V1_LOSER, abs=1),
                 },
                 outcomes=IGNORED
             )
@@ -1305,7 +1310,7 @@ async def test_rating_adjustment_1v1_ladder_pro_vs_global_pro(
             PersistedResults(
                 rating_type=RatingType.GLOBAL,
                 ratings={
-                    1: pytest.approx((1766, 429), abs=1),
+                    1: pytest.approx(NEWBIE_1V1_WINNER, abs=1),
                 },
                 outcomes=IGNORED
             )


server/rating_service/rating_service.py
```diff
@@ -148,192 +143,353 @@ async def _handle_rating_queue(self) -> None:
...
             outcome_map
         )
+
+        # TODO: If we add hidden ratings, make sure to check for them here.
+        # Hidden ratings should not affect global.
+        if summary.rating_type == RatingType.GLOBAL:
+            return
+
+        # Now perform global rating adjustments.
+        async with self._db.acquire() as conn:
```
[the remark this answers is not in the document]
Askaholic: I think I'm going to leave it as is so that we're not holding the database connection open while posting to rabbitmq. Not sure if it really matters, but it just seems like a good idea to me to drop the connection while we're not using it and instead waiting on rabbitmq messages and then acquire a new connection after when we do the global rating adjustment.

server/rating_service/rating_service.py
```diff
@@ -148,192 +143,353 @@ async def _handle_rating_queue(self) -> None:
...
         if self._rating_type_ids is None:
             self._logger.warning(
-                "Tried to fetch player data before initializing service."
+                "Tried to write player data before initializing service."
             )
             raise ServiceNotReadyError("RatingService not yet initialized.")
 
-        rating_type_id = self._rating_type_ids.get(rating_type)
-        if rating_type_id is None:
-            raise ValueError(f"Unknown rating type {rating_type}.")
```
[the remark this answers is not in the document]
Askaholic: I'm not sure it really matters. It just means we'll raise a `KeyError` on this line `self._rating_type_ids[rating_type]` instead of raising a `ValueError`. I guess it was there for nicer logging in the exception handler. I suppose I could put it back in.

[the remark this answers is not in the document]
Askaholic: I just refactored the checking to happen at the very start of `_rate`. Since the other functions are private we can expect them to be called only in the places where we call them internally, so we know that this check will always have happened.

server/rating_service/game_rater.py
```diff
@@ -15,32 +17,99 @@ class GameRatingError(Exception):
...
+        self.base_ratings = base_ratings
+
+    def compute_rating(
+        self,
+        ratings: Dict[PlayerID, Rating]
+    ) -> Dict[PlayerID, Rating]:
+        """
+        Perform rating adjustment based on another set of ratings. For
+        each player, this will rate the game with trueskill as if their
+        rating for the base rating type were equal to the rating we are
```
BlackYps, line 75:
```suggestion
        each player, this will rate the game with trueskill as if they
        played this game with the rating we are adjusting instead of the
        base rating.
```
This makes it clearer what is happening imo

BlackYps (commented):
I didn't have time yet to look at the tests. I will follow up with them later

server/rating_service/game_rater.py
Askaholic: Haha, yea I refactored it to be more generic and then failed epically to update the comments

tests/unit_tests/test_game_rating.py
```diff
@@ -250,12 +282,17 @@ async def test_on_game_end_ladder_ratings_persisted(ladder_game, players):
...
+    assert ladder_results.rating_type == RatingType.LADDER_1V1
+    assert players.hosting.id in ladder_results.ratings
+    assert players.joining.id in ladder_results.ratings
+    assert ladder_results.outcomes[players.hosting.id] is GameOutcome.VICTORY
+    assert ladder_results.outcomes[players.joining.id] is GameOutcome.DEFEAT
+
+    assert global_results.rating_type == RatingType.GLOBAL
+    assert players.hosting.id in global_results.ratings
+    assert players.joining.id not in global_results.ratings
+    assert global_results.outcomes[players.hosting.id] is GameOutcome.VICTORY
```
BlackYps, line 295: What about the defeat for the second player?

tests/unit_tests/test_game_rating.py
```diff
@@ -277,17 +314,22 @@ async def test_on_game_end_ladder_ratings_published(ladder_game, players):
...
+    assert host_global_result["rating_type"] == RatingType.GLOBAL
+    assert "new_rating_mean" in host_global_result
+    assert "new_rating_deviation" in host_global_result
+    assert host_global_result["outcome"] == GameOutcome.VICTORY.value
+
+    join_ladder_result, = results[players.joining.id]
+    assert join_ladder_result["rating_type"] == RatingType.LADDER_1V1
+    assert "new_rating_mean" in join_ladder_result
+    assert "new_rating_deviation" in join_ladder_result
+    assert join_ladder_result["outcome"] == GameOutcome.DEFEAT.value
```
BlackYps, line 332: It's also left out here and on following tests. I feel like I am missing something...
The adjustment should happen for both players, shouldn't it?

tests/unit_tests/test_game_rating.py
```diff
@@ -984,10 +1047,292 @@ async def test_single_wrong_report_still_rated_correctly(game: Game, player_fact
...
+    rating_service,
+    player_factory,
+    ratings,
+    expected_results
+):
+    team1 = set()
+    team2 = set()
+    teams = (team1, team2)
+    for player_id, ratings in ratings.items():
+        _ = player_factory(player_id=player_id)
```
BlackYps, line 1074: What does this line do?

tests/unit_tests/test_game_rating.py
```diff
@@ -984,10 +1047,292 @@ async def test_single_wrong_report_still_rated_correctly(game: Game, player_fact
...
+            PersistedResults(
+                rating_type=RatingType.GLOBAL,
+                ratings={1: pytest.approx((1766, 429), abs=1)},
+                outcomes=IGNORED
+            )
+        ]
+    )
+
+
+async def test_rating_adjustment_1v1_ladder_newbies_global_pros(
```
BlackYps, line 1128: Should add a test (or modify this one) where the global is higher than ladder, but still under 1400 displayed rating

tests/unit_tests/test_game_rating.py
```diff
@@ -984,10 +1047,292 @@ async def test_single_wrong_report_still_rated_correctly(game: Game, player_fact
...
+                    1: pytest.approx((2038, 348), abs=1),
+                    2: pytest.approx((1340, 419), abs=1),
+                },
+                outcomes=IGNORED
+            )
+        ]
+    )
+
+
+async def test_rating_adjustment_1v1_ladder_pros_global_mixed(
```
BlackYps, line 1193: I would add another test where everything is the same, but player 2 has lower global rating to show that the global adjustment of player 1 is not dependend on the other's global rating

tests/unit_tests/test_game_rating.py
```diff
@@ -984,10 +1047,292 @@ async def test_single_wrong_report_still_rated_correctly(game: Game, player_fact
...
+                ratings={
+                    1: pytest.approx((1373, 208), abs=1),
+                },
+                outcomes=IGNORED
+            )
+        ]
+    )
+
+
+async def test_rating_adjustment_1v1_ladder_pro_vs_global_pro(
```
BlackYps, line 1233: Not really sure what you want to demonstrate with this test...

tests/unit_tests/test_game_rating.py
```diff
@@ -984,10 +1047,292 @@ async def test_single_wrong_report_still_rated_correctly(game: Game, player_fact
...
+                outcomes=IGNORED
+            )
+        ]
+    )
+
+
+async def test_rating_adjustment_2v2_newbies(rating_service, player_factory):
+    IGNORED = object()
+
+    # Both players have no ratings yet
```
BlackYps, line 1269:
```suggestion
    # All players have no ratings yet
```

tests/unit_tests/test_game_rating.py
```diff
@@ -984,10 +1047,292 @@ async def test_single_wrong_report_still_rated_correctly(game: Game, player_fact
...
+    await do_test_rating_adjustment(
+        rating_service,
+        player_factory,
+        ratings={
+            1: {}, 3: {},
```
BlackYps, line 1274: Maybe it makes sense to give one of them a global rating to show how the system behaves

tests/unit_tests/test_game_rating.py
```diff
@@ -984,10 +1047,292 @@ async def test_single_wrong_report_still_rated_correctly(game: Game, player_fact
...
+                    1: pytest.approx((1688, 466), abs=1),
+                    3: pytest.approx((1688, 466), abs=1),
+                },
+                outcomes=IGNORED
+            )
+        ]
+    )
+
+
+async def test_rating_adjustment_3v3_newbies(rating_service, player_factory):
```
BlackYps, line 1302: Same here basically

server/rating_service/rating_service.py
```diff
@@ -148,192 +138,304 @@ async def _handle_rating_queue(self) -> None:
...
+    ) -> GameRatingResult:
+        """
+        Rates a game using a particular rating_type and GameRater.
+        """
+        uninitialized_ratings = {
+            # Querying the key will create the value using rating
+            # initialization, sort of like a defaultdict.
+            player_id: Rating(*player_ratings[player_id][rating_type])
+            for player_id in player_ratings.keys()
+            if rating_type not in player_ratings[player_id]
```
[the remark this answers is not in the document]
Askaholic: Yes `in` checks correspond to the `__contains__` method which is unchanged for default dicts. Only `[]` access which corresponds to the `__getitem__` method will create defaults.

```python
>>> d = defaultdict(list)
>>> 1 in d
False
>>> d[1]
[]
>>> 1 in d
True
```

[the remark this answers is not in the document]
Askaholic: Same with `PlayerRatings`

```python
>>> d = PlayerRatings({}, init=False)
>>> "global" in d
False
>>> d["global"]
Rating(mean=1500, dev=500)
>>> "global" in d
True
```

tests/integration_tests/test_game.py
```diff
@@ -562,15 +578,16 @@ async def test_ladder_game_draw_bug(lobby_server, database):
...
     # Both start at (1500, 500).
     # When rated as a draw the means should barely change.
-    assert abs(ratings["ladder1"][0] - 1500.) < 1
-    assert abs(ratings["ladder2"][0] - 1500.) < 1
+    assert abs(ratings["ladder1"]["ladder_1v1"][0] - 1500.) < 1
+    assert abs(ratings["ladder2"]["ladder_1v1"][0] - 1500.) < 1
+
+    # Global rating should also be updated
+    assert abs(ratings["ladder1"]["global"][0] - 1500.) < 1
+    assert abs(ratings["ladder2"]["global"][0] - 1500.) < 1
```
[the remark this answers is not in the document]
Askaholic: Good idea

tests/integration_tests/test_game.py
```diff
@@ -562,15 +578,16 @@ async def test_ladder_game_draw_bug(lobby_server, database):
...
         })
 
-    ratings = await get_player_ratings(
-        proto1, "ladder1", "ladder2",
-        rating_type="ladder_1v1"
-    )
+    ratings = await get_player_ratings_all(proto1, "ladder1", "ladder2")
 
     # Both start at (1500, 500).
     # When rated as a draw the means should barely change.
```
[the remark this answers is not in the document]
Askaholic: Well it's supposed to assert that the bug does not happen, so naturally the test needs to check that the game was in fact rated as a draw like it should be. This test only covers the scenario that the server can handle though. I think what really happens right now is the server gets even fewer game results than this test provides. Not sure, I haven't looked into it in a while.

tests/unit_tests/test_game_rater.py
```diff
@@ -1,90 +1,79 @@
...
+    ratings = {
+        player1_id: Rating(1500, 500),
+        player2_id: Rating(1400, 400),
+    }
+
+    return GameRatingSummary(
+        1,
+        RatingType.GLOBAL,
+        [
+            TeamRatingSummary(GameOutcome.VICTORY, set((player1_id,)), []),
```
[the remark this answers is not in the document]
Askaholic: Good question :P

tests/unit_tests/test_game_rating.py
```diff
@@ -28,41 +30,61 @@ class PersistenceError(Exception):
...
         nonlocal mock_ratings
         nonlocal mock_service
         mock_service._logger.debug(
             f"Set mock {rating_type} rating for player {player_id}: {rating}"
         )
-        mock_ratings[(player_id, rating_type)] = rating
+        mock_ratings[player_id][rating_type] = rating
 
-    def get_mock_rating(*args, **kwargs):
+    def get_mock_ratings(conn, player_ids, **kwargs):
```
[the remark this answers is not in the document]
Askaholic: It needs to match the signature of `_get_all_player_ratings` because it will be mocking that function.

tests/unit_tests/test_game_rating.py
Askaholic: The defeated player's rating adjustment would have caused the displayed rating to decrease therefore it should not be applied.

tests/unit_tests/test_game_rating.py
Askaholic: Only if the displayed rating increases.

tests/unit_tests/test_game_rating.py
Askaholic: Nothing lol. It used to be used just for the player_id but then I refactored it.

tests/unit_tests/test_game_rating.py
Askaholic: Mostly to show a situation where the global rating adjustment is going to be pretty huge. The adjustment will be rated as if a 0 rated player (1500, 500) was able to beat a 1600 rated player (1900, 75) in a 1v1. The 0 rated player's displayed rating will jump to 479 in a single game.

tests/unit_tests/test_game_rating.py
BlackYps: Can we get a comment there explaining this?

tests/unit_tests/test_game_rater.py
```diff
@@ -1,90 +1,79 @@
...
-        TeamRatingData(
-            team2_outcome, {player3_id: player3_rating, player4_id: player4_rating}
-        ),
-    ]
+    return GameRatingSummary(
+        1,
+        RatingType.GLOBAL,
+        [
+            TeamRatingSummary(GameOutcome.VICTORY, set((player1_id, player2_id)), []),
+            TeamRatingSummary(GameOutcome.DEFEAT, set((player3_id, player4_id)), [])
```
BlackYps, line 49: Maybe a question of preference, but you could use {} here too. Then it is also more consistent with the fixture above

tests/unit_tests/test_game_rating.py
```diff
@@ -984,10 +1047,426 @@ async def test_single_wrong_report_still_rated_correctly(game: Game, player_fact
...
+                rating_type=RatingType.LADDER_1V1,
+                ratings={
+                    1: pytest.approx((923, 122), abs=1),
+                    2: pytest.approx((677, 122), abs=1),
+                },
+                outcomes=IGNORED
+            ),
+            PersistedResults(
+                rating_type=RatingType.GLOBAL,
+                ratings={1: pytest.approx((1006, 75), abs=1)},
```
BlackYps, line 1184: This is actually undesireable behaviour. Because we only adjust for positive changes everybody would over time float to 1400 global even if they are stuck at 500 in the matchmakers.
We need an additional check in the rater to only adjust the rating if it is lower than the matchmaker rating.

tests/unit_tests/test_game_rating.py
BlackYps: This actually shows a bug. We want to have the player "play" with his global rating against the ladder rating of player 2 for the adjustment. So it should be the result as if two (1500, 500) players played against each other.
However the code looks like it should do exactly that, so I am not sure where the bug is.

tests/integration_tests/test_game.py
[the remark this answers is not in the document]
BlackYps: Yeah, it doesn't need to be adressed in this PR anyway

BlackYps:
Can you put further changes in separate commits? It makes it easier to keep track of the changes and in the end we squash anyway.

tests/unit_tests/test_game_rater.py
Askaholic: Yea I like that better, I just wasn't thinking of it at the time.

tests/unit_tests/test_game_rating.py
Askaholic: Ah, yea forgot that one

tests/unit_tests/test_game_rating.py
Askaholic: Oh I guess you're right. It's not bugged though because (1766, 429) is actually just what the winner of a (1500, 500) vs (1500, 500) game gets. I'll refactor that to a constant to make it more clear

tests/unit_tests/test_game_rating.py
BlackYps: .

BlackYps (commented):
Looks good, I think you just forgot this one

BlackYps:
That didn't work as expected, lol. I meant this: 
grafik

tests/unit_tests/test_game_rating.py
Askaholic: Oh yea, missed it since it was from previous review. Do we really need a comment though because it already has the line  
```python
 assert players.joining.id not in global_results.ratings
```
asserting that the rating of the second player should not change. I think there are a number of tests that all have this same behavior

BlackYps: Oh yeah, when I first looked at it I didn't notice the `not`. I guess it is ok

BlackYps (approved):

PR #854, opened 2021-10-27
Make RabbitMQ exchanges durable

### LINKED ISSUE #853: faf-lobby RabbitMQ exchange is not created as durable
But it should be.

Also it causes weird Spring errors because it doesn't have the expected properties. IF I make it durable, then the python server fails with the same error (but inverted booleans):

```
PRECONDITION_FAILED - inequivalent arg 'durable' for exchange 'faf-lobby' in vhost '/faf-core': received 'false' but current is 'true'
```

Brutus5000:
Fixes #853
As long as we do not have any session-related data,
all exchanges and queues should be durable to avoid
data loss on server restart.

### DIFF
--- a/server/message_queue_service.py
+++ b/server/message_queue_service.py
@@ -87,7 +87,7 @@ class MessageQueueService(Service):
         self._logger.debug("Connected to RabbitMQ %r", self._connection)
 
     async def declare_exchange(
-        self, exchange_name: str, exchange_type: ExchangeType = ExchangeType.TOPIC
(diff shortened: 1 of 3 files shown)


Askaholic:
That's what I mean, we need a corresponding PR on the league service side.

Brutus5000:
Yes I was aware, I meant I could sneak it in since it it's easy to do, but I wasn't sure about the other changes on the develop branch here.

So I opened a PR now and we have some time: https://github.com/FAForever/faf-league-service/pull/36

PR #856, opened 2021-11-14
Issue/#685 add game_type to game_launch message

KaukaHan:


server/lobbyconnection.py
```diff
@@ -1038,7 +1038,9 @@ async def launch_game(
             # options are. Currently, options for ladder are hardcoded into the
             # client.
             "name": game.name,
+            # Depricated: init_mode can be inferred from game_type
             "init_mode": game.init_mode.value,
+            "game_type": game.game_type.value,
```
Askaholic, line 1043: These should be strings. That’s part of the reason for switching.

tests/unit_tests/test_lobbyconnection.py
```diff
@@ -334,6 +341,7 @@ async def test_command_game_join_calls_join_game(
         "mod": "faf",
         "name": "Test Game Name",
         "init_mode": InitMode.NORMAL_LOBBY.value,
+        "game_type": GameType.CUSTOM.value,
```
Askaholic, line 344: I think it’s better to avoid using the enum in the test assertions and put the value directly. It’s a bit more work if it changes, but it makes it more clear what we expect the message to actually look like.

Askaholic (changes requested):
Also please put "Closes #856" in the PR description.

server/lobbyconnection.py
```diff
@@ -1038,7 +1038,9 @@ async def launch_game(
             # options are. Currently, options for ladder are hardcoded into the
             # client.
             "name": game.name,
+            # Depricated: init_mode can be inferred from game_type
```
Askaholic, line 1041: Also `DEPRICATED` should be all caps

server/lobbyconnection.py
KaukaHan: both of them ?

Askaholic: No just GameType. Init mode needs to stay as it is since it’s being used currently, changing it would break the client

KaukaHan: is it better to have the enum members be strings or are to/from_string methods preferred ?

Askaholic: I think making them be strings is fine. I’m honestly not sure anymore why we have the to/from methods. I don’t think there was any particular reason though, just didn’t think of doing it the other way.

PR #857, opened 2021-11-15
Issue/#685 add `game_type` to `game_launch` message

### LINKED ISSUE #685: Use game_type instead of init_mode in game_launch message

KaukaHan:
Closes #685


KaukaHan:
the unittests were successfull locally
semms like gw was causing the failure

server/games/typedefs.py
```diff
@@ -36,21 +36,10 @@ class Victory(Enum):
...
-    def from_string(value: str) -> Optional["GameType"]:
-        """
-        # Returns
-        `GameType` or `None` if the string is not valid
-        """
-        return GameType.__members__.get(value.upper())
-
-    def to_string(self) -> str:
-        return self.name.lower()
+class GameType(str, Enum):
```
Askaholic, line 39: Why does it inherent from `str`?

Askaholic:
That test sometimes fails randomly as it has to deal with weak references.

server/games/typedefs.py
KaukaHan: without it does not have the ability to compare.

Askaholic: Does one of the tests fail? Normally you compare enums using `is`. I’m sure we can fix our tests so that we don’t need to inherent from `str` here.

Askaholic:
Also check codacy, there is an unused import.

server/games/typedefs.py
KaukaHan: FAILED tests/integration_tests/test_matchmaker.py::test_game_launch_message_game_options - concurrent.futures._base.TimeoutError
FAILED tests/integration_tests/test_matchmaker.py::test_game_matchmaking_timeout - concurrent.futures._base.TimeoutError
FAILED tests/integration_tests/test_matchmaker.py::test_anti_map_repetition - concurrent.futures._base.TimeoutError
FAILED tests/integration_tests/test_teammatchmaker.py::test_game_ratings - concurrent.futures._base.TimeoutError
FAILED tests/integration_tests/test_teammatchmaker.py::test_ratings_initialized_based_on_global_persisted - concurrent.futures._base.TimeoutError

these fail without the str inheritage
in contrast to: 
FAILED tests/integration_tests/test_matchmaker.py::test_game_launch_message_game_options - concurrent.futures._base.TimeoutError
FAILED tests/integration_tests/test_matchmaker.py::test_game_matchmaking_disconnect - concurrent.futures._base.TimeoutError
with the str inheritage

KaukaHan: i have the idea with str from here:
https://stackoverflow.com/questions/58608361/string-based-enum-in-python

Askaholic: Can you figure out what exception is causing those tests to fail? I don’t think we should need to compare the enum to strings anywhere, that seems like an issue with the code to me rather than an issue with the enum definition.

KaukaHan: yeah ill hang on to it.

KaukaHan: should work now, seems like a rare error, not connected to my changes.

Askaholic:
Oh you actually found a different unused import from the one codacy was complaining about :D

Askaholic (approved):

PR #862, opened 2021-12-01
Issue/#288 Explicitly set game start time

### LINKED ISSUE #288: Explicitly set startTime
The server needs to set `game_stats.startTime` explicitly. See FAForever/db#83

KaukaHan:
closes #288

### DIFF
--- a/server/games/game.py
+++ b/server/games/game.py
@@ -419,7 +419,7 @@ class Game():
         if [conn for conn in self.connections if not conn.finished_sim]:
             return
         self.ended = True
-        self.endTime = datetime.now(datetime.timezone.utc)
+        self.endTime = datetime.datetime.now(datetime.timezone.utc)
         async with self._db.acquire() as conn:
             await conn.execute(
                 game_stats.update().where(
@@ -714,7 +714,7 @@ class Game():
     async def on_game_launched(self):
         for player in self.players:
             player.state = PlayerState.PLAYING
-        self.startTime = datetime.now(datetime.timezone.utc)
+        self.startTime = datetime.datetime.now(datetime.timezone.utc)
         await self.update_game_stats()
         await self.update_game_player_stats()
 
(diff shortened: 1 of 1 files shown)


server/games/game.py
```diff
@@ -758,6 +758,7 @@ async def update_game_stats(self):
                     mapId=self.map_id,
                     gameName=self.name,
                     validity=self.validity.value,
```
Askaholic, line 760: I think the point of the issue is that the collection of the start time and the database insert should be separated. As in it should be possible to move the database insert to an earlier or later time and the start time should still be recorded correctly.

Askaholic:
You should get rid of the unrelated commits. Let me know if you need some help with the git commands to do this

KaukaHan:
> You should get rid of the unrelated commits. Let me know if you need some help with the git commands to do this
that would be helpfull

server/games/game.py
```diff
@@ -80,6 +81,8 @@ def __init__(
         self.game_service = game_service
         self._player_options: Dict[int, Dict[str, Any]] = defaultdict(dict)
         self.launched_at = None
+        self.startTime = None
+        self.endTime = None
```
Askaholic, line 85: Variable names should be snake_case

Askaholic: I know the table columns aren’t, but that’s just cuz the database schema is a bit of a mess

Askaholic (commented):
Looks good. Now we just need some unit tests to check that the variables are set correctly and maybe one to check that the right value is written to the database.

server/games/game.py
KaukaHan: do i have to use hardcoded timestemps for the tests ?

Askaholic: You should probably mock the calls to `datetime.now` to return a specific value and then assert that the value shows up in the database.

server/games/game.py
```diff
@@ -416,12 +419,13 @@ async def check_sim_end(self):
         if [conn for conn in self.connections if not conn.finished_sim]:
             return
         self.ended = True
+        self.endTime = datetime.datetime.now(datetime.timezone.utc)
```
Askaholic, line 422: I'd probably import like:
```python3
from datetime import datetime, timezone
```
Just so these calls are a little cleaner.

Askaholic:
No work happening on this

PR #863, opened 2021-12-02
add player_ids field to game_info message

KaukaHan:
closes #812

PR #864, opened 2021-12-02
issue/#812 add player ids to game_info

### LINKED ISSUE #812: Lobby list json additional info.
May be add player id and map version id fields to lobby json?
If player change name he cant be found until re login.
Find map by name may cost a lot for server sometimes.

KaukaHan:
Replacement for #863 

closes #812


KaukaHan:
doesn't this make the `teams` field depricated ?

Askaholic:
Yea it’s a bit of a tricky one to do with the versioning, because really what we want is to keep `teams` but make it use player ids instead of names. So I think what we have to do is call it `teams_ids` for now and then make a comment to rename it to `Teams` and remove the old one when we go to 2.0. 

How does that sound @Sheikah45?

Sheikah45:
Yes that is what has to be done. There is no way to include team and id info in the current team structure without breaking all the clients.

Brutus5000:
Classical backwards-compatibility issues with flat json arrays. Seems efficient first to just store strings or ids, until you need more.

Askaholic:
Maybe we should take the opportunity to redesign the message a little bit to make it more backwards-friendly in case of changes. We should probably do that on the issue though.

Kazbek:
Any news here?

tests/unit_tests/test_game.py
```diff
@@ -630,6 +630,16 @@ async def test_to_dict(game, player_factory):
         "rating_min": game.displayed_rating_range.lo,
         "rating_max": game.displayed_rating_range.hi,
         "enforce_rating_range": game.enforce_rating_range,
+        "teams_ids": [
```
Askaholic, line 633: Agh, I hate that this test still doesn't use hardcoded values. It's basically a reimplementation of the to_dict function.

Askaholic (approved):
Looks good to me. @Sheikah45 any comments?

Sheikah45:
Looks good to me although teams_ids sounds weird in my head 

> Yea it’s a bit of a tricky one to do with the versioning, because really what we want is to keep `teams` but make it use player ids instead of names. So I think what we have to do is call it `teams_ids` for now and then make a comment to rename it to `Teams` and remove the old one when we go to 2.0.
> 
> How does that sound @Sheikah45?

Also just as a note as stated this will never really be possible as it would require first removing recognition of the old teams from all clients first. Probably easier to just leave it as teams_ids going forward

KaukaHan:
'player_ids_per_team' is a more descriptive alternative i can think of.

tests/unit_tests/test_game.py
KaukaHan: does this only test one (game, player_factory) or multiple combinations of the two ? in the later case how woud you hardcode that ?

Issue #866, opened 2021-12-14
Matchmaker game timeout sometimes cancels the wrong game


Askaholic:
Here, the first game `15904960` failed to start at around 17:59:29 however, no match cancelled message was sent. After about 10 minutes the handler in LadderService finally saw that the game didn't launch and sent the match cancelled message at 18:10:46 which ended up cancelling a different game that had launched in the mean time.

I think we should first of all include a game id in the match cancelled message so that the client can check if the message that's arriving is actually meant for the current game. And second of all figure out why it took 10 minutes for the game to time out when the maximum timeout should only have been `2.66` minutes.

```
INFO     Dec 13  17:58:38 LadderGame.15904960            Added game connection GameConnection(Player(TreefolkNoob, 206383, Rating(mean=2128.88, dev=85.6139), Rating(mean=2159.26, dev=65.3348)), Game(15904960, TreefolkNoob, maps/turtle_rocks.v0002.zip))
INFO     Dec 13  17:58:44 LadderGame.15904960            Added game connection GameConnection(Player(Itness, 282306, Rating(mean=1518.76, dev=81.5663), Rating(mean=1953.86, dev=366.154)), Game(15904960, TreefolkNoob, maps/turtle_rocks.v0002.zip))
INFO     Dec 13  17:58:45 LadderGame.15904960            Added game connection GameConnection(Player(Wpierdolo, 282284, Rating(mean=1547.42, dev=80.9831), Rating(mean=1155.62, dev=66.3089)), Game(15904960, TreefolkNoob, maps/turtle_rocks.v0002.zip))
INFO     Dec 13  17:59:29 LadderGame.15904960            Removed game connection GameConnection(Player(TreefolkNoob, 206383, Rating(mean=2128.88, dev=85.6139), Rating(mean=2159.26, dev=65.3348)), Game(15904960, TreefolkNoob, maps/turtle_rocks.v0002.zip))
INFO     Dec 13  17:59:29 LadderGame.15904960            Game cancelled pre launch
INFO     Dec 13  17:59:38 LadderGame.15904960            Removed game connection GameConnection(Player(Wpierdolo, 282284, Rating(mean=1547.42, dev=80.9831), Rating(mean=1155.62, dev=66.3089)), Game(15904960, TreefolkNoob, maps/turtle_rocks.v0002.zip))
[...]

PR #867, opened 2021-12-14
Issue/#866 matchmaker timeout

Askaholic:
- Add `game_id` to `match_cancelled` message so the client can choose to ignore it if the message doesn't apply to the current game
- Prevent `start_game` from stalling on hung connections. The timeouts should always be followed now.

@Sheikah any preference on what to call the `game_id` field? I think `game_id` makes sense but in `game_info` it's called `uid`.

Mitigates some of #866 but doesn't fix it completely.

### DIFF
--- a/server/lobbyconnection.py
+++ b/server/lobbyconnection.py
@@ -41,6 +41,7 @@ from .games import (
     CoopGame,
     CustomGame,
     FeaturedModType,
+    Game,
     GameConnectionState,
     GameState,
     InitMode,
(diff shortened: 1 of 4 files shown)


Sheikah45:
Game_id works

server/lobbyconnection.py
```diff
@@ -86,11 +87,11 @@ def __init__(
...
         self.oauth_service = oauth_service
         self._authenticated = False
-        self.player = None  # type: Player
-        self.game_connection = None  # type: GameConnection
-        self.peer_address = None  # type: Optional[Address]
+        self.player: Optional[Player] = None
+        self.game_connection: Optional[GameConnection] = None
+        self.peer_address: Optional[Address] = None
         self.session = int(random.randrange(0, 4294967295))
-        self.protocol: Protocol = None
```
norraxx, line 93: I like which way typing ia going! 👍

tests/integration_tests/test_matchmaker.py
```diff
@@ -303,7 +303,7 @@ async def test_game_matchmaking_disconnect(lobby_server):
 
     msg = await read_until_command(proto2, "match_cancelled", timeout=120)
 
-    assert msg == {"command": "match_cancelled"}
+    assert msg == {"command": "match_cancelled", "game_id": 41956}
```
BlackYps, line 306: Where is this id set?

BlackYps (approved):

tests/integration_tests/test_matchmaker.py
Askaholic: It should be 1 higher than the highest game id in the test data.

PR #872, opened 2021-12-24
Fix "Can't xxxx game while in state xxxx" messages

Askaholic:
The messages happen after you click to join or host a game and the client sends `game_host` but then never actually starts FA after getting the `game_launch` message. While I do consider this a client bug, this behavior is more correct and is needed for the matchmaker timeout stuff.

### DIFF
--- a/tests/integration_tests/test_game.py
+++ b/tests/integration_tests/test_game.py
@@ -420,7 +420,7 @@ async def test_game_ended_broadcasts_rating_update(lobby_server, channel):
 
 
 @fast_forward(30)
-async def test_double_join_message(lobby_server):
+async def test_double_host_message(lobby_server):
     _, _, proto = await connect_and_sign_in(
         ("test", "test_password"), lobby_server
     )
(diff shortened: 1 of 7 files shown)


BlackYps:
Do we know some scenarios when the game sends an idle state?

BlackYps:
Do I understand this correctly that there seems to be a client bug that it doesn't open fa every time whent it gets a game_launch message? We can't fix that here but we get rid of the leftover `joining` states by only setting them when the game has opened, right?

tests/integration_tests/test_teammatchmaker.py
```diff
@@ -330,7 +335,12 @@ async def test_game_matchmaking_multiqueue_timeout(lobby_server):
     )
     assert msg["state"] == "stop"
 
-    # Don't send any GPGNet messages so the match times out
+    await client_response(protos[0])
+
+    await asyncio.gather(*[
+        idle_response(proto) for proto in protos[1:2]
+    ])
```
BlackYps, line 342: isn't this identical to `await idle_response(protos[1])`?

BlackYps (changes requested):

tests/integration_tests/test_teammatchmaker.py
Askaholic: Oh good catch. I think it should be `protos[1:3]` but I'll double check

tests/integration_tests/test_game.py
```diff
@@ -409,6 +419,26 @@ async def test_game_ended_broadcasts_rating_update(lobby_server, channel):
         await asyncio.wait_for(mq_proto_all.read_message(), timeout=10)
 
 
+@fast_forward(30)
+async def test_double_join_message(lobby_server):
```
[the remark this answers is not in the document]
Askaholic: I guess this would be hosting twice, but I should make one for joining twice as well although it should trigger exactly the same code.

tests/integration_tests/test_game.py
```diff
@@ -409,6 +419,55 @@ async def test_game_ended_broadcasts_rating_update(lobby_server, channel):
...
+    })
+    await read_until_command(proto, "game_launch", timeout=10)
+
+
+@fast_forward(30)
+async def test_double_join_message(lobby_server):
+    _, _, host_proto = await connect_and_sign_in(
+        ("test", "test_password"), lobby_server
+    )
+    guest_id, _, guest_proto = await connect_and_sign_in(
```
BlackYps, line 447: Get rid of the unused guest_id then it looks good

BlackYps (approved):

tests/integration_tests/test_game.py
Askaholic: Aaargh. Copy/paste got me

Sheikah45:
> Do I understand this correctly that there seems to be a client bug that it doesn't open fa every time whent it gets a game_launch message? We can't fix that here but we get rid of the leftover `joining` states by only setting them when the game has opened, right?

I had an instance of this happen to me today but it seemed to have been caused by the server or something closing the tcp connection. Unforuntately I did not have debug logging on but did see the connection was closed

PR #879, opened 2022-01-07
Add more prometheus metrics

### LINKED ISSUE #877: Add more grafana logs
We need some more metrics to improve the matchmaker
- number or percentage of cancelled matches
- some better quality metrics, right now it just uses trueskill quality. The matchmaker uses a more complicated system to find maches. E.g. I would like to be able to see if the low quality matches were made because of high time bonuses.
- size of search parties. This is relevant for being able to properly test new matchmaker configurations. Right now we have no idea how many people team up to search.
- live games by leaderboard

BlackYps:
Closes #877 

Not sure if I put the MatchLaunch enum in the best file.
Can Prometheus deal with the changed metric names and labels?
My idea with the game quality metrics is to display dots in a timeline graph instead of using the heatmap. This should enable us to compare the different queues in one graph.

### DIFF
--- a/server/matchmaker/matchmaker_queue.py
+++ b/server/matchmaker/matchmaker_queue.py
@@ -34,7 +34,7 @@ class MatchmakerSearchTimer:
         else:
             status = "errored"
 
-        metric = metrics.matchmaker_searches.labels(self.queue_name, status)
+        metric = metrics.matchmaker_search_duration.labels(self.queue_name, status)
         metric.observe(total_time)
 
 
@@ -184,12 +184,17 @@ class MatchmakerQueue:
 
         self._register_unmatched_searches(unmatched_searches)
 
-        number_of_matches = len(matches)
-        metrics.matches.labels(self.name).set(number_of_matches)
-
         for search1, search2 in matches:
-            # TODO: Move this into algorithm, then don't need to recalculate
-            # quality_with? Probably not a major bottleneck though.
+            self._report_party_sizes(search1)
+            self._report_party_sizes(search2)
+
+            rating_imbalance = abs(search1.cumulative_rating - search2.cumulative_rating)
+            metrics.match_rating_imbalance.labels(self.name).set(rating_imbalance)
+
+            ratings = search1.displayed_ratings + search2.displayed_ratings
+            rating_variety = max(ratings) - min(ratings)
+            metrics.match_rating_variety.labels(self.name).set(rating_variety)
+
             metrics.match_quality.labels(self.name).observe(
                 search1.quality_with(search2)
             )
@@ -197,6 +202,12 @@ class MatchmakerQueue:
                 self.on_match_found(search1, search2, self)
             except Exception:
                 self._logger.exception("Match callback raised an exception!")
+
+    def _report_party_sizes(self, team):
+        for search in team.get_original_searches():
+            metrics.matched_matchmaker_searches.labels(
+                self.name, len(search.players)
+            ).inc()
 
     def _register_unmatched_searches(
         self,
--- a/server/metrics.py
+++ b/server/metrics.py
@@ -4,18 +4,48 @@ Prometheus metric definitions
 
 from prometheus_client import Counter, Gauge, Histogram, Info
 
+
+class MatchLaunch:
+    SUCCESSFUL = "successful"
+    TIMED_OUT = "timed out"
+    ABORTED_BY_PLAYER = "aborted by player"
+    ERRORED = "errored"
+
+
 info = Info("build", "Information collected on server start")
 
 # ==========
 # Matchmaker
 # ==========
-matches = Gauge("server_matchmaker_queue_matches", "Number of matches made", ["queue"])
+matches = Counter(
+    "server_matchmaker_queue_matches_total",
+    "Number of matches made",
+    ["queue", "status"]
+)
+
+matched_matchmaker_searches = Counter(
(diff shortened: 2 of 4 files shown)


Brutus5000:
> Can Prometheus deal with the changed metric names and labels?

Prometheus doesn't care about changed metric names and labels. It will just create them if they didn't exist before. In Grafana you will just lose history. If you can live with that it's fine. Prometheus will keep data only for 12 weeks anyway.

BlackYps:
The metric history we would lose doesn't seem really important to me

server/matchmaker/matchmaker_queue.py
```diff
@@ -179,17 +179,24 @@ async def find_matches(self) -> None:
         for match in proposed_matches:
             if self.match(match[0], match[1]):
                 matches.append(match)
+                for team in match:
+                    for search in team.get_original_searches():
+                        metrics.matched_matchmaker_searches.labels(
+                            self.name, len(search.players)
+                        ).inc()
```
Askaholic, line 186: Hmm that's maybe too many levels of indentation. Maybe extract to a method and put it down in the other loop where you're reporting the quality metrics.

server/metrics.py
```diff
@@ -113,6 +135,12 @@
     ["game_mode", "game_state"],
 )
 
+rated_games = Counter(
+    "server_game_rated_games_total",
+    "Number of rated games",
+    ["leaderboard"]
+)
+
```
Askaholic, line 143: I wonder if this would be better as a label on the `active_games` metric

Askaholic (commented):
We also should have a grafana instance on the test server that you can use to play around with the dashboards before merging, to make sure the metrics types work for the sort of graphs you want to have.

server/games/typedefs.py
```diff
@@ -14,6 +14,13 @@ class GameState(Enum):
     ENDED = 3
 
 
+@unique
+class MatchLaunch(Enum):
+    SUCCESSFUL = "successful"
+    TIMED_OUT = "timed out"
```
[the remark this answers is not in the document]
BlackYps: I checked the documentation and it claims that any unicode character is allowed here

BlackYps:
That is a good idea, but the test server doesn't have any activity, so what data would there be to show?

server/games/typedefs.py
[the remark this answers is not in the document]
Askaholic: Ok. Maybe it makes more sense to put this enum in the `metrics` module rather than the `game.typedefs`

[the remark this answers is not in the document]
BlackYps: simply at the start of the metrics.py?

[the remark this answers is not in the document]
Askaholic: Sure

server/metrics.py
BlackYps: There is only really the faf gamemode that has more than one leaderboard. The documentation advises against having too many possible combinations of lables as that is bad for performance. I would want to have these two in different graphs anyway, so I think it is better to keep them separate.

server/ladder_service.py
```diff
@@ -495,19 +497,23 @@ def make_game_options(player: Player) -> GameLaunchOptions:
                         )
             await game.wait_launched(60 + 10 * len(all_guests))
             self._logger.debug("Ladder game launched successfully %s", game)
+            metrics.matches.labels(queue.name, MatchLaunch.SUCCESSFUL).inc()
```
Askaholic, line 500: Does this work actually? The types are a bit weird here since `MatchLaunched.SUCCESSFUL` is an Enum not a string. If you did `MatchLaunch.SUCCESSFUL.value` it would definitely work, but maybe the library knows how to do that automatically?

server/matchmaker/matchmaker_queue.py
```diff
@@ -184,12 +184,17 @@ async def find_matches(self) -> None:
...
-        metrics.matches.labels(self.name).set(number_of_matches)
-
         for search1, search2 in matches:
-            # TODO: Move this into algorithm, then don't need to recalculate
-            # quality_with? Probably not a major bottleneck though.
+            self._report_party_sizes(search1)
+            self._report_party_sizes(search2)
+
+            rating_imbalance = abs(search1.cumulative_rating - search2.cumulative_rating)
+            metrics.match_rating_imbalance.labels(self.name).set(rating_imbalance)
```
Askaholic, line 192: Not sure if this will be that useful. Setting a Gauge in a loop like this will make it so you basically only ever see the value of the last one set. Maybe a `Histogram` would be more useful? Same for `match_rating_variety` below.

server/metrics.py
```diff
@@ -1,21 +1,53 @@
 """
 Prometheus metric definitions
 """
+from enum import Enum, unique
 
 from prometheus_client import Counter, Gauge, Histogram, Info
 
+
+@unique
+class MatchLaunch(Enum):
```
Askaholic, line 10: The issue with Enum.value would also be solved if you just not make this an Enum. Then it will just be a class for holding some string constants like the RatingType class.

server/matchmaker/matchmaker_queue.py
BlackYps: Each metric generates a time series. https://prometheus.io/docs/concepts/data_model/
So we won't miss any data. The loop will make it so the timestamps are really close for batches of data, but this should not be a problem.

Askaholic (approved):

Issue #873, opened 2021-12-25
Cancel matches immediately if one player closes FA


Askaholic:
We can detect when FA is closed by the 'GameState: Ended' command. If one of the players in an automatch does this we can immediately cancel the match without waiting for the timeout as we know it will never be able to recover.

PR #882, opened 2022-01-07
Cancel matches if someone closes the game

BlackYps:
#873 
I's not unlikely that the cancellation doesn't happen at the ideal place

### DIFF
--- a/server/games/game.py
+++ b/server/games/game.py
@@ -47,16 +47,6 @@ class GameError(Exception):
     pass
 
 
-class GameClosedError(Exception):
-    """
-    The game has been closed during the setup phase
-    """
-
-    def __init__(self, player: Player, *args, **kwargs):
-        super().__init__(*args, **kwargs)
-        self.player = player
-
-
 class Game:
     """
     Object that lasts for the lifetime of a game on FAF.
@@ -128,7 +118,6 @@ class Game:
         }
         self.mods = {}
         self._hosted_event = asyncio.Event()
-        self._launch_future = asyncio.Future()
 
         self._logger.debug("%s created", self)
         asyncio.get_event_loop().create_task(self.timeout_game(setup_timeout))
@@ -276,20 +265,8 @@ class Game:
 
         return list(teams.values()) + ffa_players
 
-    async def wait_hosted(self, timeout: float):
-        return await asyncio.wait_for(
-            self._hosted_event.wait(),
-            timeout=timeout
-        )
-
     def set_hosted(self):
         self._hosted_event.set()
-
-    async def wait_launched(self, timeout: float):
-        return await asyncio.wait_for(
-            self._launch_future,
-            timeout=timeout
-        )
 
     async def add_result(
         self,
@@ -400,14 +377,12 @@ class Game:
         if self.state is GameState.LOBBY and player.id in self._player_options:
             del self._player_options[player.id]
 
-        if not self._launch_future.done() and (
-                self.state is GameState.INITIALIZING or self.state is GameState.LOBBY
-        ):
-            self._launch_future.set_exception(GameClosedError(player))
-
-        await self.check_sim_end()
-
         self._logger.info("Removed game connection %s", game_connection)
+
+        await self.handle_game_end(player)
+
+    async def handle_game_end(self, player):
+        await self.check_sim_end()
 
         host_left_lobby = (
             player == self.host and self.state is not GameState.LIVE
@@ -718,7 +693,6 @@ class Game:
         await self.on_game_launched()
         await self.validate_game_settings()
 
-        self._launch_future.set_result(None)
         self._logger.info("Game launched")
 
     async def on_game_launched(self):
--- a/server/games/game.py
+++ b/server/games/game.py
@@ -47,6 +47,10 @@ class GameError(Exception):
     pass
 
 
+class GameClosedError(Exception):
+    pass
+
+
 class Game():
     """
     Object that lasts for the lifetime of a game on FAF.
@@ -390,8 +394,8 @@ class Game():
         if self.state is GameState.LOBBY and player.id in self._player_options:
             del self._player_options[player.id]
 
-        if self.state is GameState.INITIALIZING:
-            self._launch_future.cancel()
+        if self.state is GameState.INITIALIZING or self.state is GameState.LOBBY:
+            self._launch_future.set_exception(GameClosedError(player))
 
         await self.check_sim_end()
 
--- a/server/games/game.py
+++ b/server/games/game.py
@@ -48,10 +48,16 @@ class GameError(Exception):
 
 
 class GameClosedError(Exception):
-    pass
+    """
+    The game has been closed during the setup phase
+    """
+
+    def __init__(self, player: Player, *args, **kwargs):
+        super().__init__(*args, **kwargs)
+        self.player = player
 
 
-class Game():
+class Game:
     """
     Object that lasts for the lifetime of a game on FAF.
     """
@@ -394,7 +400,9 @@ class Game():
         if self.state is GameState.LOBBY and player.id in self._player_options:
             del self._player_options[player.id]
 
-        if self.state is GameState.INITIALIZING or self.state is GameState.LOBBY:
+        if not self._launch_future.done() and (
+                self.state is GameState.INITIALIZING or self.state is GameState.LOBBY
+        ):
             self._launch_future.set_exception(GameClosedError(player))
 
         await self.check_sim_end()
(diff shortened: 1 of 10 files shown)


server/ladder_service.py
```diff
@@ -500,6 +500,11 @@ def make_game_options(player: Player) -> GameLaunchOptions:
                     "Ladder game failed to start! %s setup timed out",
                     game
                 )
+            elif isinstance(e, asyncio.CancelledError):
```
Askaholic, line 503: Hmm, I think using set_exception with a custom error would be better because cancelled error could be raised for many reasons.

server/games/game.py
```diff
@@ -390,6 +390,9 @@ async def remove_game_connection(self, game_connection):
         if self.state is GameState.LOBBY and player.id in self._player_options:
             del self._player_options[player.id]
 
+        if self.state is GameState.INITIALIZING:
```
Askaholic, line 393: Yea this might work.

tests/unit_tests/test_game.py
```diff
@@ -398,6 +399,9 @@ async def test_remove_game_connection(
     game.add_game_connection(mock_game_connection)
     await game.remove_game_connection(mock_game_connection)
     assert players.hosting not in game.players
+    e = game._launch_future.exception()
+    assert type(e) is GameClosedError
+    assert e.args == mock_game_connection.player
```
BlackYps, line 404: This fails for some reason

Askaholic: args is a tuple

server/ladder_service.py
```diff
@@ -500,6 +501,11 @@ def make_game_options(player: Player) -> GameLaunchOptions:
                     "Ladder game failed to start! %s setup timed out",
                     game
                 )
+            elif isinstance(e, GameClosedError):
+                self._logger.info(
+                    "Ladder game %s failed to start! Player %s closed his game instance",
```
Askaholic, line 506:
```suggestion
                    "Ladder game %s failed to start! Player %s closed their game instance",
```

server/games/game.py
```diff
@@ -47,6 +47,10 @@ class GameError(Exception):
     pass
 
 
+class GameClosedError(Exception):
```
Askaholic, line 50: Probably define a constructor that takes 1 arg. That will be less confusing than using the default constructor plus you can make use of type annotations that way

tests/unit_tests/test_ladder_service.py
```diff
@@ -256,6 +256,37 @@ async def wait_forever(*args, **kwargs):
     assert p1.lobby_connection.write_launch_game.called
 
 
+async def test_start_game_game_closed(
+        ladder_service: LadderService,
```
Askaholic, line 260: Indented too much

Askaholic (changes requested):

BlackYps:
test_game_matchmaking_close_fa_and_requeue seems to be flaky

server/games/game.py
```diff
@@ -390,6 +400,11 @@ async def remove_game_connection(self, game_connection):
         if self.state is GameState.LOBBY and player.id in self._player_options:
             del self._player_options[player.id]
 
+        if not self._launch_future.done() and (
+                self.state is GameState.INITIALIZING or self.state is GameState.LOBBY
```
Askaholic, line 404: Extra indent

Askaholic: Could also do `self.state in (GameState.INITIALIZING, GameState.LOBBY)`. Might make it cleaner?

server/ladder_service.py
```diff
@@ -500,6 +501,11 @@ def make_game_options(player: Player) -> GameLaunchOptions:
                     "Ladder game failed to start! %s setup timed out",
                     game
                 )
+            elif isinstance(e, GameClosedError):
+                self._logger.info(
+                    "Ladder game %s failed to start! Player %s closed their game instance",
+                    game, e.player
```
Askaholic, line 507: When I get around to finishing the matchmaker ban for leaving game, I will use this to register a violation

server/games/game.py
Askaholic: I also wonder if it might be better to move all of the `wait_*` and related functionality to the `LadderGame` class. Since this doesn't really make sense for non-matchmaker games.

Askaholic (changes requested):

server/games/game.py
BlackYps: you are probably right. The game class is quite big right now

server/games/ladder_game.py
```diff
@@ -1,23 +1,61 @@
...
 
 logger = logging.getLogger(__name__)
 
 
+class GameClosedError(Exception):
+    """
+    The game has been closed during the setup phase
+    """
+
+    def __init__(self, player: Player, *args, **kwargs):
```
Askaholic, line 20: Why does it even have `args` and `kwargs` still?

Askaholic (commented):
Looks pretty good, we should test it out on the test server

server/games/ladder_game.py
BlackYps: because in `BaseException` it looks like this:
```python
    def __init__(self, *args, **kwargs): # real signature unknown
        pass
```
I'm not really sure how to interpret that comment. I just assumed that I would need the two arguments as well.

BlackYps:
Do you want to finish your temp ban logic first before we test or should we test that separately?

server/games/ladder_game.py
Askaholic: Just remove them. It’s that way in base exception because it doesn’t know what specific function signatures the subclasses will have. But for a custom error you should just define the parameters you need. In this case we just need 1 for player and that’s it.

Askaholic:
No let’s do them separately. It might still be a while until I get around to finishing mine.

server/games/ladder_game.py
```diff
@@ -1,23 +1,61 @@
...
     init_mode = InitMode.AUTO_LOBBY
     game_type = GameType.MATCHMAKER
 
+    def __init__(self, id_, *args, **kwargs):
+        super().__init__(id_, *args, **kwargs)
+        self._launch_future = asyncio.Future()
+
+    async def wait_hosted(self, timeout: float):
+        return await asyncio.wait_for(
+            self._hosted_event.wait(),
```
Askaholic, line 37: Another thing I just thought of. Maybe we want to apply the same cancellation logic for the host as well? Like if the host opens the game I to the Idle state but then closes it again before getting to the Lobby state.

BlackYps: yes, good idea

BlackYps:
Do you know why this small change in the last commit makes the two unit tests fail? I have no idea why `check_sim_end()` doesn't get called properly. It is literally the first line in the function I call.

Askaholic:
It’s just because the game object is a Mock, and you changed the name of the function that is being called. You just need to update the tests as well to check for a call to the new function.

BlackYps:
Oh I see, so the function calls of the game object are just mocks. Yeah, that makes sense.

server/games/ladder_game.py
```diff
@@ -1,23 +1,66 @@
...
+            timeout=timeout
+        )
+
+    async def launch(self):
+        await super().launch()
+        self._launch_future.set_result(None)
+
+    async def check_game_finish(self, player):
+        if not self._hosted_future.done() and (
+            self.state in (GameState.INITIALIZING, GameState.LOBBY)
```
Askaholic, line 53: Shouldn't we be checking that `player == self.host` here?

BlackYps: as long as the hosted future is not done the guests won't start connecting. So there is no way they could trigger this

BlackYps:
Looking at the code I just noticed the TODO about the match_cancelled message again. Sheikah said that the client supports it now. So we could change it.

server/games/ladder_game.py
Askaholic: Then maybe add `assert self.host == player` to document this and prevent refactors from accidentally breaking this behavior.

Askaholic:
Yea, it's merged on the `v2-develop` branch already.

server/games/ladder_game.py
```diff
@@ -1,23 +1,67 @@
...
 logger = logging.getLogger(__name__)
 
 
+class GameClosedError(Exception):
+    """
+    The game has been closed during the setup phase
+    """
+
+    def __init__(self, player: Player):
+        super().__init__()
```
Askaholic, line 21: You don't need to call super here as it will do nothing

Askaholic (approved):
Lets test it on the test server

PR #885, opened 2022-01-17
Issue/#870 stop players from joining the queue after failing to connect to games

Askaholic:
This PR adds functionality to try and figure out which players are causing matchmaker games to fail by not connecting, and then warning them, and if a player causes several games to fail within a short time, then temporarily ban them (put them in 'time-out') from the matchmaker. There are a few caveats:

- At this time we have no way of knowing what the connection status is of all (n choose 2) connections, we only know who is connected to the host and who isn't. So it's possible for two guests to both be connected to the host but not be connected to eachother and for the game to fail. In this case, we don't know who to punish so we don't punish anyone.
- Even if the game launches, some people might have a spotty connection or a slow CPU that causes severe lag. We currently have no way to detect this (although it might be possible with the `Bottleneck` GPGNet command)
- The violation tracking is completely reset after 1 hour of not playing, or not failing matches. We might want to make it step down over a period of time rather than resetting all at once.

### DIFF
--- a/server/gameconnection.py
+++ b/server/gameconnection.py
@@ -517,6 +517,13 @@ class GameConnection(GpgNetServerProtocol):
         """
         pass
 
+    async def handle_launch_status(self, status: str):
+        """
+        Currently is sent with status `Rejected` if a matchmaker game failed
+        to start due to players using differing game settings.
+        """
+        pass
+
     async def handle_bottleneck(self, *args: list[Any]):
         """
         Not sure what this command means. This is currently unused but
@@ -612,25 +619,26 @@ class GameConnection(GpgNetServerProtocol):
 
 
 COMMAND_HANDLERS = {
-    "Desync":               GameConnection.handle_desync,
-    "GameState":            GameConnection.handle_game_state,
-    "GameOption":           GameConnection.handle_game_option,
-    "GameMods":             GameConnection.handle_game_mods,
-    "PlayerOption":         GameConnection.handle_player_option,
     "AIOption":             GameConnection.handle_ai_option,
-    "ClearSlot":            GameConnection.handle_clear_slot,
-    "GameResult":           GameConnection.handle_game_result,
-    "OperationComplete":    GameConnection.handle_operation_complete,
-    "JsonStats":            GameConnection.handle_json_stats,
-    "EnforceRating":        GameConnection.handle_enforce_rating,
-    "TeamkillReport":       GameConnection.handle_teamkill_report,
-    "TeamkillHappened":     GameConnection.handle_teamkill_happened,
-    "GameEnded":            GameConnection.handle_game_ended,
-    "Rehost":               GameConnection.handle_rehost,
     "Bottleneck":           GameConnection.handle_bottleneck,
     "BottleneckCleared":    GameConnection.handle_bottleneck_cleared,
-    "Disconnected":         GameConnection.handle_disconnected,
-    "IceMsg":               GameConnection.handle_ice_message,
     "Chat":                 GameConnection.handle_chat,
-    "GameFull":             GameConnection.handle_game_full
+    "ClearSlot":            GameConnection.handle_clear_slot,
+    "Desync":               GameConnection.handle_desync,
+    "Disconnected":         GameConnection.handle_disconnected,
+    "EnforceRating":        GameConnection.handle_enforce_rating,
+    "GameEnded":            GameConnection.handle_game_ended,
+    "GameFull":             GameConnection.handle_game_full,
+    "GameMods":             GameConnection.handle_game_mods,
+    "GameOption":           GameConnection.handle_game_option,
+    "GameResult":           GameConnection.handle_game_result,
+    "GameState":            GameConnection.handle_game_state,
+    "IceMsg":               GameConnection.handle_ice_message,
+    "JsonStats":            GameConnection.handle_json_stats,
+    "LaunchStatus":         GameConnection.handle_launch_status,
+    "OperationComplete":    GameConnection.handle_operation_complete,
+    "PlayerOption":         GameConnection.handle_player_option,
+    "Rehost":               GameConnection.handle_rehost,
+    "TeamkillHappened":     GameConnection.handle_teamkill_happened,
+    "TeamkillReport":       GameConnection.handle_teamkill_report,
 }
--- a/server/ladder_service/__init__.py
+++ b/server/ladder_service/__init__.py
@@ -1,0 +1,7 @@
+from .game_name import game_name
+from .ladder_service import LadderService
+
+__all__ = (
+    "LadderService",
+    "game_name",
+)
--- a/server/ladder_service/ladder_service.py
+++ b/server/ladder_service/ladder_service.py
@@ -531,6 +531,7 @@ class LadderService(Service):
                 )
                 if game:
                     connected_players = game.get_connected_players()
+                    print(f"Connected players {connected_players}")
                     non_connected_players.extend(
                         player for player in all_players
                         if player not in connected_players
@@ -553,6 +554,7 @@ class LadderService(Service):
             msg = {"command": "match_cancelled", "game_id": game_id}
             for player in all_players:
                 if player.state == PlayerState.STARTING_AUTOMATCH:
+                    print(f"Setting player state to IDLE {player}")
                     player.state = PlayerState.IDLE
                 player.write_message(msg)
 
@@ -563,7 +565,8 @@ class LadderService(Service):
                 )
                 self.violation_service.register_violations(abandoning_players)
                 for player in non_connected_players:
-                    if player.lobby_connection:
+                    if player.lobby_connection and player.state == PlayerState.IDLE:
+                        print(f"Removing game_connection for {player}")
                         player.lobby_connection.game_connection = None
 
     async def get_game_history(
--- a/server/ladder_service/ladder_service.py
+++ b/server/ladder_service/ladder_service.py
@@ -550,7 +550,12 @@ class LadderService(Service):
                         ),
                         matchmaker_queue_game.c.matchmaker_queue_id == queue_id
                     )
-                ).order_by(game_stats.c.startTime.desc()).limit(limit)
+                ).order_by(
+                    game_stats.c.startTime.desc(),
+                    # Timestamps only have second resolution, so for this to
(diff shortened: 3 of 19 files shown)


NapMorax:
> Add a message informing player about violation, how many tries he has before ban and how long ban will be.

Make sure it shows the duration of the ban so that people do not think it is permanent!

Askaholic:
Any message I add on the server side will only be temporary anyways since all the necessary info is already sent to the client in a machine readable way that it can use to display the message with proper localization etc. Once that’s implemented on the client, I will delete the server notice messages.

BlackYps:
I think for the bans I would do 0, 6, 15, 45 minutes. I don't see a reason to let people dodge the queue more than once without punishment. Instead we can lower the times a little. I would also tick down the count in increments instead of completely resetting it at once.

Strogoo:
Yep, I agree. 1 (or 2 at max) try and then short ban, so others players in que can start a game in next wave and don't have to wait 3+ waves until player with bad connection will get banned.

Askaholic:
The bans are currently:
1 violation: None
2 violations: 10 minutes
3+ violations: 30 minutes

NapMorax:
> The bans are currently: 1 violation: None 2 violations: 10 minutes 3+ violations: 30 minutes

Should we not discuss the ban lengths for violations as a group? Looks okay to me but might want to open this discussion up a bit first.

Askaholic:
I believe this has already been discussed. If you have anything to add feel free to do it.

server/ladder_service/ladder_service.py
```diff
@@ -458,55 +519,38 @@ def get_player_mean(player):
...
 
             def make_game_options(player: Player) -> GameLaunchOptions:
-                return options._replace(
+                return GameLaunchOptions(
+                    mapname=mapname,
+                    expected_players=len(all_players),
+                    game_options=game_options,
                     team=game.get_player_option(player.id, "Team"),
-                    faction=player.faction,
+                    faction=game.get_player_option(player.id, "Faction"),
```
BlackYps, line 529: we are now setting this to `player.faction.value`. Is there a difference to plain `faction.value`?

server/ladder_service/ladder_service.py
```diff
@@ -565,32 +674,6 @@ async def shutdown(self):
...
-
-    clans = {player.clan for player in team}
-
-    if len(clans) == 1:
-        name = clans.pop() or player_1_name
-    else:
-        name = player_1_name
-
-    return f"Team {name}"
+class NotConnectedError(asyncio.TimeoutError):
```
BlackYps, line 677: I think GameClosedError and NotConnectedError should be defined at the same place.

server/ladder_service/ladder_service.py
```diff
@@ -519,6 +563,66 @@ def make_game_options(player: Player) -> GameLaunchOptions:
...
+    async def launch_match(
+        self,
+        game: LadderGame,
+        host: Player,
+        guests: list[Player],
+        make_game_options: Callable[[Player], GameLaunchOptions]
+    ):
+        # Launch the host
+        if host.lobby_connection is None:
+            raise NotConnectedError([host])
```
BlackYps, line 582: Why do we make this extra check? To fail faster?

server/ladder_service/ladder_service.py
```diff
@@ -519,6 +563,66 @@ def make_game_options(player: Player) -> GameLaunchOptions:
...
+
+                guest.lobby_connection.write_launch_game(
+                    game,
+                    is_host=False,
+                    options=make_game_options(guest)
+                )
+        try:
+            await game.wait_launched(60 + 10 * len(guests))
+        except asyncio.TimeoutError:
+            connected_players = game.get_connected_players()
```
BlackYps, line 620: I don't understand when we set a player as connected. It seems that we need to call `set_player_option()` at some point to mark him as connected, but I can't find when this would actually happen.

BlackYps (changes requested):

NapMorax:
> I believe this has already been discussed. If you have anything to add feel free to do it.

Yeah, I see the above. I agree with the times and offense count and have shared it with the general forum announcements, matchmaker team, etc. I don't want to delay this going out so will just say "go ahead."

Nice work, Askaholic! Kudos to @Strogoo  for the push.

server/ladder_service/ladder_service.py
Askaholic: It's done when the game host sends the `PlayerOption` GPGNet command (through GameConnection). This is how we know that the host has successfully established a connection to that player.

server/players.py
```diff
@@ -173,7 +173,7 @@ def __repr__(self) -> str:
                 f"clan={self.clan}, game_count={dict(self.game_count)})")
 
     def __hash__(self) -> int:
-        return self.id
+        return hash(self.id)
```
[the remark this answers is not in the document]
Askaholic: This should produce more uniformly distributed values. It probably doesn't matter too much in practice since none of our dictionaries that use players as keys should be getting anywhere close to being as large as the total player base, but in theory at least this should give general better dict performance and probably prevent some DOS attacks as well.

server/gameconnection.py
```diff
@@ -517,6 +517,13 @@ async def handle_rehost(self, *args: list[Any]):
         """
         pass
 
+    async def handle_launch_status(self, status: str):
+        """
+        Currently is sent with status `Rejected` if a matchmaker game failed
+        to start due to players using differing game settings.
+        """
+        pass
```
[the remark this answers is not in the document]
Askaholic: No, I added this when we made the game options changes to autolobby.lua. The autolobby now sends this if a player detects that the host is not using the game options that they should be.

server/lobbyconnection.py
```diff
@@ -1031,6 +1031,13 @@ def write_launch_game(
         is_host: bool = False,
         options: GameLaunchOptions = GameLaunchOptions(),
     ) -> None:
+        if self.game_connection is not None:
+            self._logger.warning(
+                "%s launched a new game while old GameConnection was active",
```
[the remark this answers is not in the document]
Askaholic: There are a few ways I think this could happen:
1. The game hard crashes, so it doesn't send `GameState: Ended` on exit. Maybe the ICE adapter will detect that and still send the command though, idk. Either way, I guess the ICE adapter could crash as well and then we'd have the same problem.
2. The server sends a `game_launch` command but for whatever reason the client doesn't actually open the game. Because the way the server code is written, a `GameConnection` object has to exist in order to receive GPGNet commands however, we don't actually know for sure if the game has opened until we receive a GPGNet command. This is the most likely scenario, in which case we can safely just remove the old `GameConnection` obejct.

server/ladder_service/violation_service.py
```diff
@@ -0,0 +1,127 @@
...
+
+    A violation could be anything, but it is usually any time a player fails
+    to connect to a game.
+    """
+
+    def __init__(self):
+        self.violations: dict[Player, Violation] = {}
+
+    async def initialize(self):
+        self._cleanup_task = at_interval(5, func=self.clear_expired)
```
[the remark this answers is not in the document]
Askaholic: Because `at_interval` calls `asyncio.get_running_loop`, which might not exist yet when `__init__` is called, but will always exist when called from an `async` function. This is mostly an issue with the unit tests iirc.

server/ladder_service/__init__.py
```diff
@@ -0,0 +1,7 @@
+from .game_name import game_name
+from .ladder_service import LadderService
+
+__all__ = (
+    "LadderService",
+    "game_name",
```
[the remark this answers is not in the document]
Askaholic: Because you should import it with the full module path `from ladder_service.violation_service import ViolationService`

server/ladder_service/ladder_service.py
```diff
@@ -216,6 +225,48 @@ def start_search(
...
+                    "expires_at": violation.get_ban_expiration().isoformat()
+                }
+                for p, violation in timeouts.items()
+            ]
+            for player in players:
+                player.write_message({
+                    "command": "search_timeout",
+                    "timeouts": times
+                })
+                # TODO: Do we need this or is `search_timeout` enough?
```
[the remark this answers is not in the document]
Askaholic: Hence the TODO :P

server/ladder_service/ladder_service.py
```diff
@@ -216,6 +225,48 @@ def start_search(
         queue_name: str,
         on_matched: OnMatchedCallback = lambda _1, _2: None
     ):
+        timeouts = self.violation_service.get_violations(players)
+        self._logger.debug("timeouts: %s", timeouts)
```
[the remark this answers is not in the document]
Askaholic: I did when I was debugging it, but now that it's working probably not.

server/ladder_service/ladder_service.py
Askaholic: I don't think so, since Enums should get serialized to their value anyways when running them through `json.dumps`

Askaholic: It might make it harder to tell what the faction is in the TRACE logs as before the faction would be printed as `repr` of the enum which includes both the name and value.

server/ladder_service/ladder_service.py
Askaholic: Because otherwise we'd just get a NoneType error and it wouldn't include the information about which player caused the problem

server/ladder_service/ladder_service.py
Askaholic: Hmm, GameClosedError is only raised from withing LadderGame, but NotConnectedError is only raised from LadderService. So IDK where a better place to put them would be.

server/ladder_service/violation_service.py
```diff
@@ -0,0 +1,127 @@
...
+
+    def is_expired(self, now: Optional[datetime] = None) -> bool:
+        """
+        Whether the violation history should be reset. This is different from
+        the ban expiration time which should be checked by calling
+        `get_ban_expiration`.
+        """
+        now = now or datetime_now()
+        # TODO: Config?
+        return self.time + timedelta(hours=1) <= now
```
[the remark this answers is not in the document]
Askaholic: Can we just make this an issue to fix later? Because it's gonna take a bit more refactoring effort than I want put in right now and I don't think it's worth blocking the PR for this.

PR #892, opened 2022-03-12
add "hosted_at" field to game_info message

KaukaHan:
closes issue/#333

### DIFF
--- a/server/games/game.py
+++ b/server/games/game.py
@@ -10,7 +10,6 @@ from sqlalchemy import and_, bindparam
 from sqlalchemy.exc import DBAPIError
 from sqlalchemy.sql.functions import now as sql_now
 
-from server.timing import datetime_now
 from server.config import FFA_TEAM
 from server.db.models import (
     game_player_stats,
@@ -28,6 +27,7 @@ from server.games.game_results import (
     resolve_game
 )
 from server.rating import InclusiveRange, RatingType
+from server.timing import datetime_now
 
 from ..players import Player, PlayerState
 from .typedefs import (
--- a/server/games/game.py
+++ b/server/games/game.py
@@ -1,4 +1,5 @@
 import asyncio
+import datetime
 import json
 import logging
 import time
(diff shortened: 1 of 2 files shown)


server/games/game.py
```diff
@@ -900,6 +902,7 @@ def to_dict(self):
             "num_players": len(connected_players),
             "max_players": self.max_players,
             "launched_at": self.launched_at,
+            "hosted_at": self.hosted_at,
```
Askaholic, line 905: Need to convert to ISO string

KaukaHan: is timezone information required ?

Askaholic: Yes I think so. There should be a helper function in`server.timing.datetime_now` that takes care of the time zone.

server/games/game.py
```diff
@@ -70,6 +72,7 @@ def __init__(
         enforce_rating_range: bool = False,
         max_players: int = 12,
         setup_timeout: int = 60,
+        hosted_at: datetime = datetime_now
```
Askaholic, line 75: `datetime_now` is a function. You probably need to set the default to `None` and then do `self.hosted_at = hosted_at or datetime_now()`

server/games/game.py
```diff
@@ -70,6 +72,7 @@ def __init__(
         enforce_rating_range: bool = False,
         max_players: int = 12,
         setup_timeout: int = 60,
+        hosted_at: datetime = datetime_now()
```
Askaholic, line 75: This isn't going to work because argument default are only created once at the time the function is defined. I.e. this default datetime will always be the time when the server is booted up. You have to do it like I wrote earlier, where you use `None` as the default and call the function later. 

It's the same reason why you have to be really careful with using mutable values as defaults:
```python
def foo(data=[]):
    data.append(10)
    return data
```
Just paste that into the interpreter and call it a few times. You'll see what I mean.

This one will work as expected:
```python3
def foo(data=None):
    data = data or []
    data.append(10)
    return data
```

PR #895, opened 2022-03-14
Issue/#893 player hash implementation

### LINKED ISSUE #893: Issues with `Player.__hash__` and `Player.__eq__` implementation
Right now Player objects are considered equal if their ID's are equal. However, this can cause issues when players reconnect where a new Player object is created and there are still some old ones hanging around.

For example:
1. Player starts a matchmaker game
2. The game ends and the player leaves the game and immediately reconnects
3. The player tries to queue for matchmaker

What happens is in step 3, sometimes the `on_connection_lost` logic has not triggered yet for the services, so the old Party object is re-used for the new player due to the hash/eq implementations. However, the party object still references the old player which has a player state of PLAYING, so the new player is falsely prevented from queuing.

Really we should rework the hash implementation so that different objects have different hashes. My initial thought was something like `hash(id(self))` or `hash((self.id, id(self)))`, but this seems to cause some of the unit tests to fail due to a changed ordering of entries in certain dictionaries, so it will require a bit of work to figure out why that changes happens and how to make it work properly.

Askaholic:
Removes the custom implementation of the `__hash__` and `__eq__` which seem to have done nothing but cause bugs. The only thing we lose is that `set`'s of Players will no longer be sorted by player id when iterated over. Seems like sets (unlike dicts) still do not preserve insertion order, so I had to change one `to_dict` method to sort the set items by player id.

Closes #893

### DIFF
--- a/tests/unit_tests/test_matchmaker_algorithm_bucket_teams.py
+++ b/tests/unit_tests/test_matchmaker_algorithm_bucket_teams.py
@@ -24,6 +24,7 @@ def player_factory(player_factory):
         deviation: int = 500,
         ladder_games: int = config.NEWBIE_MIN_GAMES + 1,
         name=None,
+        **kwargs
     ):
         """Make a player with the given ratings"""
         player = player_factory(
@@ -31,6 +32,7 @@ def player_factory(player_factory):
             ladder_games=ladder_games,
             login=name,
             lobby_connection_spec=None,
+            **kwargs
         )
         return player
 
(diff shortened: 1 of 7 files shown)


tests/unit_tests/test_matchmaker_algorithm_bucket_teams.py
```diff
@@ -214,26 +219,33 @@ def test_distribute_pairs_1(player_factory):
...
+        for i in range(8)
+    ]
     searches = [(Search([player]), 0) for player in players]
     p1, p2, p3, p4, p5, p6, p7, p8 = players
 
     grouped = [
         search.players for search in algorithm.bucket_teams._distribute(searches, 2)
     ]
-    assert grouped == [[p1, p4], [p2, p3], [p5, p8], [p6, p7]]
+    assert grouped == [[p1, p7], [p2, p8], [p3, p5], [p4, p6]]
```
Askaholic, line 232: Hilariously, these tests never actually worked properly since all player objects had the same id and therefore compared equal. They were only able to assert that the returned list had the correct shape.

BlackYps (approved):
looks good to me

PR #896, opened 2022-03-21
Set game_id_counter to 0 if no games

Brutus5000:
If the game_stats table is empty, the creation of new games will fail. Picking the max(id) of game_stats is null,
but incrementing null by 1 throws an error.

### DIFF
--- a/server/game_service.py
+++ b/server/game_service.py
@@ -82,7 +82,8 @@ class GameService(Service):
             # doing LAST_UPDATE_ID to get the id number, and then doing an UPDATE when the actual
             # data to go into the row becomes available: we now only do a single insert for each
             # game, and don't end up with 800,000 junk rows in the database.
-            self.game_id_counter = await conn.scalar("SELECT COALESCE(MAX(id),0) FROM game_stats")
+            sql = "SELECT MAX(id) FROM game_stats"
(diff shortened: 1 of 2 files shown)


server/game_service.py
```diff
@@ -82,7 +82,7 @@ async def initialise_game_counter(self):
             # doing LAST_UPDATE_ID to get the id number, and then doing an UPDATE when the actual
             # data to go into the row becomes available: we now only do a single insert for each
             # game, and don't end up with 800,000 junk rows in the database.
-            self.game_id_counter = await conn.scalar("SELECT MAX(id) FROM game_stats")
```
Askaholic, line 85: Unit test?

You could probably also do it in python rather than sql:

```python
id = await conn.scalar(…) or 0
```

Brutus5000: Are you sure? I worked with Python for years and I am not. Await returns some async object...? So o I need to put it in brackets or not for operator priorities? Or does it not matter? Does the `or` even fire or not? I don't know what type conn.scalar even returns in the null case...

The SQL method is dead simple and clear.

Askaholic: Yes I am sure, you don’t need parenthesis. `await` has higher precedence than `or` so you can do this.

I imagine that sql `null` is converted to `None` which is what was giving you the errors in the first place.

PR #900, opened 2022-04-04
Issue/#858 validity in game info

### LINKED ISSUE #858: Add validity to `game_info` message
It would be nice if people could see that a game will be unrated because of certain lobby settings during the lobby phase. Obviously we can’t predict if there will be desyncs or disconnects, but if the game is modded or a bad map is selected we could certainly show it as unrated. 

We should add a field `validity` to the `game_info` message that will contain certain validity strings (we would probably have to figure out some better names since the naming of the enum is currently pretty inconsistent).

Adding this would require some reworking of when the validity is calculated as it’s currently only computed once when the game is launched. We would probably want to break a certain subset of those validity checks out into checks that are recomputed whenever the game_info message is requested. Those pre-launch validity checks should not make writes to the database since they aren’t final yet, so some of the code will have to be refactored.

Examples of validity states that should always be checked:
* bad map
* bad mod
* wrong victory condition
* has AI

These validity states could only be checked when the game is full as they may fail and pass again as players join:
* uneven teams
* single player

Askaholic:
Adds game validity to the `game_info` messages. This requires quite a lot of refactoring because the validity needs to be re-evaluated every time any of the attributes that affect validity change. Previously, each of the checks only ran at certain times such as when the game launched or after it was finished. Refactoring this was also made more difficult by the fact that the game handling logic is already such a mess. Hopefully this approach of moving the validation code out of the `Game` class helps with that, but I'm still not 100% happy with it. There are still some problems with it:

- The `ValidationRule`s are very tightly coupled to the `Game` class, some of them even use 'private' variables from `Game`
- The validity computation is somewhat expensive as it is likely to be called many times, especially while a game is in the lobby state. It is also extremely difficult to cache because it has so many dependencies that would need to invalidate the cache whenever they change.

Some other issues that came up:
- Whenever the `ScenarioFile` game option is set, we need to query the database in order to get the map id and to check if the map is ranked. This is an async operation, but it doesn't otherwise make sense to make `set_game_option` an async function. So it's a little unclear where exactly the `await` should happen.
- Games are marked as dirty quite liberally, even if nothing that affects the `to_dict` method would actually change
- We might want to rename some of the validity states since the enum names are used in the protocol

Closes #858

### DIFF
--- a/server/gameconnection.py
+++ b/server/gameconnection.py
@@ -463,10 +463,6 @@ class GameConnection(GpgNetServerProtocol):
             return
 
         elif state == "Lobby":
-            # TODO: Do we still need to schedule with `ensure_future`?
-            #
-            # We do not yield from the task, since we
-            # need to keep processing other commands while it runs
             await self._handle_lobby_state()
 
         elif state == "Launching":
--- a/server/gameconnection.py
+++ b/server/gameconnection.py
@@ -314,7 +314,9 @@ class GameConnection(GpgNetServerProtocol):
             self._logger.warning("OperationComplete called for non-coop game")
             return
 
-        if self.game.validity != ValidityState.COOP_NOT_RANKED:
+        validity = self.game.get_validity()
+        if validity is not ValidityState.COOP_NOT_RANKED:
+            self._logger.info("Game was not valid: %s", validity)
             return
 
         secondary, delta = secondary, str(delta)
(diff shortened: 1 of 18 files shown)


Eternal-ll:
Api returns HAS_AI, but server defined only HAS_AI_PLAYERS

https://github.com/FAForever/server/blob/7c96b3b727867d7ea3aa4893d55f33b97acf3c87/server/games/typedefs.py#L68

image

PR #904, opened 2022-05-02
Issue/#903 Fix games being rated multiple times

### LINKED ISSUE #903: Sometimes games are rated three or more times
There are some entries in the `leaderboard_rating_journal` that indicate that some games are being rated multiple times. The rating changes show that it is not just an issue of the same record being inserted multiple times, but actually the game appears to be rated multiple times.

Example for a single player:
https://api.faforever.com/data/gamePlayerStats/33454629?include=ratingChanges

Example for the whole game:
https://api.faforever.com/data/game/16909196?include=playerStats,playerStats.ratingChanges

You can see that every player in that game got 3 entries in the rating journal, so the game was rated 3 times in a row, and each time the players ratings after the previous change were used.

Askaholic:
The game ending logic could be triggered multiple times simultaneously. Adding a lock ensures that the game state is checked in serial so the game ending logic won't trigger more than once.

Closes #903

### DIFF
--- a/server/games/game.py
+++ b/server/games/game.py
@@ -119,6 +119,7 @@ class Game:
         }
         self.mods = {}
         self._hosted_future = asyncio.Future()
+        self._finish_lock = asyncio.Lock()
 
         self._logger.debug("%s created", self)
         asyncio.get_event_loop().create_task(self.timeout_game(setup_timeout))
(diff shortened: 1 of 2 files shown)


BlackYps (approved):
Looks good to me

PR #912, opened 2022-07-23
Remove steam id checks from server

Sheikah45:
The steam ID check in the lobby server is entirely redundant and unnecessary.

Login with tokens requires the lobby scope to be in the token and to have the lobby scope requires that game ownership has been verified.

So everyone who logs in has been verified. also this enables disabling the deprecated password login on the server to force a verified login.

Also only checking steam IDs is not valid as verification can occur using gog as well now.

Also this will allow for the eventual migration of sensitive user details to be stored on a separate machine.


server/lobbyconnection.py
```diff
@@ -555,38 +519,42 @@ async def command_auth(self, message):
...
         unique_id: str,
         method: str
     ):
         conforms_policy = await self.check_policy_conformity(
             player_id, unique_id, self.session,
-            ignore_result=(
-                steamid is not None or
-                self.player_service.is_uniqueid_exempt(player_id)
-            )
+            # All players are required to have game ownership verified so this is for informational purposes only
```
Askaholic, line 556:
```suggestion
            # All players are required to have game ownership verified so this
            # is for informational purposes only
```

Askaholic (changes requested):

Askaholic (approved):

PR #915, opened 2022-07-30
Issue/#914 configuration refresh

### LINKED ISSUE #914: Config is refreshed after `USE_UVLOOP` variable is checked
When running the devserver I noticed this:

```
INFO     Jul 30  11:03:47 root                           Lobby dev (Python 3.9.9) on linux
INFO     Jul 30  11:03:47 root                           Event loop: <uvloop.Loop running=True closed=False debug=False>
...
INFO     Jul 30  11:03:47 ConfigurationStore             New value for USE_UVLOOP: True -> False
```

Even though the dev config has uvloop disabled, it's still used because the config values aren't refreshed until after the loop is already set up and running. We should call `config.refresh()` in the main setup code in `server.py` as early as we can.

Askaholic:
Refreshes the config before attempting to use it

Closes #914

### DIFF
--- a/server.py
+++ b/server.py
@@ -153,6 +153,9 @@ if __name__ == "__main__":
         )
     )
     logger.addHandler(stderr_handler)
+    logger.setLevel(logging.INFO)
+
+    config.refresh()
     logger.setLevel(config.LOG_LEVEL)
 
     if config.USE_UVLOOP:


Sheikah45 (approved):
This works for me. 

Although admittedly I never ran into this issue previously because I always used the environment variable rather than the command argument to start the server.

Askaholic:
Oh, I see. It's only happening because the pipenv script uses the command line parameter to set the config file. Turns out this means that in the docker container the first time the file is loaded, the changed values aren't logged since the refresh happens at import time which is before the log level is actually configured to be INFO or lower. I supposed this is a good thing since it means our database secrets and stuff are not printed to the log, but it's only by accident.

PR #917, opened 2022-08-20
Refine matchmaker metrics

BlackYps:
We need histograms instead of gauges to be able to properly display the match quality in Grafana. The number of currently active matchmaker games is also nice to have.

### DIFF
--- a/server/game_service.py
+++ b/server/game_service.py
@@ -9,6 +9,7 @@ import aiocron
 from sqlalchemy import select
 
 from server.config import config
+
 from . import metrics
 from .core import Service
 from .db import FAFDatabase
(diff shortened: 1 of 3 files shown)


Sheikah45:
Might you also want to track the failed matches as well so you have a clear indication of what fails and what succeeds? Just a thought.

BlackYps:
We're already doing that. There is already a panel in Grafana which shows how many matches succeed and if they don't how many time out and how many are aborted by a player.

Askaholic (approved):

Sheikah45:
Got it although I was talking about the variety and imbalance of the failed matches as well

BlackYps:
As matches fail for technical reasons I think the distribution of variety and imbalance should be the same for failed and actually started matches. So I don't think we can gain any information from that.
Besides, as those matches do not start the match quality is irrelevant anyway.
Or maybe I misunderstand you?

PR #921, opened 2022-09-24
Tournament game service

### LINKED ISSUE #675: RFC: Take matchmaking requests from RabbitMQ
This should be the missing piece in my envisioned galactic war architecture.

Similar to ladder or tmm I need a way to force players into an automatch (no lobby).
As this is not always successfull, I also need a feedback loop.

So we should implement a rabbitmq protocol such as:

**routing key request.match.create:**

```json
{
  "requestId": "<random uuid>",
  "featuredMod": 13,
  "mapVersion": 1234,
  "participants": [
   {
       "playerId": 13443,
       "team": 2,
        "slot": 1
   },
   {
       "playerId": 13443,
       "team": 3,
       "slot": 2
   },
]
}
```

**routing key request.match.success:**
```json
{
  "requestId": "<uuid of original request>",
  "gameId": 98765
}
```

**routing key request.match.error:**
```json
{
  "requestId": "<uuid of original request>",
  "errorCode": 123,
  "args": [
  {
     "playerId": 13443
   }
  ]
}
```

The error codes need to be defined precisely as it might cause punishments if it fails because a player was already in a game or similar.

1-alex98:
Closes https://github.com/FAForever/server/issues/675


tests/unit_tests/test_gameconnection.py
```diff
@@ -415,7 +415,8 @@ async def test_handle_action_GameOption(
     assert game.max_players == 7
     # I don't know what these paths actually look like
     await game_connection.handle_action("GameOption", ["ScenarioFile", "C:\\Maps\\Some_Map"])
-    assert game.map_file_path == "maps/some_map.zip"
+    assert game.map_name == "some_map"
+    assert game.map_scenario_path == "C:/Maps/Some_Map"
```
1-alex98, line 419: since game is a mock the setter is not invoked and map_file_path is not set by the setter

server/game_service.py
```diff
@@ -37,12 +37,12 @@ class GameService(Service):
...
-        player_service,
-        game_stats_service,
-        rating_service: RatingService,
-        message_queue_service: MessageQueueService
+            self,
+            database: FAFDatabase,
+            player_service,
+            game_stats_service,
+            rating_service: RatingService,
+            message_queue_service: MessageQueueService
```
Askaholic, line 45: Please fix these. This is not how we are formatting in this repo. We use only 1 level of indentation.

server/tournament_service.py
```diff
@@ -0,0 +1,282 @@
...
+            game.players.append(player)
+        return game.state == TournamentGameState.SCHEDULED
+
+    async def add_tournament_game(self, game):
+        self._games.add(game)
+
+    async def on_is_ready_response(self, message, player):
+        game: TournamentGame = await self._get_game_for_request_id(message["request_id"])
+        if game is None:
+            await player.lobby_connection.send_warning("You try to ready up for a game that does not exist")
```
Askaholic, line 165: Generally these should be either `ClientError`'s or some custom message that can be localized by the client depending on whether or not reaching this code indicates a client bug.

server/tournament_service.py
```diff
@@ -0,0 +1,282 @@
...
+        game.players_ready_ids.add(player.id)
+        if game.is_ready():
+            await self._launch(game)
+
+    async def _get_game_for_request_id(self, request_id):
+        for game in self._games.copy():
+            if game.request_id == request_id:
+                return game
+
+    async def _launch(self, tournament_game: TournamentGame):
```
Askaholic, line 188: This looks like a rather large copy/paste. We should try to refactor this.

server/tournaments/__init__.py
```diff
@@ -0,0 +1,11 @@
+"""
+The team matchmaking system
+
+Used to keep track of current player parties, manage players joining/leaving and matching them against each other
+used for matchmaking in the global rating system
+"""
+__all__ = [
+    "TournamentGame"
+]
```
Askaholic, line 9: I just recently learned that PEP 8 actually says to put the `__all__` variable before the imports. That probably wants to be changed all over the place in this code because I've been doing it the other way around. I've also been using tuples instead of lists. Not sure if that's a good thing though :P

server/tournaments/tournament_game.py
```diff
@@ -0,0 +1,70 @@
...
+    PLAYER_NOT_CONFIRMING = 7
+    PLAYER_NOT_STARTING = 8
+    PLAYER_NOT_CONNECTING = 9
+
+    def is_done(self):
+        return self not in [
+            TournamentGameState.SCHEDULED,
+            TournamentGameState.CONFIRMATION_PENDING,
+            TournamentGameState.STARTING
+        ]
```
Askaholic, line 23: I'd probably use a set.
```suggestion
        return self not in {
            TournamentGameState.SCHEDULED,
            TournamentGameState.CONFIRMATION_PENDING,
            TournamentGameState.STARTING
        }
```

server/tournaments/tournament_game.py
```diff
@@ -0,0 +1,70 @@
...
+
+
+class TournamentGame:
```
Askaholic, line 26: IMO this is confusing since there's another class called `TournamentGame`.

server/tournaments/tournament_game.py
```diff
@@ -0,0 +1,70 @@
...
+    def __init__(self, create_tournament_request: dict):
+        self.created_time = datetime.utcnow()
```
Askaholic, line 28: Use `datetime_now` helper

server/tournaments/tournament_game.py
```diff
@@ -0,0 +1,70 @@
...
+        self.players = list()
```
Askaholic, line 29:
```suggestion
        self.players = []
```

server/tournaments/tournament_game.py
```diff
@@ -0,0 +1,70 @@
...
+        self.response_time_seconds = 30
+
+    def __contains__(self, player: Player) -> bool:
+        return player in self.players
+
+    def __iter__(self):
+        return iter(self.players)
+
+    def is_ready(self) -> bool:
+        return set([player.id for player in self.players]) == self.players_ready_ids
```
Askaholic, line 48:
```suggestion
        return set(player.id for player in self.players) == self.players_ready_ids
```

server/tournaments/tournament_game.py
```diff
@@ -0,0 +1,70 @@
...
+    def get_slot_of_player(self, player) -> int:
+        return self.get_participant_of_player(player)["slot"]
+
+    def get_faction_of_player(self, player) -> int:
+        return self.get_participant_of_player(player)["faction"]
+
+    def get_participant_of_player(self, player):
+        for participant in self.participants:
+            if participant["player_id"] == player.id:
+                return participant
```
Askaholic, line 66: Consider constructing a lookup table for participants by player_id in the `__init__`

server/tournaments/tournament_game.py
```diff
@@ -0,0 +1,70 @@
...
+    def is_done(self):
+        return self not in [
+            TournamentGameState.SCHEDULED,
+            TournamentGameState.CONFIRMATION_PENDING,
+            TournamentGameState.STARTING
+        ]
+
+
+class TournamentGame:
+    def __init__(self, create_tournament_request: dict):
```
Askaholic, line 27: You could use the dict unpacking trick here again.

server/tournaments/tournament_game.py
```diff
@@ -0,0 +1,70 @@
...
+
+    def __iter__(self):
+        return iter(self.players)
+
+    def is_ready(self) -> bool:
+        return set([player.id for player in self.players]) == self.players_ready_ids
+
+    def is_confirmation_overdue(self):
+        time_passed = datetime.utcnow() - self.created_time
+        return time_passed.seconds > self.response_time_seconds + 5
```
Askaholic, line 52: Why hardcode +5 here? Just set the response time to 35 seconds?

Askaholic (commented):
Sort of half reviewed it. Biggest thing is please undo the indentation changes. It looks like you might have an autoformatter or something that is changing those.

server/game_service.py
1-alex98: Then why did nobody set it that way in intellij code style?

Askaholic: We don't use intellij

1-alex98: I meant Pycharm but yeah. Also making an .editorconfig would do the trick

server/ladder_service/ladder_service.py
```diff
@@ -427,12 +427,14 @@ def on_match_found(
                 "Error processing match between searches %s, and %s",
                 s1, s2
             )
+            for player in s1.players + s2.players:
+                player.state = PlayerState.IDLE
```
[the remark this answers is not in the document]
1-alex98: yes I have no idea I will remove it

server/ladder_service/ladder_service.py
```diff
@@ -516,23 +518,23 @@ def get_player_mean(player: Player) -> float:
             if game_options:
                 game.gameOptions.update(game_options)
 
-            mapname = re.match("maps/(.+).zip", map_path).group(1)
+            map_name = re.match("maps/(.+).zip", map_path).group(1)
```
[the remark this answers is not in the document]
1-alex98: Where to?

server/tournaments/tournament_game.py
1-alex98: I did give the client 5 more seconds to answer because of network delay etc.
What do you mean by just set the response time higher? That would also cause the client to display a longer time to the user.

server/tournament_service.py
```diff
@@ -0,0 +1,282 @@
...
+            self._logger.info("Received Tournament game message")
+            await self._process_create_game(message)
+        except Exception:
+            self._logger.exception("Unknown failure creating tournament game")
+            await self.message_queue_service.publish(
+                config.MQ_EXCHANGE_NAME,
+                "tourneylauncher.createGame.failed",
+                {
+                    "request_id": message.correlation_id,
+                    "error_code": "OTHER"
```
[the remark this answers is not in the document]
1-alex98: Not sure if it is that important :thinking: This is an unexpected error so not much the server could do any automated handling with

server/tournament_service.py
1-alex98: ok

server/message_queue_service.py
```diff
@@ -48,9 +52,27 @@ async def initialize(self) -> None:
...
         await self._declare_exchange(config.MQ_EXCHANGE_NAME, ExchangeType.TOPIC)
+        await self._connect_listening_queues()
+        self._is_ready = True
+
+    async def _connect_listening_queues(self):
+        for listening_info in self._listening_info:
+            await self._declare_listening_queue(listening_info)
+
+    async def listen_to_message(self, queue_name: str, routing_key: str,
+                                callback: Callable[[AbstractIncomingMessage], Any]):
```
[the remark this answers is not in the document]
Askaholic: We format like this to avoid long lines:
```suggestion
    async def listen_to_message(
        self, 
        queue_name: str, 
        routing_key: str,
        callback: Callable[[AbstractIncomingMessage], Any]
    ):
```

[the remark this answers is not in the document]
Askaholic: Also it makes git diffs smaller when you add/remove parameters which makes merge conflicts less common and easier to resolve.

server/tournaments/tournament_game.py
Askaholic: Ok I see. I didn't catch that the response time is sent to the client.

.editorconfig
```diff
@@ -0,0 +1,766 @@
+[*]
+charset = utf-8
```
Askaholic, line 2: I need to look into this editorconfig stuff. Would probably be best to do it in a separate PR. This file also looks like it has way too much stuff in it. I would want to trim out anything we don't need.

server/ladder_service/ladder_service.py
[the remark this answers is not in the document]
Askaholic: How about just making use of the properties you added to the Game class? The map_file_path is already being set on line 532.

```python3
map_name = game.map_name
```

server/rating_service/rating_service.py
```diff
@@ -141,6 +140,10 @@ async def _handle_rating_queue(self) -> None:
     async def _rate(self, summary: GameRatingSummary) -> None:
         assert self._rating_type_ids is not None
 
+        if summary.rating_type is None:
+            self._logger.debug(f"Not processing game {summary.game_id} since it is not rated.")
```
Askaholic, line 144: We don't use f-strings in log messages.

```suggestion
            self._logger.debug("Not processing game %s since it is not rated.", summary.game_id)
```

PR #922, opened 2022-10-08
Matchmaker violation wording change

maudlin27:
Makes the matchmaker wording slightly less alarming

### DIFF
--- a/tests/unit_tests/test_ladder_service.py
+++ b/tests/unit_tests/test_ladder_service.py
@@ -274,7 +274,7 @@ async def test_start_game_timeout_on_send(
         mock.call({
             "command": "notice",
             "style": "info",
-            "text": You have caused a matchmaking connection failure 1 time(s). "
+            "text": "You have caused a matchmaking connection failure 1 time(s). "
                     "Multiple failures result in temporary time-outs from matchmaker. "
                     "Please seek support on the forums or discord for persistent issues."
         })
--- a/tests/unit_tests/test_ladder_service.py
+++ b/tests/unit_tests/test_ladder_service.py
@@ -218,10 +218,10 @@ async def test_start_game_timeout(
             "command": "notice",
             "style": "info",
             "text": (
-                    "You have caused a matchmaking connection failure 1 time(s). "
-                    "Multiple failures result in temporary time-outs from matchmaker. "
-                    "Please seek support on the forums or discord for persistent issues."
-                )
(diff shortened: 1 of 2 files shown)


Askaholic:
Looks good to me. There's another test that needs to be updated though.

maudlin27:
Updated for the other test so it has consistent wording (although getting different errors with the unit-test check now)

tests/unit_tests/test_ladder_service.py
```diff
@@ -217,8 +217,9 @@ async def test_start_game_timeout(
         mock.call({
             "command": "notice",
             "style": "info",
-            "text": "You have received 1 violations for failing to connect "
-            "to matchmaker games."
+            "text": "You have caused a matchmaking connection failure 1 time(s). "
+                    "Multiple failures result in temporary time-outs from matchmaker. "
+                    "Please seek support on the forums or discord for persistent issues."
```
Askaholic, line 222: Can you format the multiline strings like this.
```suggestion
            "text": (
                "You have caused a matchmaking connection failure 1 time(s). "
                "Multiple failures result in temporary time-outs from matchmaker. "
                "Please seek support on the forums or discord for persistent issues."
            )
```

tests/unit_tests/test_ladder_service.py
```diff
@@ -273,8 +274,9 @@ async def wait_forever(*args, **kwargs):
...
```
Askaholic, line 279: Here as well

Askaholic (commented):
I think the test fail is probably a fluke. Those integration tests tend to be a little flaky.

tests/unit_tests/test_ladder_service.py
maudlin27: Added the brackets/format

Askaholic: Aaah, sorry you didn't format it quite like I said. There should only be 1 additional level of indentation after the parens.

maudlin27: Sorry hadn't realised the importance of indents in python (assumed it was just a code readability thing) - Is the updated commit what you had in mind?

Askaholic: Yep that looks good! In this case it is just a code readability thing. We use one level of indentation everywhere in the code base so I'm just trying to keep it consistent.

PR #926, opened 2022-10-22
Tune matchmaker

BlackYps:
This is implementing the changes discussed in this thread: https://forum.faforever.com/topic/4945/allow-us-to-pick-rating-brackets-for-tmm/58

Additionally we lower the queue pop timer to provide a better user experience.

### DIFF
--- a/server/ladder_service/ladder_service.py
+++ b/server/ladder_service/ladder_service.py
@@ -99,7 +99,7 @@ class LadderService(Service):
                 queue.featured_mod = info["mod"]
                 queue.rating_type = info["rating_type"]
                 queue.team_size = info["team_size"]
-                queue.rating_peak = self.fetch_rating_peak(conn, info["rating_type"])
+                queue.rating_peak = await self.fetch_rating_peak(conn, info["rating_type"])
             queue.map_pools.clear()
             for map_pool_id, min_rating, max_rating in info["map_pools"]:
                 map_pool_name, map_list = map_pool_maps[map_pool_id]
@@ -237,15 +237,15 @@ class LadderService(Service):
         )
 
         rating_peak = 1000.0
-        if len(rows) > 0:
+        if rows.rowcount > 0:
             rating_peak = statistics.mean(
                 row.rating_mean_before - 3 * row.rating_deviation_before for row in rows
             )
 
-        if len(rows) < 100:
+        if rows.rowcount < 100:
             self._logger.warning(
                 "Could only fetch %s ratings for %s queue.",
-                len(rows),
+                rows.rowcount,
                 rating_type
             )
 
--- a/server/ladder_service/ladder_service.py
+++ b/server/ladder_service/ladder_service.py
@@ -244,6 +244,7 @@ class LadderService(Service):
                 rating_peak = statistics.mean(
                     row.rating_mean_before - 3 * row.rating_deviation_before for row in rows
                 )
+            metrics.leaderboard_rating_peak.labels(rating_type).set(rating_peak)
 
             if rowcount < 100:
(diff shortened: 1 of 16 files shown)


server/ladder_service/ladder_service.py
```diff
@@ -221,6 +225,48 @@ async def fetch_matchmaker_queues(self, conn):
                 errored.add(name)
         return matchmaker_queues
 
+    async def fetch_rating_peak(self, conn, rating_type):
+        result = await conn.execute(
```
Sheikah45, line 229: Why pass the conn as a parameter? Seems like this method should create the connection and use it.

Sheikah45 (changes requested):

server/ladder_service/ladder_service.py
BlackYps: This is a remnant of some refactoring. I'll fix it

server/matchmaker/algorithm/bucket_teams.py
```diff
@@ -21,12 +21,12 @@ class BucketTeamMatchmaker(Matchmaker):
     """
 
     def find(
-        self, searches: Iterable[Search], team_size: int
+            self, searches: Iterable[Search], team_size: int, rating_peak: float
```
Askaholic, line 24: Don't add extra indents
```suggestion
        self, searches: Iterable[Search], team_size: int, rating_peak: float
```

server/matchmaker/algorithm/stable_marriage.py
```diff
@@ -81,7 +81,7 @@ class StableMarriageMatchmaker(Matchmaker):
     """
 
     def find(
-        self, searches: Iterable[Search], team_size: int
+            self, searches: Iterable[Search], team_size: int, rating_peak: float
```
Askaholic, line 84: Same here

server/matchmaker/algorithm/team_matchmaker.py
```diff
@@ -57,12 +57,14 @@ class TeamMatchMaker(Matchmaker):
     9. repeat 8. until the list is empty
     """
 
-    def find(self, searches: Iterable[Search], team_size: int) -> tuple[list[Match], list[Search]]:
+    def find(
+            self, searches: Iterable[Search], team_size: int, rating_peak: float
```
Askaholic, line 61: Same here

tests/unit_tests/test_matchmaker_algorithm_team_matchmaker.py
```diff
@@ -224,16 +226,16 @@ def test_maximum_game_quality_for_even_teams(player):
     search = Search([player])
     team_a = CombinedSearch(*[search] * 4)
     team_b = CombinedSearch(*[search] * 4)
-    game = TeamMatchMaker().assign_game_quality((team_a, team_b), 4)
+    game = TeamMatchMaker().assign_game_quality((team_a, team_b), 4, 1000)
 
-    assert game.quality == 1.0
+    assert game.quality >= 1.0
```
Askaholic, line 231: Game quality can be better than 1 now?

Askaholic (changes requested):

tests/unit_tests/test_matchmaker_algorithm_team_matchmaker.py
BlackYps: Yes, because of the minority bonus. I set the minority bonus to 0 in the test above, which checks that the maximum game quality without bonuses is still 1. This test here also assures that the miority bonus is only positive.

Askaholic (approved):

PR #929, opened 2022-11-13
Issue/#878 Use MariaDB in github actions

### LINKED ISSUE #878: Update GitHub Actions to use MariaDB for tests
The db service should be using `mariadb:10.6` as we switched to mariadb a while ago.

Askaholic:
Also makes some other updates to actions to fix deprecation warnings.

Closes #878

### DIFF
--- a/.github/workflows/build.yml
+++ b/.github/workflows/build.yml
@@ -9,11 +9,11 @@ jobs:
   docker-image:
     runs-on: ubuntu-latest
     steps:
-      - uses: actions/checkout@v2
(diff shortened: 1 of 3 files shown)


Brutus5000 (approved):
I can't verify line 16 but the rest looks good :P

Askaholic:
I tested it in my shell and it worked. I need to double check the variable I’m writing to though. I might have copied the wrong one from the docs.

Sheikah45 (approved):
All looks good to me. I just migrated the client to the new output as well.

PR #932, opened 2022-11-19
Issue/#931 Violation tracking fix

### LINKED ISSUE #931: only party leader gets checked for violations
When a player gets a matchmaking ban because of too many violations for failing to connect, they can circumvent the ban by getting invited to another search party.

Askaholic:
The issue isn't with parties but with relogs. Since I refactored the `Player` objects to no longer hash by id because it was causing confusing and incorrect behavior in other parts of the code, using `Player`s as keys directly made the violation service think you were a different player after you relogged. Tracking by player id explicitly should fix this.

Closes #931

### DIFF
--- a/server/ladder_service/violation_service.py
+++ b/server/ladder_service/violation_service.py
@@ -63,24 +63,25 @@ class ViolationService(Service):
     """
 
     def __init__(self):
-        self.violations: dict[Player, Violation] = {}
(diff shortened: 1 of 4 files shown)


BlackYps (approved):
I didn't review too closely because I am constrained on time at the moment, but it looks fine to me

Issue #939, opened 2022-12-19
Messages using SimpleJsonProtocol are limited to 64 KiB (LimitOverrunError)


Askaholic:
The default limit for `StreamReader.readline` is to stop after 64 KiB. It seems that this limit is sometimes being reached:

```
Traceback (most recent call last):
  File "/usr/local/lib/python3.9/asyncio/streams.py", line 540, in readline
    line = await self.readuntil(sep)
  File "/usr/local/lib/python3.9/asyncio/streams.py", line 635, in readuntil
    raise exceptions.LimitOverrunError(
asyncio.exceptions.LimitOverrunError: Separator is found, but chunk is longer than limit
```

This probably never happened with the QDataStream protocol because it doesn't use `readline`, instead implementing the message deserialization in python using `readexactly`.

There are some messages that the client sends which can be pretty large, in particular the GameStats message contains a large JSON structure which can probably exceed the 64 KiB limit. We probably need to increase that limit to something much higher.

Askaholic:
We don't seem to have encountered this error for quite some time now, so I'm closing this issue.

PR #943, opened 2022-12-23
Issue/#939 Increase asyncio buffer size to allow for longer messages

Askaholic:
Bumped the max buffer size to 10 MiB. Hopefully this will be enough, I'm not sure what the max message size is that we see.

I'll leave the issue open for now so we can monitor the logs if this stops the errors. #939

### DIFF
--- a/server/servercontext.py
+++ b/server/servercontext.py
@@ -17,6 +17,9 @@ from .lobbyconnection import LobbyConnection
 from .protocol import DisconnectedError, Protocol, QDataStreamProtocol
 from .types import Address
 
+MiB = 2 ** 20
+LIMIT = 10 * MiB
+
 
 @with_logger
 class ServerContext:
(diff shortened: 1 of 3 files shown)


server/servercontext.py
```diff
@@ -17,6 +17,9 @@
 from .protocol import DisconnectedError, Protocol, QDataStreamProtocol
 from .types import Address
 
+MiB = 2 ** 20
+LIMIT = 10 * MiB
```
Askaholic, line 21: Any point in making this a config value? We couldn't change it without restarting the server anyway.

Sheikah45: I don't think so really. Only if this has some non minimal impact but I doubt it

Sheikah45 (approved):

server/servercontext.py
BlackYps: creating config values that need a restart anyway seems counterintuitive to me. Additionally I don't think there is much value in being able to easily change this value

BlackYps (approved):

PR #945, opened 2022-12-26
Configure server contexts through yaml config

Askaholic:
This enables us to set up the lobby ports using the yaml config file. It's not entirely dynamic in that changing the config will still require a server restart to take effect (mostly because it was going to be a little complicated to implement and I felt like it wasn't worth it). This will let us turn off the legacy QDataStream socket whenever we feel like it's time without needing to make a code change.

I also have some plans for a subsequent PR to add a 'v2' port that will be enabled in dev/test configs but not in prod, so we can start deploying some breaking protocol changes running side by side with the stable version to develop against.

### DIFF
--- a/server.py
+++ b/server.py
@@ -115,9 +115,9 @@ async def main():
     }
     for cfg in config.LISTEN:
         try:
-            host = cfg.get("address", "")
-            port = cfg.get("port", 0)
-            proto_class_name = cfg.get("protocol", QDataStreamProtocol.__name__)
+            host = cfg["ADDRESS"]
+            port = cfg["PORT"]
+            proto_class_name = cfg["PROTOCOL"]
             proto_class = PROTO_CLASSES[proto_class_name]
 
             await instance.listen(
--- a/server.py
+++ b/server.py
@@ -124,17 +124,13 @@ async def main():
                 address=(host, port),
                 protocol_class=proto_class
             )
-        except Exception:
-            logger.exception(
-                "Failed to start server instance with config: %s",
-                cfg
-            )
-            done.set_result(-1)
+        except Exception as e:
+            raise RuntimeError(f"Error with server instance config: {cfg}") from e
 
     if not instance.contexts:
-        logger.warning(
-            "The server was not configured to listen on any ports! No one will "
-            "be able to connect."
+        raise RuntimeError(
+            "The server was not configured to listen on any ports! Check the "
+            "config file and try again."
         )
 
     server.metrics.info.info({
--- a/server.py
+++ b/server.py
@@ -123,7 +123,7 @@ async def main():
         time.perf_counter() - startup_time
     )
 
-    await done
+    exit_code = await done
 
     shutdown_time = time.perf_counter()
 
@@ -133,6 +133,8 @@ async def main():
 
     # Close DB connections
     await database.close()
+
+    return exit_code
 
 
 if __name__ == "__main__":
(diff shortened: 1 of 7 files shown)


server.py
```diff
@@ -109,8 +109,33 @@ async def restart_control_server():
...
-    await instance.listen(("", 8002), SimpleJsonProtocol)
+    PROTO_CLASSES = {
+        QDataStreamProtocol.__name__: QDataStreamProtocol,
+        SimpleJsonProtocol.__name__: SimpleJsonProtocol
+    }
+    for cfg in config.LISTEN:
+        try:
+            host = cfg.get("address", "")
+            port = cfg.get("port", 0)
+            proto_class_name = cfg.get("protocol", QDataStreamProtocol.__name__)
```
Askaholic, line 120: Not sure if these are the best defaults. Using `0` for port makes it listen on a random open high port.

server/config.py
```diff
@@ -35,6 +35,18 @@ def __init__(self):
         Change default values here.
         """
         self.CONFIGURATION_REFRESH_TIME = 300
+        self.LISTEN = [
+            {
+                "address": "",
+                "port": 8001,
+                "protocol": "QDataStreamProtocol",
```
Askaholic, line 42: We can rename these keys to whatever. Maybe caps make sense to be consistent with the rest of the config?

server.py
Sheikah45: Why have defaults? Wouldn't it be better to have it fail fast with a misconfiguration?

server.py
```diff
@@ -109,8 +109,33 @@ async def restart_control_server():
...
+                "Failed to start server instance with config: %s",
+                cfg
+            )
+            done.set_result(-1)
+
+    if not instance.contexts:
+        logger.warning(
+            "The server was not configured to listen on any ports! No one will "
+            "be able to connect."
+        )
```
Sheikah45, line 138: If this is done on startup would it be better to just have it exit with an error so that we know when there are no ports set quickly rather than having to parse the logs

server/config.py
Sheikah45: Would these be set with dot notation in the yaml? It might be good to be consistent

Sheikah45 (changes requested):

server.py
Askaholic: We could do it that way as well

server/config.py
Askaholic: Not sure what you mean by 'dot notation'. In yaml it looks like this:
```yaml
LISTEN:
  - address: ""
    port: 8001
    protocol: QDataStreamProtocol

  - address: ""
    port: 8002
    protocol: SimpleJsonProtocol
```

server.py
Askaholic: Yea I guess maybe it makes more sense to just exit. There is some stuff that could be listening like the control server and the rabbitmq connection, but I'm not sure if there is actually a use case where you'd ever want it to run without allowing clients to connect. Then again, if this does get misconfigured it's not like it's going to go unnoticed since no one will be able to play...

server/config.py
Sheikah45: Yeah sorry meant to say they would be their own fields.

Then yeah looking at this I might say it would be good to be consistent.

Although I am sure others might say otherwise to indicate it is a sub field. But I think the indent solves that

server.py
Sheikah45: Yeah maybe but that requires people complaining ha. And then digging into logs which personally is more of a pain if the service still says it is running fine

Sheikah45 (approved):

PR #949, opened 2022-12-31
Issue/#946 Update achievements over RabbitMQ

### LINKED ISSUE #946: Achievements are broken again
Achievements have been broken for quite some time due to a change in how the API does authentication. The error currently looks like this:
```
ERROR    Dec 25  22:33:11 SessionManager                 404, message='Not Found', url=URL('https://api.test.faforever.com/oauth/token')
  File "/code/server/api/api_accessor.py", line 40, in get_session
    await self.session.fetch_token()
  File "/code/server/api/oauth_session.py", line 43, in fetch_token
    creds = await self._make_request(data=data)
  File "/code/server/api/oauth_session.py", line 69, in _make_request
    async with session.post(self.token_url, data=data) as resp:
  File "/usr/local/lib/python3.9/site-packages/aiohttp/client.py", line 1141, in __aenter__
    self._resp = await self._coro
  File "/usr/local/lib/python3.9/site-packages/aiohttp/client.py", line 643, in _request
    resp.raise_for_status()
  File "/usr/local/lib/python3.9/site-packages/aiohttp/client_reqrep.py", line 1005, in raise_for_status
    raise ClientResponseError(
aiohttp.client_exceptions.ClientResponseError: 404, message='Not Found', url=URL('https://api.test.faforever.com/oauth/token')
```

The short term solution would be to fix the authentication code to work with the new method, however, we should really be persisting the achievements in some more reliable manner. Sending them to the API via HTTP is just bound to keep breaking in the future.

Askaholic:
Reworks the achievements updating code to send messages to rabbitmq instead of trying to post them to the API over http. This allows us to get rid of a whole bunch of broken boilerplate code for handling API requests. The only thing that becomes more annoying is handling the responses as it needs to function even when the other side is temporarily down, so we need to handle responses asynchronously and can't assume they will come in within any particular time frame.

Currently, if the player disconnects before getting a response from the API, the response will just be ignored when it does eventually come in. We could make things more complex by adding logic to allow those message to come in even after re-logs. It would just require more memory and optionally also a hard time out after which we would ignore the responses.

Closes #946


Sheikah45:
What even are the messages the server sends to the client about achievements?

When we reworked the code for the messages in the client. I don't think any achievement handling was even implemented. I think it all just queries the api now.

Askaholic:
Idk. If it’s not used in the client I’d be happy to remove it from the server. I assume it would have been for showing a notification when you unlock an achievement.

Sheikah45:
If I knew the message command name I could check if we even would recognize it

Askaholic:
It’s called `updated_achievements`
https://github.com/FAForever/server/blob/dee639f4e271662b593f44f449169eb06a8d3ec1/server/lobbyconnection.py#L306

Sheikah45 (approved):
looks good to me. Although will this have to be timed with the api change?

Askaholic:
Nope. The nice thing about rabbitmq is that if nothing is listening on the other side, the messages just disappear into the void. We could preemptively configure a queue to capture the messages, and then once the API implementation is deployed, anything sitting in the queue will start to be processed. Up to @Brutus5000 really if he wants to do that.

Brutus5000:
I have no pain in deployment order because right now everything is broke anyway. However, before merging we should make sure that it is compatible (or can be made compatible) with api code.

Askaholic:
`playerId` is in there. See the integration test.

Brutus5000 (approved):
Tested successfully against API PR on test server

PR #951, opened 2023-01-07
More info logging

Askaholic:
Some changes that will make it easier to search the logs in grafana, and some changes to make it easier to correlate messages with the player that caused them. I also moved some messages into the info level that happen when people connect/disconnect. This means there will be a log message containing the IP address of everyone who connects:

```
INFO     Jan 07  13:13:13 ServerContext                  LobbyServer[SimpleJsonProtocol]: Client connected from 127.0.0.1:39214
```

### DIFF
--- a/server/gameconnection.py
+++ b/server/gameconnection.py
@@ -5,6 +5,7 @@ Game communication over GpgNet
 import asyncio
 import contextlib
 import json
+import logging
 from typing import Any
 
 from sqlalchemy import select
@@ -13,7 +14,6 @@ from server.db import FAFDatabase
 
 from .config import TRACE
 from .db.models import coop_leaderboard, coop_map, teamkills
-from .decorators import with_logger
 from .game_service import GameService
 from .games import (
     CoopGame,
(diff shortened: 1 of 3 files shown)


Sheikah45:
I don't think we want to be exposing people's IP addresses in logs that are more or less publicly viewable. I think that should probably be removed else we will have to be more secure about the logs in grafana for the server. 

@Brutus5000 what are your thoughts?

Askaholic:
BTW, IP addresses can already show up in the logs if the connection is aborted before the player logs in.

Sheikah45:
Then those should likely be removed as well if we don't want IPs in the logs

Brutus5000:
As long as the sensitive logs get rotated out eventually I have no issues with logging IPs. We also keep ips for other purposes such as fraud login detection.

Sheikah45:
Alright then wasn't sure since these logs are more widely available in grafana besides just server admins

Askaholic:
So is this good to be merged then?

Brutus5000:
I can't say anything from the python side of things, but sure, go ahead.

Sheikah45 (approved):

PR #952, opened 2023-01-24
Change map pool selection

BlackYps:
Changes the map pool selection from lowest rating to average rating as per request of the matchmaking team.
Also fixes a minor bug I noticed where the slot pairs are only made based on mean rating which can lead to unexpected results for players with high deviation.

I feel like it would be good to have unit tests for this behavior, but I am not sure how to access the relevant variables to check their contents after the method has run.


Blodir:
New to the codebase, but looks good to me!

PR #954, opened 2023-02-27
Improve matchmaker info message

BlackYps:
Removes the fake news about random matching as this doesn't happen in TMM.
The second messages seems redundant and is a bit confusing with the included percentage figure, so I removed it.


Askaholic (approved):
Once the docker build fix is in we can merge

PR #955, opened 2023-03-02
Fix docker build

Askaholic:
Seems like some change in setuptools broke our docker build, however, this really just exposed a bug in our build where the `.git` directory was not being copied to the docker context and therefore the git version tag couldn't be found from the `setup.py` file. I went ahead and rewrote the dockerfile to use a multi-stage approach so we should no longer have build tools like `git` and `pipenv` installed in the final container.

Closes #953


Brutus5000 (commented):
I think you need to create the user earlier (or use an id) and chown on copy. This is an example from a quarkus skeleton app:

```
FROM registry.access.redhat.com/ubi8/openjdk-17:1.14

ENV LANGUAGE='en_US:en'


# We make four distinct layers so if there are application changes the library layers can be re-used
COPY --chown=185 build/quarkus-app/lib/ /deployments/lib/
COPY --chown=185 build/quarkus-app/*.jar /deployments/
COPY --chown=185 build/quarkus-app/app/ /deployments/app/
COPY --chown=185 build/quarkus-app/quarkus/ /deployments/quarkus/

EXPOSE 8080
USER 185
ENV JAVA_OPTS="-Dquarkus.http.host=0.0.0.0 -Djava.util.logging.manager=org.jboss.logmanager.LogManager"
ENV JAVA_APP_JAR="/deployments/quarkus-run.jar"


```

Askaholic:
I don’t think so. The python files are just installed to the system anyway. The user isn’t needed at all during the build process, it just exists for the container to use at run time.

Brutus5000:
So "other" users have read/execute permissions on the code directory?

Askaholic:
Yes, the code is installed as a python package with pip. Just like if you were to system install any other dependency with pip.

Askaholic:
Tested on the test server and it still works with the existing configuration.

PR #964, opened 2023-05-14
Add support for the PROXY protocol

Askaholic:
Adds a new option to the `LISTEN` config which can enable the PROXY protocol for the given socket. This allows the lobby server to be placed behind a TCP proxy and still receive correct IP information for logging and GEOIP location. When a socket is placed into PROXY mode it should ONLY be accessed via the proxy. Direct traffic to the socket must be blocked by a firewall.

Proxy mode supports both v1 and v2 of the PROXY protocol which will be automatically detected from the connection headers.

### DIFF
--- a/Pipfile
+++ b/Pipfile
@@ -20,6 +20,7 @@ humanize = ">=2.6.0"
 maxminddb = "*"
 oauthlib = "*"
 prometheus_client = "*"
+proxy-protocol = "*"
 pyjwt = {version = ">=2.4.0", extras = ["crypto"]}
 pyyaml = "*"
 sortedcontainers = "*"
--- a/Pipfile.lock
+++ b/Pipfile.lock
@@ -1,7 +1,7 @@
 {
     "_meta": {
         "hash": {
-            "sha256": "b3da1b047f70a6659401f07cd87ba09dbec53472d498b333134e4e7e4db7d9f6"
+            "sha256": "816ffa2da529a42d808057d7505f783ebc0d3e5118df782527cd6e0611d3ae73"
         },
         "pipfile-spec": 6,
         "requires": {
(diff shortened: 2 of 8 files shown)


Sheikah45 (commented):
This looks mostly good to me.

The only thing I might partially be worried about is that this doesn't mix will with a separate websocket protocol since the proxying is defined at the context level rather than connection protocol level. But then again we haven;t started web sockets yet so...

Askaholic:
How would it interfere with websockets? I figured the websocket bridge would just connect to the lobby via the PROXY protocol and handle the population of the original IP fields. I have been able to connect to the lobby server via websockets using this proxy in the past: https://github.com/Askaholic/lobby-bridge/

Sheikah45:
I wasn't sure if websockets utilize the same proxy protocol or if they would even require a proxy in the same way since they could be protected by cloudflare. 

If they use the same protocol then great, although just becomes a question of if they need the proxy, but no technical limitations.

Sheikah45:
And it looks like your bridge is for converting the websocket to raw TCP where as I was thinking of having the websockets connect directly to the lobby server

Sheikah45:
Admittedly I am also not clear on why websockets requires a separate bridge but that is out of scope for this pr

Askaholic:
It's because the websocket library in python doesn't integrate well with our code. We need to be able to make fire-and-forget calls to the message sending interface for broadcasts, however the websocket library doesn't support that. They have their own `broadcast` function to work around that, but it wouldn't support the complex filtering that we do for game visibility. Plus I have big concerns about performance. IMO it's easier and safer to write a separate bridge application to handle websockets in Rust instead and just convert the websocket messages into our normal protocol that the lobby server can already handle reliably.

p4block:
My only comment is that, yes, we will firewall it, but the code is supposed to reject non-PROXY calls anyway (and it does!)

PR #978, opened 2023-10-15
Issue/#977 Remove IRC password functionality

### LINKED ISSUE #977: Anope database is going away with IRC changes
The IRC server is being replaced by Ergochat which has more modern login capabilities. It supports customizing the login process so clients will be able to log in with a token, and for the users that want to log into IRC outside of the FAF client, there will be a separate mechanism for setting a password which will be built outside of the lobby server. So as far as the lobby server is concerned, all knowledge/logic around IRC can be dropped.

https://github.com/FAForever/faf-stack/pull/225

Askaholic:
This will prevent the logs from being filled with warnings about failing to update NickServ password whenever someone logs in.

Closes #977

### DIFF
--- a/server/lobbyconnection.py
+++ b/server/lobbyconnection.py
@@ -5,19 +5,16 @@ Handles requests from connected clients
 import asyncio
 import contextlib
 import json
-import os
 import random
 import urllib.parse
 import urllib.request
-from binascii import hexlify
 from datetime import datetime
 from functools import wraps
-from hashlib import md5
 from typing import Optional
 
 import aiohttp
 from sqlalchemy import and_, func, select
-from sqlalchemy.exc import DBAPIError, OperationalError, ProgrammingError
+from sqlalchemy.exc import DBAPIError
 
 import server.metrics as metrics
 from server.db import FAFDatabase
@@ -511,14 +508,16 @@ class LobbyConnection:
 
             username = row.login
 
-        new_irc_password = hexlify(os.urandom(16)).decode()
+        # DEPRECATED: IRC passwords are handled outside of the lobby server.
+        # This message remains here for backwards compatibility, but the data
+        # sent is meaningless and can be ignored by clients.
         await self.send({
             "command": "irc_password",
-            "password": new_irc_password
+            "password": "deprecated"
         })
 
         await self.on_player_login(
-            player_id, username, new_irc_password, unique_id, auth_method
+            player_id, username, unique_id, auth_method
         )
 
     async def command_hello(self, message):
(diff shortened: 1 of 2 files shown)


server/lobbyconnection.py
```diff
@@ -511,14 +511,16 @@ async def command_auth(self, message):
...
 
-        new_irc_password = hexlify(os.urandom(16)).decode()
+        # DEPRECATED: IRC passwords are handled outside of the lobby server.
+        # This message remains here for backwards compatibility, but the data
+        # sent is meaningless and can be ignored by clients.
         await self.send({
             "command": "irc_password",
-            "password": new_irc_password
+            "password": hexlify(os.urandom(16)).decode()
         })
```
Sheikah45, line 520: Do we even need to send the message?

I know removing it from the spec would be a backwards breaking change but it seems like just not sending it should be fine.

Sheikah45 (approved):

server/lobbyconnection.py
Askaholic: Mmm, according to my notes about version numbers I think not sending it would be a breaking change. I would consider it part of the `command_auth` message flow where if you send a valid token then you always expect the next message that comes back to you to be `irc_password`, so removing that would be a breaking change to any code that was built on that message flow. This is also supported by my integration tests which fail if I remove that command.

https://github.com/FAForever/server/blob/develop/CONTRIBUTING.md#version-numbers

Askaholic: Maybe there don't exist any client implementations that rely on that in practice, but IDK. I feel like it doesn't really hurt much to keep it and it ensures that the protocol definitely remains backwards compatible.

Brutus5000: In general yes, but the websocket change is a breaking change and we'll force all users to update to new irc+websocket for lobby

Askaholic: Sure, but that's just a change to the message transport. As far as like the logic of how to handle certain messages that wouldn't be affected by the websocket change.

Sheikah45: Ah didn't realize it was considered part of the auth flow. I just considered it as a message that comes at some point after auth

Askaholic: It's certainly a bit subjective what the 'flow' really is, but since the message is sent unconditionally after validating the token I would consider it to be part of the 'flow'.

Brutus5000: Wouldn't it be more expressive for future us to set it to a fixed string "deprecated" instead of using random bits?

Askaholic: Sure, I can get behind that

PR #981, opened 2023-11-04
Remove all Twilio references

Brutus5000:
FAF nevers used the Twilio service + it has been discontinued


Pipfile.lock
```diff
@@ -827,14 +827,6 @@
             "index": "pypi",
             "version": "==0.4.5"
         },
-        "twilio": {
-            "hashes": [
```
Askaholic, line 831: You need to use pipenv to uninstall the package so that the lock file is updated correctly. Just deleting twilio from here is not correct because there are sub-dependencies that are no longer needed and there is also a meta section in the lockfile that contains a checksum that also needs to be updated.

You either need to run `pipenv uninstall twilio` which will remove twilio and its dependencies, but will also upgrade all other dependencies. This will be the easiest.

Or, run `pipenv graph` to figure out what twilios dependencies are:
```
twilio==8.10.0
  - aiohttp [required: >=3.8.4, installed: 3.8.6]
  - aiohttp-retry [required: >=2.8.3, installed: 2.8.3]
  - PyJWT [required: >=2.0.0,<3.0.0, installed: 2.8.0]
  - requests [required: >=2.0.0, installed: 2.31.0]
```
And then uninstall the ones that are not not required by twilio and are not otherwise listed in the Pipfile: `pipenv uninstall --keep-outdated twilio aiohttp-retry requests`. Even when I do this method it still seems to be upgrading `charset-normalizer` so I'd probably just go with method 1.

Askaholic (changes requested):

Pipfile.lock
Askaholic: I actually just decided to commit what I had since I checked out the branch anyway.

Brutus5000: Good to know. I'm not deep into the Python packaging.

Askaholic (approved):

PR #982, opened 2023-11-04
Remove Coturn brokering

Brutus5000:
This featured has been moved to a dedicated API.


integration_tests/test_server.py
```diff
@@ -6,20 +6,6 @@ async def test_ping(client_factory):
...
-
-    await client.send_command("ice_servers")
-    msg = await client.read_until_command("ice_servers")
-
-    assert msg["ttl"] > 60
-    assert msg["ice_servers"]
-    for server in msg["ice_servers"]:
-        assert server["urls"]
-        assert server["username"]
-        assert server["credential"]
```
Askaholic, line 20: I think we probably want to keep this actually, but adding a `# DEPRECTATED:` comment would be good so it doesn't get forgotten when the command is removed.

Askaholic (commented):
I don't think the Pipfile.lock actually needs to be updated. I think that happened because of the changes you made when you removed twilio. I can fix it up for you though.

integration_tests/test_server.py
Brutus5000: There still is `test_ice_servers_empty`. Isn't that sufficient? (since it will always be empty)

Askaholic: Ah, in that case, that's fine

Askaholic (approved):

PR #998, opened 2024-01-06
Get default config values from environment

Askaholic:
@Brutus5000 requested this feature to make managing secrets easier in kubernetes. This way values like the database password which are never going to change dynamically anyway can be stored in a `Secret` object and mapped to the environment while non-sensitive data can be stored in a `ConfigMap` and mapped to a volume that will be able to trigger the dynamic changes. Since kubernetes doesn't really support merging those two things into a single file, this feature will make setting up the configuration a lot easier and more idiomatic.

NOTE: This will only work for values which are expected to be strings! This should not be a problem since this feature is only intended to be used for passwords.

### DIFF
--- a/server/config.py
+++ b/server/config.py
@@ -156,7 +156,9 @@ class ConfigurationStore:
         self.QUEUE_POP_TIME_MOVING_AVG_SIZE = 5
 
         self._defaults = {
-            key: value for key, value in vars(self).items() if key.isupper()
+            key: value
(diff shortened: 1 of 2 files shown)


Brutus5000:
Is refresh guaranteed to be called on startup?

Askaholic:
Yep, it’s called in the constructor

Brutus5000 (approved):

PR #1004, opened 2024-02-03
Perform client time sync on login

### LINKED ISSUE #889: Perform client time sync on login
We would like to send absolute times (ISO timestamps) instead of timedeltas, however, it turns out that a significant number of people have their system time configured incorrectly for whatever reason, which makes the computed deltas show up incorrectly on the client. To mitigate this we should send the current time to the client on login. The client can then compute the difference between this timestamp and the system time and adjust all future timestamps sent by the server using this offset. This should even help to adjust for client latency to the server assuming that the latency is relatively stable.

- Add `current_time` to the command `welcome` sent to the client on login.

baterflyrity:
Add current_time to the command welcome sent to the client on login.

Closes #889

### DIFF
--- a/server/lobbyconnection.py
+++ b/server/lobbyconnection.py
@@ -18,6 +18,7 @@ from sqlalchemy.exc import DBAPIError
 
 import server.metrics as metrics
 from server.db import FAFDatabase
+from server.timing import datetime_now
 
 from .config import TRACE, config
 from .db.models import (
--- a/server/lobbyconnection.py
+++ b/server/lobbyconnection.py
@@ -693,7 +693,7 @@ class LobbyConnection:
         await self.send({
             "command": "welcome",
             "me": self.player.to_dict(),
-            "current_time": datetime.utcnow().timestamp(),
+            "current_time": datetime_now().isoformat(),
 
             # For backwards compatibility for old clients. For now.
             "id": self.player.id,
--- a/server/lobbyconnection.py
+++ b/server/lobbyconnection.py
@@ -693,6 +693,7 @@ class LobbyConnection:
         await self.send({
             "command": "welcome",
             "me": self.player.to_dict(),
+            "current_time": datetime.utcnow().timestamp(),
 
             # For backwards compatibility for old clients. For now.
             "id": self.player.id,
--- a/tests/integration_tests/test_login.py
+++ b/tests/integration_tests/test_login.py
@@ -1,4 +1,3 @@
-from datetime import datetime
 from time import time
 
 import jwt
@@ -121,7 +120,7 @@ async def test_server_login_valid(lobby_server):
         assert msg == {
             "command": "welcome",
             "me": me,
-            "current_time": '1970-01-01T00:00:00+00:00',
+            "current_time": "1970-01-01T00:00:00+00:00",
             "id": 3,
             "login": "Rhiza"
         }
@@ -168,7 +167,7 @@ async def test_server_login_valid_admin(lobby_server):
         assert msg == {
             "command": "welcome",
             "me": me,
-            "current_time": '1970-01-01T00:00:00+00:00',
+            "current_time": "1970-01-01T00:00:00+00:00",
             "id": 1,
             "login": "test"
         }
@@ -214,7 +213,7 @@ async def test_server_login_valid_moderator(lobby_server):
         assert msg == {
             "command": "welcome",
             "me": me,
-            "current_time": '1970-01-01T00:00:00+00:00',
+            "current_time": "1970-01-01T00:00:00+00:00",
             "id": 20,
             "login": "moderator"
         }
@@ -314,7 +313,7 @@ async def test_server_login_token_valid(lobby_server, jwk_priv_key, jwk_kid):
         assert msg == {
             "command": "welcome",
             "me": me,
-            "current_time": '1970-01-01T00:00:00+00:00',
+            "current_time": "1970-01-01T00:00:00+00:00",
             "id": 3,
             "login": "Rhiza"
         }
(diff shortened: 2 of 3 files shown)


baterflyrity:
One failed test (_tests/integration_tests/test_matchmaker.py→test_game_matchmaking_timeout[qstream]: asyncio.exceptions.TimeoutError_) is irrelevant to this PR, so search someone fix for that in other issues.

server/lobbyconnection.py
```diff
@@ -693,6 +693,7 @@ async def on_player_login(
         await self.send({
             "command": "welcome",
             "me": self.player.to_dict(),
+            "current_time": datetime.utcnow().timestamp(),
```
Askaholic: There's a helper function in `server.timing` called `datetime_now` that we use for getting the current time as a timezone aware object. Using that also makes it easier to mock the current time in unit tests.

We also want to be using ISO formated strings rather than timestamps.

```suggestion
            "current_time": datetime_now().isoformat(),
```

tests/integration_tests/test_login.py
```diff
@@ -118,9 +121,12 @@ async def test_server_login_valid(lobby_server):
     assert msg == {
         "command": "welcome",
         "me": me,
+        "current_time": msg.get("current_time", None),
```
Askaholic: I think it's probably cleaner to mock the `datetime_now` call to a fixed time and then include that in the assertion here.

tests/integration_tests/test_login.py
```diff
@@ -118,9 +121,12 @@ async def test_server_login_valid(lobby_server):
...
         "id": 3,
         "login": "Rhiza"
     }
+    assert isinstance(msg.get("current_time", None), float)
+    assert abs(msg.get("current_time", -1) - datetime.utcnow().timestamp()) <= TIMEOUT
```
Askaholic: Then you don't need this

Askaholic (commented):
Thanks for the contribution! The test failure was just a flaky test, there are still a few that can happen from time to time. I reran the tests which should probably take care of that.

server/lobbyconnection.py
baterflyrity: Sure. There are several `now = datetime.utcnow()` [lines](https://github.com/baterflyrity/server/blob/eea6c98ae1b420046a25bb78d8a424c5e083af5a/server/lobbyconnection.py#L458) hence used that style.

Thus maybe it would be nice to migrate these UTCs to *server.timing* too.

baterflyrity:
Also there are several `now = datetime.utcnow()` [lines](https://github.com/baterflyrity/server/blob/eea6c98ae1b420046a25bb78d8a424c5e083af5a/server/lobbyconnection.py#L458) thus one would like to migrate them to *server.timing* too.

Askaholic:
> Also there are several `now = datetime.utcnow()` [lines](https://github.com/baterflyrity/server/blob/eea6c98ae1b420046a25bb78d8a424c5e083af5a/server/lobbyconnection.py#L458) thus one would like to migrate them to _server.timing_ too.

Ah, yea you should leave those. They actually are a bit of a special case because they are compared to datetime objects created automatically by sqlalchemy which are always timezone **un**aware. The `datetime_now` helper returns a timezone aware datetime object so it would not be able to be compared against the ban expiration time returned by sqlalchemy.

baterflyrity:
Nice, all checks passed finally!

@Askaholic ure right, mocked _server.timing_ plus added context manager helper for further cases.

Can this be merged now?

p.s.
Consider moving all datetime operation to _server.timing_ because utc (un)awareness is not obvious within the code.

baterflyrity:
@Askaholic , hey! Hope ure going well. What is the status of this PR?

Askaholic:
Sorry, I haven’t had time to take care of this yet, my weekends have been pretty busy recently. 

It looks great though! My only suggestion would be if you were up for it, to rework the `fixed_time` context manager to be a fixture that works much like the `monkeypatch` fixture. I think that would be slightly more ergonomic to work with. However, I’m also happy to merge as is, I just need to find some time on the weekend to do a final review and testing.

baterflyrity:
Sure, just checking everything is fine.

> It looks great though! My only suggestion would be if you were up for it, to rework the `fixed_time` context manager to be a fixture that works much like the `monkeypatch` fixture. I think that would be slightly more ergonomic to work with. However, I’m also happy to merge as is, I just need to find some time on the weekend to do a final review and testing.

Well, that is not quite true because there can be timing-relying tests (like check timeout/latency/performance of net io) hence fixture with **hard fix all timestamps** will fail. I have several ideas:

- add fixture alongside with context manager
- yield different easy-to-calculate timings e.g. 1, 2, 3
- make [fixture with  parameters](https://docs.pytest.org/en/7.1.x/how-to/fixtures.html#using-markers-to-pass-data-to-fixtures)

What do you think about these?

Askaholic:
Is that an issue in any of these tests though? I think generally for timing related things internally we use either `time.time()` or `loop.time()` which wouldn't be affected by this fixture. I think the `datetime_now` function is mostly just used when we need to return an iso timestamp to the client, and there are a few places where it calculates long term expirations like the ban expiration.

I'm thinking for scenarios like testing the ban expiration it would actually be really nice to be able to do something like:

```python
def test_ban_expired(fixed_time):
    fixed_time.set(100)
    
    # Insert 30 minute ban into database
    ...

    fixed_time.set(100 + 30)  # Whatever the units would be
    
    # Assert that ban has expired
    ...
    
```

baterflyrity:
Np, if you assume no sequence timings are considered in tests.

Now your example would be:
```python
def test_ban_expired(fixed_time):
    assert server.lobbyconnection.datetime_now().timestamp == 0.
    fixed_time(100)
    assert server.lobbyconnection.datetime_now().timestamp == 100.
    
    # Insert 30 minute ban into database
    ...

    fixed_time(100 + 30)  # Whatever the units would be
    assert server.lobbyconnection.datetime_now().timestamp == 130.
    
    # Assert that ban has expired
    ...
    
```

Askaholic (approved):

PR #1009, opened 2024-03-06
Issue #800 - Invalid Command Response on Database Outage

### LINKED ISSUE #800: Command invalid when db is down
When the database is down and someone tries to login the error message from the server for why it fails claims that the command is invalid rather than saying the server can't reach the db.

It would probably be good to change this just for debugging purposes

benjamin-lawson:
Closes #800.

### DIFF
--- a/server/lobbyconnection.py
+++ b/server/lobbyconnection.py
@@ -224,7 +224,7 @@ class LobbyConnection:
                 "text": "Unable to connect to database. Please try again later."
             })
             # Make sure to abort here to avoid a thundering herd problem.
-            await self.abort('Error connecting to database')
+            await self.abort("Error connecting to database")
         except Exception as e:  # pragma: no cover
             await self.send({"command": "invalid"})
             self._logger.exception(e)
--- a/server/lobbyconnection.py
+++ b/server/lobbyconnection.py
@@ -216,9 +216,9 @@ class LobbyConnection:
                 self.get_user_identifier(),
                 cmd
             )
-        except OperationalError as e:
+        except OperationalError:
             # When the database goes down, SqlAlchemy will throw an OperationalError
-            self._logger.error(f'Encountered OperationalError on message received. This could indicate DB is down.')
+            self._logger.error(f"Encountered OperationalError on message received. This could indicate DB is down.")
             await self.send({
                 "command": "notice",
                 "style": "error",
--- a/server/lobbyconnection.py
+++ b/server/lobbyconnection.py
@@ -216,8 +216,9 @@ class LobbyConnection:
                 self.get_user_identifier(),
                 cmd
             )
-        except OperationalError:
+        except OperationalError as e:
             # When the database goes down, SqlAlchemy will throw an OperationalError
+            self._logger.error(f'Encountered OperationalError on message received. This could indicate DB is down.')
             await self.send({
                 "command": "notice",
                 "style": "error",
--- a/server/lobbyconnection.py
+++ b/server/lobbyconnection.py
@@ -218,7 +218,7 @@ class LobbyConnection:
             )
         except OperationalError:
             # When the database goes down, SqlAlchemy will throw an OperationalError
-            self._logger.error(f"Encountered OperationalError on message received. This could indicate DB is down.")
+            self._logger.error("Encountered OperationalError on message received. This could indicate DB is down.")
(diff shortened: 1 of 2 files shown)


Askaholic:
Awesome, thanks for the PR! Sorry if it takes me a little while to get to reviewing it properly, but at first glance it looks great!

benjamin-lawson:
> Awesome, thanks for the PR! Sorry if it takes me a little while to get to reviewing it properly, but at first glance it looks great!

No worries at all! Just wanted to give you a heads up. I'll try to look at some of the other open issues.

server/lobbyconnection.py
```diff
@@ -216,6 +216,15 @@ async def on_message_received(self, message):
                 self.get_user_identifier(),
                 cmd
             )
+        except OperationalError:
+            # When the database goes down, SqlAlchemy will throw an OperationalError
```
Askaholic, line 220: Would we still want to log the exception here?

server/lobbyconnection.py
```diff
@@ -216,6 +216,15 @@ async def on_message_received(self, message):
...
+            await self.send({
+                "command": "notice",
+                "style": "error",
+                "text": "Unable to connect to database. Please try again later."
+            })
+            # Make sure to abort here to avoid a thundering herd problem.
```
Askaholic, line 227: I think we need to use the `kick` style to prevent the client from trying to automatically reconnect. @Sheikah45?

Askaholic (commented):
Hey, thanks for you patience. Sorry it took me a while to get to this. I finally replaced the battery on my laptop so now I can use it without having to be plugged in all the time which makes me significantly more likely actually open it up haha.

server/lobbyconnection.py
benjamin-lawson: I can go ahead and add logging here. I was avoiding it only because if the DB is truly down, it could cause a lot of logs to the logging service to be ingested due to failed commands. But if we feel like the logging service can handle that volume without issue, I have no qualms about adding it in.

server/lobbyconnection.py
benjamin-lawson: I was thinking that's what https://github.com/FAForever/server/pull/1009/files#diff-de94292bdcec0d8f98b8eb76040f97830866bed0c1fb5b55165a9e54f175ea0bR227 would do for the client, but if that is not the case, we for sure need to send the client a "Stop and don't retry" message.

server/lobbyconnection.py
Askaholic: Hmm, good question. I'm not sure it's every happened before, but I think it's safer to have it log something even if it ends up spamming a lot of messages than to have it error silently and require users to start complaining before we can see that there's an issue.

server/lobbyconnection.py
Askaholic: Yea, there's currently no good system for telling the client that it should not attempt to reconnect. I believe the `kick` style message is all we have. It might be cleaner to add a dedicated message for that but that would be for another PR and also require client changes.

server/lobbyconnection.py
benjamin-lawson: Too easy. I've added that logging in the most recent version.

server/lobbyconnection.py
benjamin-lawson: I think https://github.com/FAForever/downlords-faf-client/blob/ca8554d8767e748498dee37caa39cf3f657284a6/src/main/java/com/faforever/client/remote/FafServerAccessor.java#L115 could be an issue since the logic is if we were connected and suddenly were disconnected, retry the login and connect. Net new connections should be okay since it will be going from connecting to disconnected.

Sheikah45: The main problem with the kick style is that it will also close the client. If this is desired behavior then sure but otherwise it shouldn't be kick.

benjamin-lawson: @Sheikah45 @Askaholic I think this should work for what we want then. This follows the same pattern as if someone were banned where it sends a message and aborts the connection.

Askaholic (approved):

server/lobbyconnection.py
Askaholic: Sounds good, we'll leave it like this then.

PR #1014, opened 2024-04-20
Issue/#970 Prevent queue pop timer from becoming negative

### LINKED ISSUE #970: TMM queue pop timer can march to negative infinity
It looks like there is an issue where the pop timer can get into some kind of loop that marches down to 0 then negative infinity and then can never recover.

Server graph attached
image

Not sure from the logs what caused it but occurred around 17:07 US east on July 27th

Askaholic:
This adds a quick fix to prevent the timer from going negative by adding a minimum pop time to the config as well. Ideally the queue rate calculation should be fixed as well in a separate PR.

Closes #970

### DIFF
--- a/server/config.py
+++ b/server/config.py
@@ -148,6 +148,8 @@ class ConfigurationStore:
         self.MAP_POOL_RATING_SELECTION = "mean"
         # The maximum amount of time in seconds to wait between pops
         self.QUEUE_POP_TIME_MAX = 90
+        # The minimum amount of time in seconds to wait between pops
+        self.QUEUE_POP_TIME_MIN = 15
         # The number of possible matches we would like to have when the queue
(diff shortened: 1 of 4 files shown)


Sheikah45 (approved):

BlackYps:
Is five seconds enough to allow recovery of the timer or will it stay stuck on the five seconds?

Askaholic:
Good question. I think it depends on how many players are in the queue. If the number of players in the queue stays the same or increases then the timer will continue to be stuck at the minimum. I’m not sure what that number has to be though.

BlackYps:
15 seconds sounds like a reasonable minimum to me, even if it stays stuck at this for some time.
5 seconds is very low

Askaholic:
Fine with me

PR #1021, opened 2024-06-17
Add Basic Bug Report Issue Template

MrRowey:


Askaholic:
Cool, I’ll have to look into what those config options do in more detail later. Just glancing at it I feel like there are maybe too many required sections in the bug template. People rarely ever open issues on this repo so I don’t think we need a template that is quite so aggressive about requesting so much detail. Sometimes you notice a bug and you don’t know how to reproduce it, that’s the nature of a server side application. It’s still useful to report what happened.

MrRowey:
> Cool, I’ll have to look into what those config options do in more detail later. Just glancing at it I feel like there are maybe too many required sections in the bug template. People rarely ever open issues on this repo so I don’t think we need a template that is quite so aggressive about requesting so much detail. Sometimes you notice a bug and you don’t know how to reproduce it, that’s the nature of a server side application. It’s still useful to report what happened.

That's fine it's as simple as just setting the validation for that part to false.

PR #1026, opened 2024-09-06
Fix issue where players added or removed from friend or foe lists did not have games updated

### LINKED ISSUE #1002: Newly foed/friended players don't receive necessary game_info messages
## Scenario 1 - Adding a foe
Imagine you are hosting a game, a player joins the lobby who is just trolling and you don't want them in the game so you kick them. But they keep joining back into the lobby, so you add them as a foe to keep them out of the lobby. The server will now consider the game to be 'invisible' to that player and will prevent them from joining. But because the game is 'invisible' to them, they will stop receiving updates about the game including the updates that say the game has launched or ended and should be removed from the games view. The lobby will stick around on their client until they re-log.

## Scenario 2 - Adding a friend
Similarly, imagine you hosted a game for 'friends only'. Now you want to let someone join who wasn't on your friends list yet, so you add them as a friend. The server now considers the game to be visible to them, however it will not actually send them the `game_info` message needed to display the game in their client until an update is naturally triggered via another player joining/leaving or the game settings being changed.

## Log messages
This can cause the following log message to appear:
```
Client error: You cannot join games hosted by this player.
```

### Scenario 1
1. Player 1 host a game
2. Player 1 add player 2 as a foe

### Scenario 2
1. Player 1 host a game with visibility set to 'friends'
2. Player 1 add player 2 as a friend

## Solution
The `social_add` command should check if the current player is hosting a game, and if so immediately send a `game_info` message to the added player as follows:

If the player was foed and the game was previously visible to the player:
* Send a 'fake' `game_info` message with `"state": "closed"` to the foed player immediately.
Otherwise, if the player was friended and the game was previously not visible to the player:
* Send a `game_info` message to the friended player immediately.

benjamin-lawson:
**Changes**
- Add `send_game_list_to_player()` that will send the game's they have access to a specific player rather than just the logged in user.
- When social add or remove is called, it will send an updated game list to the target of the social add or remove.

**Potential Issues**
- Lots of social adds and removes has the potential for abuse of resources (DDOS). This would really be apparent if there are lots of games as well.
- The client might not accept the game list updates.

**TODO**

- Need to add unit/integration tests around this functionality.

Closes #1002

### DIFF
--- a/.pre-commit-config.yaml
+++ b/.pre-commit-config.yaml
@@ -1,19 +1,19 @@
 repos:
 -   repo: https://github.com/pre-commit/pre-commit-hooks
-    rev: v4.0.1
+    rev: v4.6.0
     hooks:
     -   id: trailing-whitespace
     -   id: end-of-file-fixer
     -   id: check-docstring-first
 -   repo: https://github.com/PyCQA/flake8
-    rev: 3.9.2
+    rev: 7.1.1
     hooks:
     -   id: flake8
 -   repo: https://github.com/pre-commit/mirrors-autopep8
-    rev: v1.5.7
+    rev: v2.0.4
     hooks:
     -   id: autopep8
 -   repo: https://github.com/PyCQA/isort
-    rev: 5.9.3
+    rev: 5.13.2
     hooks:
     -   id: isort
--- a/server/lobbyconnection.py
+++ b/server/lobbyconnection.py
@@ -316,15 +316,6 @@ class LobbyConnection:
             ]
         })
 
-    async def send_game_list_to_player(self, player: Player):
-        await player.send_message({
-            "command": "game_info",
-            "games": [
-                game.to_dict() for game in self.game_service.open_games
-                if game.is_visible_to_player(player)
-            ]
-        })
-
     async def command_social_remove(self, message):
         if "friend" in message:
             subject_id = message["friend"]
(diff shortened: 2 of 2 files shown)


Askaholic:
Hey thanks for the PR! I added the issue link to the description so it shows up in GitHub nicely.

I think you can avoid much of the DOS concerns by sending only the 1 game_info message that actually would change instead of the entire game list. Since the visibility will only change for the game that the player doing the social add/remove is currently hosting. Hopefully the way I said that makes sense.

Thanks for taking this on! I have so little time for FAF development anymore but more than happy to give reviews and answer any questions you may have about the code!

benjamin-lawson:
> Hey thanks for the PR! I added the issue link to the description so it shows up in GitHub nicely.
> 
> I think you can avoid much of the DOS concerns by sending only the 1 game_info message that actually would change instead of the entire game list. Since the visibility will only change for the game that the player doing the social add/remove is currently hosting. Hopefully the way I said that makes sense.

Okay! So if I send the client a game_info message with just the single game that would be affected by the friend or foe change, the client would update only that single game? I would think it would remove all the other games in that case.

> Thanks for taking this on! I have so little time for FAF development anymore but more than happy to give reviews and answer any questions you may have about the code!

No worries at all! Trying to knock out some of the low hanging fruit on the server side to understand the codebase better.

Askaholic:
So the way the protocol is designed is that the client only removes a game from the game list when it sees a game_info message with `"state": "closed"`. That way we can send updates on a per game basis without having to send the entire game list. The main code that handles this is in the broadcast service: https://github.com/FAForever/server/blob/645d001c08228f8f30f031f30f44a8eca624e9a5/server/broadcast_service.py#L89

You can see there that only the “dirty” (i.e. games that had a state change that hasn’t been sent to the connected clients yet) games are added to the list. Any games that haven’t seen any state changes are ignored.

Hope that helps!

PR #1027, opened 2024-10-26
Improve documentation of game messages

Garanas:
Related to: https://github.com/FAForever/fa/pull/6479

Improves the documentation of some of the game messages.

### DIFF
--- a/server/gameconnection.py
+++ b/server/gameconnection.py
@@ -509,8 +509,11 @@ class GameConnection(GpgNetServerProtocol):
 
     async def handle_launch_status(self, status: str):
         """
-        Currently is sent with status `Rejected` if a matchmaker game failed
-        to start due to players using differing game settings.
+        Represents the launch status of a peer.
+
+        # Params
+        - `status`: One of "Unknown" | "Connecting" | "Missing local peers" |
+            "Rejoining" | "Ready" | "Ejected" | "Rejected" | "Failed"
         """
         pass
 
@@ -518,6 +521,8 @@ class GameConnection(GpgNetServerProtocol):
         """
         Not sure what this command means. This is currently unused but
         included for documentation purposes.
+
+        Example args: ["ack", "23381", "232191", "64218.4"]
         """
         pass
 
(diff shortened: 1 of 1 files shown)


server/gameconnection.py
```diff
@@ -547,6 +571,19 @@ async def handle_game_full(self):
         """
         pass
 
+    async def handle_established_peers(self, peerId: str, peerConnectedTo: str):
```
Askaholic: Params should be `snake_case`

server/gameconnection.py
```diff
@@ -547,6 +571,19 @@ async def handle_game_full(self):
...
+        """
+        Send by the lobby when a peer connected to another peer. Message can be send multiple times, by multiple connections.
+
+        # Params
+        - peerId: The identifier of the peer that this connection received the message from
```
Askaholic: Add backticks so it matches the docstring for teamkill_happened. These get rendered as Markdown in our docs.
```suggestion
        - `peerId`: The identifier of the peer that this connection received the message from
```

server/gameconnection.py
```diff
@@ -547,6 +571,19 @@ async def handle_game_full(self):
...
+        - peerConneectedTo: The list of identifiers that the given peer established a connection to. Is a list stored in a string, separated by spaces. As an example: "1321 22 33 43221 2132"
+        """
+
+        # peerConnectedToList = peerConnectedTo.split(" ")
```
Askaholic: You can remove the commented out code. I think I’ll be able to figure that one out :)
```suggestion
```

Askaholic (commented):
Cool, thanks! I hope to find some time this weekend to implement the established peers.

What are we expecting the server to do with the launch status information? If we’re just ignoring most of it you could also just add the possible options in the docstring and get rid of the empty if chain.

Garanas:
The launch status can help you understand what is happening to a player/connection. For example - a player that has the status `Ready` is connected to all its peers and is ready to launch. Such a player should never receive a (time) penalty if the lobby does not start for some reason.

edit: perhaps 'launch status' is the wrong term to use here.

Askaholic:
I'm just going to combine this with #1028 for simplicity. Thanks for making the initial changes!

Issue #1032, opened 2024-12-06
Working veto system


K-ETFreeman:
Need to make working veto system:
- get matchmaker_queue_map_pool's veto system data from the database
- accept setting vetoes by the client
- take them into account when creating a match in matchmaker

K-ETFreeman:
Done in #1071

PR #1033, opened 2024-12-06
Feature/#1032 working veto system

K-ETFreeman:
Working on #1032
I need some feedback, what is right, what is wrong, what needs to be changed, etc

## Summary by CodeRabbit

* **New Features**
  * Players can submit per-pool map vetoes from the lobby; matchmaking consumes those vetoes to influence initial map weights.

* **Improvements**
  * Anti-repetition weight adjustments refined for more varied map selection.
  * Map/version metadata and per-pool veto configuration (tokens, limits, minima) now propagate through matchmaking and game launch.

* **Tests**
  * Extensive unit and integration tests added covering veto flows, dynamic token logic, and selection outcomes.

### DIFF
--- a/server/ladder_service/veto_system.py
+++ b/server/ladder_service/veto_system.py
@@ -116,7 +116,7 @@ class VetoService(Service):
                         map_pool_map_version_ids=[
                             map.map_pool_map_version_id
                             for map in matchmaker_queue_map_pool.map_pool.maps.values()
-                        ] + [-1],
+                        ],
                         veto_tokens_per_player=veto_tokens_per_player,
                         max_tokens_per_map=max_tokens_per_map,
                         minimum_maps_after_veto=minimum_maps_after_veto
--- a/server/ladder_service/veto_system.py
+++ b/server/ladder_service/veto_system.py
@@ -260,9 +260,13 @@ def _is_valid_veto_config_for_queue(
 ) -> bool:
     num_maps = len(queue_config.map_pool.maps)
 
+    if (queue_config.minimum_maps_after_veto > num_maps):
+        return False
+
     if (
         queue_config.max_tokens_per_map == 0
-        and queue_config.minimum_maps_after_veto >= num_maps
+        and queue_config.minimum_maps_after_veto == num_maps
+        and queue_config.veto_tokens_per_player > 0
     ):
         return False
 
@@ -270,11 +274,6 @@ def _is_valid_veto_config_for_queue(
         total_players = queue.team_size * 2
         vetoable_maps_per_player = queue_config.veto_tokens_per_player / queue_config.max_tokens_per_map
         vetoable_maps = total_players * vetoable_maps_per_player
-
-        # tokens/map > number of maps that may be vetoed
-        # TODO: because vetoable_maps >= 0 this inequality also implies
-        # queue_config.minimum_maps_after_veto > num_maps
-
         if vetoable_maps > num_maps - queue_config.minimum_maps_after_veto:
             return False
 
--- a/server/ladder_service/veto_system.py
+++ b/server/ladder_service/veto_system.py
@@ -274,7 +274,7 @@ def _is_valid_veto_config_for_queue(
         # tokens/map > number of maps that may be vetoed
         # TODO: because vetoable_maps >= 0 this inequality also implies
         # queue_config.minimum_maps_after_veto > num_maps
-        # Why is it strictly greater than here but greater than or equal to above?
+
         if vetoable_maps > num_maps - queue_config.minimum_maps_after_veto:
             return False
 
--- a/server/matchmaker/map_pool.py
+++ b/server/matchmaker/map_pool.py
@@ -105,7 +105,7 @@ class MapPool(object):
         self._logger.debug("______map_list________________: %s", map_list)
         final_weights = [adjusted_weights.get(mp_mv_id, 0) * m.weight for mp_mv_id, m in map_list]
         self._logger.debug("______final_weights________________: %s", final_weights)
-        return random.choices([map for _, map in map_list], weights=final_weights, k=1)[0].get_map()
+        return random.choices([map for _, map in map_list], weights=final_weights, k=1)[0].get_map()  # nosec B311
 
     def __repr__(self) -> str:
         return f"MapPool({self.id}, {self.name}, {list(self.maps.values())})"
--- a/server/matchmaker/map_pool.py
+++ b/server/matchmaker/map_pool.py
@@ -38,11 +38,16 @@ class MapPool(object):
         adjusted_weights = notzero_weights.copy()
 
         def get_notrepeated_weight_transfer_targets(current_weight, rep_count):
-            for base_threshold in base_thresholds:
+            thresholds = list(base_thresholds)
+            factor = repeat_factor ** (rep_count - 1)
+            if factor < 1:
+                thresholds.extend(t * factor for t in base_thresholds if t * factor < base_thresholds[-1])
+
+            for threshold in thresholds:
                 targets = [
                     target_id for target_id in notzero_weights
                     if repetition_counts.get(target_id, 0) == 0 and
-                    notzero_weights[target_id] >= base_threshold * repeat_factor ** (rep_count - 1) * current_weight
+                    notzero_weights[target_id] >= threshold * current_weight
                 ]
                 if targets:
                     return targets
--- a/server/matchmaker/matchmaker_queue.py
+++ b/server/matchmaker/matchmaker_queue.py
@@ -66,7 +66,7 @@ class MatchmakerQueue:
         self.team_size = team_size
         self.rating_peak = 1000.0
         self.params = params or {}
-        self.map_pools = {info.id: info for info in map_pools}
+        self.map_pools = {info.map_pool.id: info for info in map_pools}
 
         self._queue: dict[Search, None] = OrderedDict()
         self.on_match_found = on_match_found
--- a/server/types.py
+++ b/server/types.py
@@ -146,6 +146,7 @@ class NeroxisGeneratedMap(NamedTuple):
             folder_name=folder_name,
             ranked=True,
             weight=self.weight,
+            map_pool_map_version_id=self.map_pool_map_version_id,
         )
 
 
--- a/tests/integration_tests/test_matchmaker_vetoes.py
+++ b/tests/integration_tests/test_matchmaker_vetoes.py
@@ -175,7 +175,7 @@ async def test_map_pool_changes_causing_silent_update_and_not_stops_search(playe
             )
         await ladder_service.update_data()
         msg = await read_until_command(proto, "vetoes_info", timeout=10)
-        print(f"Received vetoes_info: {msg}")
+
         assert msg.get("forced") is False
         assert msg["vetoes"] == gen_vetoes([])
         assert player.state == PlayerState.SEARCHING_LADDER
--- a/tests/unit_tests/test_map_pool.py
+++ b/tests/unit_tests/test_map_pool.py
@@ -250,7 +250,7 @@ def test_choose_map_raises_on_empty_map_pool(map_pool_factory):
             {1: 0, 2: 1.5, 3: 1.5}
         ),
         # Really complex redistribution test
-        # 1 -> [5,6], [2,3] -> [5,6,8], [7,9] -> 5
+        # 1 -> [5], [2,3] -> [5,6,8], [7,9] -> 5
         (
             {1: 0.9, 2: 0.6, 3: 0.4, 4: 0.2, 5: 0.75, 6: 0.6, 7: 1, 8: 0.5, 9: 1},
             [1, 1, 2, 3, 7, 9],
@@ -261,8 +261,8 @@ def test_choose_map_raises_on_empty_map_pool(map_pool_factory):
                 2: 0,
                 3: 0,
                 4: 0.2,
-                5: pytest.approx(3.655405, rel=1e-6),
-                6: pytest.approx(1.324324, rel=1e-6),
+                5: pytest.approx(4.055405, rel=1e-6),
+                6: pytest.approx(0.924324, rel=1e-6),
                 7: 0,
                 8: pytest.approx(0.770270, rel=1e-6),
                 9: 0
@@ -279,8 +279,8 @@ def test_choose_map_raises_on_empty_map_pool(map_pool_factory):
                 2: 0,
                 3: 0,
                 4: 0.2,
-                5: pytest.approx(3.655405, rel=1e-6),
-                6: pytest.approx(1.324324, rel=1e-6),
+                5: pytest.approx(4.055405, rel=1e-6),
+                6: pytest.approx(0.924324, rel=1e-6),
                 7: 0,
                 8: pytest.approx(0.770270, rel=1e-6),
                 9: 0
--- a/tests/unit_tests/test_veto_system.py
+++ b/tests/unit_tests/test_veto_system.py
@@ -27,8 +27,6 @@ def test_is_valid_veto_config_for_queue(queue_factory):
     ) is False
 
     # minimum_maps_after_veto equal to map pool size
-    # TODO: Is calling this invalid really the desired behavior? When
-    # max_tokens_per_map != 0, they are allowed to be equal
     assert _is_valid_veto_config_for_queue(
         queue,
         MatchmakerQueueMapPool(
@@ -40,7 +38,7 @@ def test_is_valid_veto_config_for_queue(queue_factory):
             max_tokens_per_map=0,
             minimum_maps_after_veto=1,
         )
-    ) is False
+    ) is True
 
     assert _is_valid_veto_config_for_queue(
         queue,


server/ladder_service/ladder_service.py
```diff
@@ -523,7 +534,25 @@ def get_displayed_rating(player: Player) -> float:
...
+            tokensTotalPerPlayer = defaultdict(int)
+
+            for (id, map) in pool.maps.items():
+                for player in all_players:
+                    vetoesMap[map.map_pool_map_version_id] += player.vetoes.get(map.map_pool_map_version_id, 0)
+                    tokensTotalPerPlayer[player.id] += player.vetoes.get(map.map_pool_map_version_id, 0)
+
+            for player in all_players:
+                if (tokensTotalPerPlayer[player.id] > veto_tokens_per_player):
+                    raise RuntimeError(f"Player {player.id} has too many vetoes!")
```
Sheikah45: We should not do validation of the number of tokens per player when starting the game. This validation and error raising should happen when the server receives the set_vetoes message. This is because as it is now a player could send the server more than the max amount of vetoes and then cause a game to never be able to launch.

server/ladder_service/ladder_service.py
```diff
@@ -673,6 +702,26 @@ async def launch_match(
...
+            divider = index - M
+            if (divider <= 0): 
+                continue
+            
+            result = sum(sorted_tokens[:index]) / divider
+            upperLimit = sorted_tokens[index] if index < len(sorted_tokens) else float('inf')
+            if (result <= upperLimit):
+                return result
+                
+        raise Exception("Failed to calculate dynamic tokens per map: impossible vetoes setup")
```
Sheikah45: Similarly here we are throwing an exception based on user input so a user could craft a veto selection that could cause the majority of games to fail to start. It would be better to have some default value rather than throwing an exception.

Sheikah45 (changes requested):

K-ETFreeman:
also idk how least_common should work
currently set just X2 multiplier if map is least_played and not vetoed by anyone
but we can set to X5 or X10, it seems really harmless 
but maybe not, depends how exactly this history works, is it being reset or not

server/ladder_service/ladder_service.py
```diff
@@ -523,7 +542,23 @@ def get_displayed_rating(player: Player) -> float:
             pool = queue.get_map_pool_for_rating(rating)
             if not pool:
                 raise RuntimeError(f"No map pool available for rating {rating}!")
-            game_map = pool.choose_map(played_map_ids)
+
+            pool, _, _, _, max_tokens_per_map, minimum_maps_after_veto = queue.map_pools[pool.id]
```
Gatsik: you can use `*_` instead of 3 `_` to unpack unneeded variables

server/ladder_service/ladder_service.py
```diff
@@ -523,7 +542,23 @@ def get_displayed_rating(player: Player) -> float:
...
+
+            vetoesMap = defaultdict(int)
```
Gatsik: why is it suddenly in camelCase?

server/ladder_service/ladder_service.py
```diff
@@ -523,7 +542,23 @@ def get_displayed_rating(player: Player) -> float:
...
+
+            for m in pool.maps.values():
+                for player in all_players:
+                    vetoesMap[m.map_pool_map_version_id] += player.vetoes.get(m.map_pool_map_version_id, 0)
+
+            if (max_tokens_per_map == 0):
+                max_tokens_per_map = self.calculate_dynamic_tokens_per_map(minimum_maps_after_veto, vetoesMap.values())
```
Gatsik: `calculate_dynamic_tokens_per_map` expects `list[int]`, but `vetoesMap.values()` are actually [not a list](https://docs.python.org/dev/library/stdtypes.html#dict-views). if you want to annotate dict's values, you can use `ValuesView` from `collections.abc`

server/ladder_service/ladder_service.py
```diff
@@ -673,6 +708,35 @@ async def launch_match(
                 if player not in connected_players
             ])
 
+    def calculate_dynamic_tokens_per_map(self, M: float, tokens: list[int]) -> float:
```
Gatsik: are you trying to achieve `sum(weights) = M`?

server/matchmaker/matchmaker_queue.py
```diff
@@ -78,12 +78,15 @@ def add_map_pool(
         self,
         map_pool: MapPool,
         min_rating: Optional[int],
-        max_rating: Optional[int]
+        max_rating: Optional[int],
```
Gatsik: why are all of those arguments not inside `MapPool` :thinking: ?

server/players.py
```diff
@@ -89,6 +90,46 @@ def faction(self, value: Union[str, int, Faction]) -> None:
...
+
+    @vetoes.setter
+    def vetoes(self, value: dict[int, int]) -> None:
+        if not isinstance(value, dict):
+            raise ValueError("Vetoes must be a dictionary")
+        if not all(isinstance(key, int) and isinstance(val, int) and val >= 0 for key, val in value.items()):
+            raise ValueError("Incorrect vetoes dictonary")
+        self._vetoes = value
+
+    async def update_vetoes(self, pools_vetodata: list[tuple[list[int], int, int]], current: dict = None) -> None:
```
Gatsik: you could create a container (NamedTuple) for this `tuple[int], int, int` structure and avoid those long type hints and maybe call it `MapPoolVeto`

also, if you put tokens and ratings information in `MapPool` (in database they are in the same table, why are they separated here?) you won't even need this additional structure and you will be able to pass `MapPool` and use its properties

finally, this can also possibly remove `get_pools_veto_data` method since it just rearranges `MapPool`'s internals

server/matchmaker/map_pool.py
```diff
@@ -50,10 +52,13 @@ def choose_map(self, played_map_ids: Iterable[int] = ()) -> Map:
                 least_common = least_common[:i]
                 break
 
-        weights = [self.maps[id_].weight for id_, _ in least_common]
+        least_common_ids = {id_ for id_, _ in least_common}
 
-        map_id = random.choices(least_common, weights=weights, k=1)[0][0]
-        return self.maps[map_id].get_map()
+        # Multiply weight by 2 if map is least common and not vetoed by anyone
+        mapList = list((map.map_pool_map_version_id, map, 2 if (map.id in least_common_ids) and (vetoesMap.get(map.map_pool_map_version_id, 0) == 0) else 1) for id, map in self.maps.items())
```
Gatsik: don't you feel like this line is a bit long?)

Gatsik (commented):
I only wanted to ask what `calculate_dynamic_tokens_per_map` function is doing, because i don't fully comprehend it from the code, but also left some other comments/suggestions.

I am not a maintainer of this repo, therefore feel free to ignore them or use them at your own risk

server/matchmaker/matchmaker_queue.py
K-ETFreeman: because its not in the mapPool in the database, but in matchmaker_queue_map_pool (queue bracket), so should be consistent between repos
and yea its that way in db due to some reasons which were discussed long ago (f.e. u can use same pool for multiple brackets)

server/players.py
K-ETFreeman: vetoes are not in map pool in database https://faforever.github.io/db/tables/map_pool.html
vetoes are here: https://faforever.github.io/db/tables/matchmaker_queue_map_pool.html
the same place where min and max rating, outside of map pool itself

server/matchmaker/matchmaker_queue.py
K-ETFreeman: map pool https://faforever.github.io/db/tables/map_pool.html
mm queue bracket https://faforever.github.io/db/tables/matchmaker_queue_map_pool.html

server/ladder_service/ladder_service.py
K-ETFreeman: calculate_dynamic_tokens_per_map finds the minimal possible max_tokens_per_map while still respecting M

server/players.py
K-ETFreeman: also i am not sure wdym removing get_pools_veto_data
it gives only necessary data and readable name for it, allows to avoid any code duplication and keeps the code more clean
i am using it in multiple places so if we remove it code will become just worse

server/matchmaker/map_pool.py
K-ETFreeman: long lines for long PP

server/ladder_service/ladder_service.py
Gatsik: but when do you decide that `max_tokens_per_map` can't be reduced anymore?

K-ETFreeman:
i'll add some improvements soon

server/players.py
Gatsik: Oh, i see

What i mean is that now `matchmaker_queue_map_pool` has additional 3 new fields and it may be viable to create corresponding data container (see, for example, `FeaturedMod`).
This way you preserve readable names and also make it easier to understand what's passed to other functions

For example, in my opinion

```python
async def update_vetoes(self, list[MatchmakerQueueMapPool], current: dict | None) -> None
```

is easier to follow than to find calls to `update_vetoes` and discover what's inside of `list[tuple[list[int], int, int]]`

also,
```python
map_pools: Iterable[MatchmakerQueueMapPool]
```
is easier to understand than
```python
map_pools: Iterable[tuple[MapPool, Optional[int], Optional[int], int, int, float]] = ()
```


where `MatchmakerQueueMapPool` is
```python
class MatchmakerQueueMapPool(NamedTuple):
    map_pool: MapPool
    min_rating: int | None
    max_rating: int | None
    veto_tokens_per_player: int
    max_tokens_per_map: int
    minimum_maps_after_veto: float
```

finally, this way long unpacking can be avoided and instead of 
```python
for pool, *_, veto_tokens_per_player, max_tokens_per_map, _ in queue.map_pools.values():
```
you can easily access named attributes
```python
for pool in queue.map_pools.values():
    pool.veto_tokens_per_player...

File: server/ladder_service/ladder_service.py (hunk 5 above)
Comment (K-ETFreeman) [reply server/ladder_service/ladder_service.py]: good notice
i'll replace calculate_dynamic_tokens_per_map tokens type with Iterable[int]

File: server/ladder_service/ladder_service.py (hunk 6 above)
Comment (K-ETFreeman) [reply server/ladder_service/ladder_service.py]: Its just some math
it solves series of equations, if vetoes set correctly, one of them is guaranteed to be correct answer
if vetoes set incorrectly, it will return 0

Comment (K-ETFreeman) [reply server/ladder_service/ladder_service.py]: i'll prepare some explanation how math works

File: server/players.py (hunk 8 above)
Comment (K-ETFreeman) [reply server/players.py]: well, yea, but this going into refactoring existing code zone rather than implementing new features so i was not sure should i do it here or not

**Comment by K-ETFreeman** [comment]
ok not soon but tomorrow ;)

File: server/players.py (hunk 8 above)
Comment (Sheikah45) [reply server/players.py]: Using and changing to a NamedTuple should be fine and probably preferred as Gatsik says

Comment (K-ETFreeman) [reply server/players.py]: cannot put named tuple MatchmakerQueueMapPool right inside types.py due to circular dependencies happening (map_pool.py uses types and types uses map_pool.py in this case)
is it ok no place it right inside map_pool.py?

Comment (K-ETFreeman) [reply server/players.py:105]: and / or create cached property for pools_vetodata on ladder_service

File: server/ladder_service/ladder_service.py [hunk 10]
```diff
@@ -673,6 +713,147 @@ async def launch_match(
...
+        conclusion: whole solution process is
+           1) sorting tokens applied in ascending order
+           cycle:
+             2) including next bunch of maps (with the same tokens applied value) to our group
+             3) checking if tokens_sum == 0
+                then its case 1, and return 1 if maps_balance > 0
+             4) otherwise
+                solving equation for current group
+                and checking the result vs upper border
+                and upper border is equal to the amount of tokens applied to the map next to last map in our group, or infinity if there is no such one
```
Gatsik: Maybe the explanation could be shorter

Lets denote `max_tokens_per_map` = $d$ and `len(tokens)` = $n$. You are trying to achieve
$\sum_{i=1}^{n}(1 - \frac{a_{i}}{d}) = M$, which means $d = \frac{\sum_{i=1}^{n}a_{i}}{n - M}$

If some $a_{i} > d$ then $1 - \frac{a_{i}}{d} < 0$, which results in contributing nothing due to $weight = \max(0, 1 - \frac{a_{i}}{d})$ . Will excluding those ${a_{i}}$ from $tokens$ decrease $d$?

Given sorted array, excluding the last element ${a_{n}}$ from $tokens$ gives new $d_{1} = \frac{\sum_{i=1}^{n-1}a_{i}}{n - M - 1}$.
This excluding won't help if $d_{1} > d$

Lets denote $\sum_{i=1}^{n-1}a_{i} = S$, then $d < d_{1}$ is achieved when $\frac{S + a_{n}}{n - M} < \frac{S}{n - M - 1}$, which means $a_{n} < \frac{S + a_{n}}{n - M}$

But $S + a_{n} = \sum_{i=1}^{n}a_{i}$, therefore $a_{n} < \frac{\sum_{i=1}^{n}a_{i}}{n - M}$ &rArr; $a_{n} < d$

This means, that we can stop trying to decrease $d$ when $\max(tokens) < d$

And you can describte the algorithm as easy as:

* Find $d = \frac{\sum_{i=1}^{n}a_{i}}{n - M}$, if $d > \max(tokens)$ then $d$ is found. otherwise exclude all $a_{i} \geq d$. Repeat while $n > M$

(And of course handle special case when $d = 0$, which means $d$ can be anything and you've chosen it to be $1$)

server/ladder_service/ladder_service.py
```diff
@@ -673,6 +713,147 @@ async def launch_match(
...
+    """
```
Gatsik: maybe put docstring inside the function definition?

server/ladder_service/ladder_service.py
```diff
@@ -673,6 +713,147 @@ async def launch_match(
...
+        for queue in self.queues.values():
+            for pool, *_, veto_tokens_per_player, max_tokens_per_map, minimum_maps_after_veto in queue.map_pools.values():
+                if max_tokens_per_map == 0 and minimum_maps_after_veto >= len(pool.maps.values()) \
+                   or max_tokens_per_map != 0 and queue.team_size * 2 * veto_tokens_per_player / max_tokens_per_map > len(pool.maps.values()) - minimum_maps_after_veto:
+                    veto_tokens_per_player = 0
+                    max_tokens_per_map = 1
+                    minimum_maps_after_veto = 1
+                    self._logger.error(f"Wrong vetoes setup detected for pool {pool.id} in queue {queue.id}")
+                result.append(
+                    MatchmakerQueueMapPoolVetoData(
```
Gatsik: I still don't really understand why you need `MatchmakerQueueMapPoolVetoData`. Why don't you just initialize `MatchmakerQueueMapPool` with correct vetoes setup and use it?

server/matchmaker/map_pool.py
```diff
@@ -50,10 +50,22 @@ def choose_map(self, played_map_ids: Iterable[int] = ()) -> Map:
...
+        weights = [max(0, (1 - vetoes_map.get(id, 0) / max_tokens_per_map) * map.weight * least_common_multiplier) for id, map, least_common_multiplier in map_list]
+        return random.choices(map_list, weights=weights, k=1)[0][1]
 
     def __repr__(self) -> str:
         return f"MapPool({self.id}, {self.name}, {list(self.maps.values())})"
+
+
+class MatchmakerQueueMapPool(NamedTuple):
+    map_pool: MapPool
+    min_rating: int | None
```
Gatsik: Maybe use `Optional[int]`, because looks like server hasn't adopted this syntax yet

server/ladder_service/ladder_service.py
K-ETFreeman: atleast to avoid having any possible issues with queue.map_pools.clear() in update_data
theoretically user can change vetoes mid-queues-update, then server will delete all user's vetoes and he will need to apply all of them again which is very bad

K-ETFreeman: Askaholic said "Technically we do run the matchmaking algorithm in a separate thread because it is the one piece of the code that has the potential to block the rest of the server if there are a lot of players in queue."

So i prefer to have robust solution here

Askaholic: I don't think that statement is relevant to vetos because the 'chose_map' stage happens after the matchmaking algorithm has finished. But you should probably set it up in such a way that vetos are locked in the moment a match is found like we do with factions in the `PlayerParty.on_matched` function. I think keeping the behavior for choosing vetos and factions consistent makes the most sense.

K-ETFreeman: > I don't think that statement is relevant to vetos because the 'chose_map' stage happens after the matchmaking algorithm has finished. But you should probably set it up in such a way that vetos are locked in the moment a match is found like we do with factions in the `PlayerParty.on_matched` function. I think keeping the behavior for choosing vetos and factions consistent makes the most sense.

its not about chose_map, it about command_set_player_vetoes in lobbyconection.py
it uses matchmaker map pool data and this is async function, and i am not 100% sure how this works here


also whats the point of locking vetoes when match found? what can possibly go wrong if we dont do that?

server/matchmaker/map_pool.py
```diff
@@ -24,39 +26,57 @@ def __init__(
...
             raise RuntimeError(f"Map pool {self.name} not set!")
 
-        # Make sure the counter has every available map
-        counter = Counter(self.maps.keys())
-        counter.update(id_ for id_ in played_map_ids if id_ in self.maps)
+        if vetoes_map is None:
+            vetoes_map = {}
+
+        # Filter and count played map IDs
+        repetition_counts = Counter(id_ for id_ in played_map_ids if id_ in self.maps)
```
Askaholic: From what I can tell, the number of repetitions is now completely ignored? That means if we're in an 8 player game, and 7 players just played map A while 1 player just played map B, both are treated equally as 'repeated maps' whereas previously map A would have been guaranteed to not be picked due to the higher play count.

server/ladder_service/ladder_service.py
Askaholic: > also whats the point of locking vetoes when match found?

As I said:
> keeping the behavior for choosing vetos and factions consistent makes the most sense.

This will be less confusing to players.

server/ladder_service/ladder_service.py
```diff
@@ -685,6 +742,151 @@ async def launch_match(
                 if player not in connected_players
             ])
 
+    def calculate_dynamic_tokens_per_map(self, M: float, tokens: Iterable[int]) -> float:
```
Askaholic: I would like to see some unit tests for this function. That might make it easier to understand just by looking at it as a black box.

server/ladder_service/ladder_service.py
```diff
@@ -685,6 +742,151 @@ async def launch_match(
...
+    def get_pools_veto_data(self) -> list[MatchmakerQueueMapPoolVetoData]:
+        result = []
+        for queue in self.queues.values():
+            for matchmaker_queue_map_pool_id, pool, *_, veto_tokens_per_player, max_tokens_per_map, minimum_maps_after_veto in queue.map_pools.values():
+                if max_tokens_per_map == 0 and minimum_maps_after_veto >= len(pool.maps.values()) \
+                   or max_tokens_per_map != 0 and queue.team_size * 2 * veto_tokens_per_player / max_tokens_per_map > len(pool.maps.values()) - minimum_maps_after_veto:
+                    veto_tokens_per_player = 0
+                    max_tokens_per_map = 1
+                    minimum_maps_after_veto = 1
+                    self._logger.error(f"Wrong vetoes setup detected for pool {pool.id} in queue {queue.id}")
```
Askaholic: Logging an error but otherwise continuing on as normal?

server/players.py
```diff
@@ -2,21 +2,29 @@
...
 
 if TYPE_CHECKING:
     from server.gameconnection import GameConnection
     from server.games import Game
     from server.lobbyconnection import LobbyConnection
 
+BracketID = int
+MapPoolMapVersionId = int
+VetoTokensApplied = int
+PlayerVetoes = dict[BracketID, dict[MapPoolMapVersionId, VetoTokensApplied]]
```
Askaholic: It might help with readability if you were to make `PlayerVetos` a real class so that you can add helper functions to it.

For instance a `get_weight` function that does this: `max(0, 1 - vetoes_map.get(id_, 0) / max_tokens_per_map)`.

Or also the `update_vetos` function currently living on the player object.

server/players.py
```diff
@@ -94,6 +105,56 @@ def faction(self, value: Union[str, int, Faction]) -> None:
...
+                fixedVetoes[matchmaker_queue_map_pool_id] = cur_fixed_vetoes
+        if fixedVetoes == self.vetoes == current:
+            return
+        self.vetoes = fixedVetoes
+        if self.lobby_connection is None:
+            return
+        await self.lobby_connection.send({
+            "command": "vetoes_changed",
+            "vetoesData": vetoDatas
+        })
```
Askaholic: This doesn't belong on the player object. Probably modeling it after the way the `PartyService` works would make more sense.

Askaholic (commented):
I'm having a very hard time reviewing this PR because it seems that at just about every step of the way there is some complicated function with 4 or 5 levels of indentation that does some sort of magic for an operation that I wouldn't have expected to be so complicated.

I think it would help a lot to have some unit tests, as tests are a great way of providing 'black box' examples of expected input/output. That might help me understand what is even going on and then make it easier to work backwards to figure out how the code is arriving at those values.

Another thing I'm always a fan of is making use of vertical whitespace to logically group operations together. At the moment there are a lot of if statements, for loops, etc. all crammed up against each other which is also contributing to the difficulty of reading the code.

If I was able to wrap my head around what was actually going on better I'd also be able to suggest more concrete ways of refactoring the code to make it easier to read. I believe there is a lot of 'extract to function' refactoring that could happen to clean things up and make it easier to see what the overarching algorithms are doing, but without a clear picture of the code I can't suggest specific refactors to do.

So yea, I think the most important next step is to start writing some unit tests as I think this will make it easier to review.

server/players.py
Askaholic: I think `PlayerRatings` would be a good example to look at.

server/matchmaker/map_pool.py
K-ETFreeman: Yes, number of repetitions is completely ignored in the code
OK, i know how it can be not-ignored, will modify it

server/ladder_service/ladder_service.py
K-ETFreeman: If you provide some mini-guide how to do so, will help a lot
I did the tests for this function in some online python compiler😂

server/ladder_service/ladder_service.py
K-ETFreeman: Yes
lines
``
 veto_tokens_per_player = 0
max_tokens_per_map = 1
minimum_maps_after_veto = 1
``
are resetting the settings, removing the vetoes from the bracket completely in this case

server/players.py
K-ETFreeman: OK

server/ladder_service/ladder_service.py
K-ETFreeman: nevermind, i am pro python tester now

K-ETFreeman: @Askaholic tests were added

server/matchmaker/map_pool.py
K-ETFreeman: > From what I can tell, the number of repetitions is now completely ignored? That means if we're in an 8 player game, and 7 players just played map A while 1 player just played map B, both are treated equally as 'repeated maps' whereas previously map A would have been guaranteed to not be picked due to the higher play count.

So, i just finished to modyfying it, and now remember why i didnt do it in a first place:
Current idea is that if some map is banned a lot more than the repeated map (weight is less than repeated one multiplied by LADDER_ANTI_REPETITION_WEIGHT_THRESHOLD), then the weight transfer doesnt happen
However, if we chain transfer from more-repeated-ones to less-repeated ones, it causes an issue: now bad map can affect good map weight, if there is some a little-banned map in between of them, and repeated one

[note: "weight" later is veto-system-weight, not map inner weight]
For example, the theshold is 0.75, so map with two repeats and weight 0.6 should not be abile to transfer weights to map with weight 1, because 1 * 0.75 = 0.75 > 0.6. But if there is middle map with repeat 1 and 0.8 weight, then weight is transferred to 0.8 weight map first, and then to map with weight 1

But if to think about it, this makes sense, kinda: the more repeats the map have, the stronger we want it to tranfer weight
so i adjusted the formula: now we multiply not by threshold, but my Math.pow(theshold, repetition_diff )
So now its not a bug, but a feature, the problem is solved

I made a lot of tests also to help to understand, what-the-hell-is-going-on here


UPD: modified the function again to resolve all issues and make it more flexible

server/ladder_service/veto_system.py
```diff
@@ -0,0 +1,213 @@
...
+    _max_stream_count = 25
+    pools_veto_data: ClassVar[list[MatchmakerQueueMapPoolVetoData]] = []
+
+    @staticmethod
+    async def apply_vetoes_to_player(player: Player, new_vetoes: Optional[dict[BracketID, dict[MapPoolMapVersionId, VetoTokensApplied]]] = None) -> None:
+        """Applies vetoes for a player and sends a message if changes occur."""
+        veto_datas = player.vetoes.apply_vetoes(new_vetoes)
+        if veto_datas:
+            try:
+                await player.send_message({
```
Askaholic: You can actually use `write_message` here instead of `send_message` which should eliminate a lot of this async worker boilerplate. `send_message` should actually only be called from within a connection handler for the player in question. If you're sending a message that is generated by some background task or by an event caused by another player, use `write_message`.

server/ladder_service/veto_system.py
```diff
@@ -0,0 +1,213 @@
...
+                    "command": "vetoes_changed",
+                    "vetoesData": veto_datas
```
Askaholic: BTW our message protocol uses snake case.
```suggestion
                    "vetoes_data": veto_datas
```

server/ladder_service/veto_system.py
K-ETFreeman: write_message is sync and i assume just doesnt care was the request fully made or not
so, if faf will have 50000 players online, and we will just do write_message in for loop, it can lead to 50000 messages sended at once and can probably explode cpu/network capacity of the server

currently i limit the max number of connections at once by _max_stream_count, waiting until each request actually finished:

```
    async def update_vetoes_of_players(players: Iterable[Player]) -> None:
        player_queue: asyncio.Queue[Player] = asyncio.Queue()
        for player in players:
            await player_queue.put(player)

        async def worker():
            while not player_queue.empty():
                player = await player_queue.get()
                try:
                    await VetoSystem.apply_vetoes_to_player(player)
                finally:
                    player_queue.task_done()

        workers = [asyncio.create_task(worker()) for _ in range(VetoSystem._max_stream_count)]
        await asyncio.gather(*workers)
```

you cannot do the same with .write_message because it doesnt wait for the request to actually finish

Askaholic: Correct. You shouldn't need to wait. Currently there are already several types of messages that are blasted out to every online player (see BroadcastService). In particular, `game_info` messages, which are certainly where a large amount of the CPU and network load comes from. However, I think the vetoes messages are probably fewer bytes than `game_info` and since this triggers only once every 10 minutes, it will be completely trivial compared to `game_info` messages which are blasted out once per second.

That said, I actually don't like this approach at all. I think we should just make the players send us valid veto selections rather than trying to adjust them automatically and then send out updated veto selections to all players.

BTW the most users I've ever seen online at a time was about 2500.

server/ladder_service/veto_system.py
```diff
@@ -0,0 +1,286 @@
...
+            "vetoes": [
+                {
+                    "map_pool_map_version_id": map_id,
+                    "veto_tokens_applied": tokens,
+                    "matchmaker_queue_map_pool_id": bracket_id
+                }
+                for bracket_id, vetoes in self._vetoes.items()
+                for map_id, tokens in vetoes.items()
+            ]
+        }
```
Askaholic, line 37: Here's an idea. Instead of using the surrogate keys for `map_pool_map_version_id` and `matchmaker_queue_map_pool_id` which really only exist for database optimization, how about we use composite keys instead. It makes no difference to us on the python side and I think it would simplify things quite a bit, since you wouldn't have to keep track of what the difference between all these different ids is.

So instead of `map_pool_map_version_id` we would just use the `map_pool_id` and the `map_version_id` together.

And instead of `matchmaker_queue_map_pool_id` we would use `queue_name` (the technical name would be used in the protocol messages, but it doesn't matter if we use queue_id or queue_name internally) and `map_pool_id`.

Immediately I notice that `map_pool_id` is shared between both of those keys which is a nice win. The message would look something like this:

```python
{
    "vetoes": [
        {
            "queue_name": "ladder1v1",
            "map_pool_id": 123,
            "map_version_id": 123456,
            "tokens": 4,
        },
        ...
    ]
}
```

But some of the redundancy could be reduced more if we use a nested dict rather than a list:

```python
{
    "vetoes": {
        "ladder1v1": {
            "123": {
                "123456": 4,
                ...
            }
        }
    },
}
```

Of course that loses the expressiveness, so maybe some balance between the two:

```python
{
    "vetoes": {
        "ladder1v1": [
            {
                "map_pool_id": 123,
                "map_version_id": 123456,
                "tokens": 4,
            },
            ...
         ]
    },
}
```

K-ETFreeman: Do as you see, my lord, i will adjust the client / commons / etc to whatever server will be in the end
I also noticed all this boilerplate with multiple ids, was thinking like making mapPoolMap actual class instead of just a protocol, so mapPoolMap has id (which currently is map_pool_map_version_id), and map inside of it has own id, so we dont have two ids in the same object

K-ETFreeman:
image

On top (1) its dynamic tokens per map case. Dynamic means it can shrink any positive amount of tokens to any positive value > 0. 
So only case this fails is when you have 5 maps in pool and minimum_maps_after veto demands to have atleast 5 maps left after bans. So we have >= here: if we have 5 maps, but mm team wants 5 or 6, players will not be able to ban anything, the setup is incorrect.
Actually need to check out case when veto_tokens_per_player = 0 so we can allow == here (only in this case), but this is another topic i guess (i'll need to remember the codebase again to say for sure is this needed to be done or not; maybe 0 tokens per player cannot be set for dynamic system?)

On the bottom (2) we are checking how many maps can be banned (by all players in total) vs how many maps max can be banned (as mm team prefers). It they are equal, that means players ban maximum amount of maps, but this is still OK

K-ETFreeman:
@ Gatsik since askaholic left, may you review this PR? 
People kinda wait this feature for 1.5 years now would be nice to finish it
Its working, tests are done (pls approve)

(note: the client for the vetos still required to be polished, so approving is ok, merging is not, too soon)
**actually whatever, we'll test it on the test server and call it a day**

server/ladder_service/veto_system.py
```diff
@@ -0,0 +1,335 @@
...
+        return False
+
+    if queue_config.max_tokens_per_map != 0:
+        total_players = queue.team_size * 2
+        vetoable_maps_per_player = queue_config.veto_tokens_per_player / queue_config.max_tokens_per_map
+        vetoable_maps = total_players * vetoable_maps_per_player
+        if vetoable_maps > num_maps - queue_config.minimum_maps_after_veto:
+            return False
+
+    return True
```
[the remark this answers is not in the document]
K-ETFreeman: not an issue, it cannot be negative

server/ladder_service/veto_system.py
```diff
@@ -0,0 +1,335 @@
...
+                        matchmaker_queue_map_pool_id=matchmaker_queue_map_pool.id,
+                        map_pool_map_version_ids=[
+                            map.map_pool_map_version_id
+                            for map in matchmaker_queue_map_pool.map_pool.maps.values()
+                        ] + [-1],
+                        veto_tokens_per_player=veto_tokens_per_player,
+                        max_tokens_per_map=max_tokens_per_map,
+                        minimum_maps_after_veto=minimum_maps_after_veto
+                    )
+                )
```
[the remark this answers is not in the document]
K-ETFreeman: i actually have no idea what it is, need to ponder for some time

server/matchmaker/map_pool.py
```diff
@@ -24,39 +26,96 @@ def __init__(
...
-        map_id = random.choices(least_common, weights=weights, k=1)[0][0]
-        return self.maps[map_id].get_map()
+        adjusted_weights = self.apply_antirepetition_adjustment(
+            initial_weights, played_map_pool_map_version_ids, config.LADDER_ANTI_REPETITION_WEIGHT_BASE_THRESHOLDS, config.LADDER_ANTI_REPETITION_REPEAT_COUNTS_FACTOR
+        )
+        self._logger.debug("______adjusted_weights________________: %s", adjusted_weights)
+        self._logger.debug("______map_list________________: %s", map_list)
+        final_weights = [adjusted_weights.get(mp_mv_id, 0) * m.weight for mp_mv_id, m in map_list]
+        self._logger.debug("______final_weights________________: %s", final_weights)
+        return random.choices([map for _, map in map_list], weights=final_weights, k=1)[0].get_map()  # nosec B311
```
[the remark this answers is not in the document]
K-ETFreeman: system doesnt allow to make fully bannable pool in a first place, this is not necessary

server/matchmaker/matchmaker_queue.py
```diff
@@ -66,7 +66,7 @@ def __init__(
         self.team_size = team_size
         self.rating_peak = 1000.0
         self.params = params or {}
-        self.map_pools = {info[0].id: info for info in map_pools}
+        self.map_pools = {info.id: info for info in map_pools}
```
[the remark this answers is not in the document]
K-ETFreeman: oops. ok.

server/types.py
```diff
@@ -29,11 +29,22 @@ class GameLaunchOptions(NamedTuple):
...
+    veto_tokens_per_player: int
+    max_tokens_per_map: float
+    minimum_maps_after_veto: float
 
 
 class MapPoolMap(Protocol):
     @property
     def id(self) -> Optional[int]: ...
+    @property
+    def map_pool_map_version_id(self) -> int: ...
```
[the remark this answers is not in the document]
K-ETFreeman: Nah, each mapPoolMap indeed must have map_pool_map_version_id.

server/ladder_service/veto_system.py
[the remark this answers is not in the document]
K-ETFreeman: good catch. it was the relic from the old days where i thought to add option for players to ban "last played map" to not have headache with antirep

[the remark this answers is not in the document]
K-ETFreeman: deleting now.

server/types.py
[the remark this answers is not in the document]
K-ETFreeman: yes, i already added it. thanks.

tests/integration_tests/test_matchmaker_vetoes.py
```diff
@@ -0,0 +1,186 @@
...
+        msg = await read_until_command(proto, "vetoes_info", timeout=10)
+        print(f"Received vetoes_info: {msg}")
+        assert msg.get("forced") is False
+        assert msg["vetoes"] == gen_vetoes([])
+        assert player.state == PlayerState.SEARCHING_LADDER
+    finally:
+        async with database.acquire() as conn:
+            await conn.execute(
+                "REPLACE INTO map_pool_map_version (id, map_pool_id, map_version_id, weight, map_params) VALUES (1, 1, 15, 1, NULL)"
+            )
```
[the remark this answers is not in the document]
K-ETFreeman: oops. deleted.

PR #1038, opened 2025-01-22
Use cached_property to create Futures

### LINKED ISSUE #894: RuntimeError raised when running some tests on their own
There are some tests that construct objects which try to access the current event loop in their constructors (such as `Search` and `Game`). If these tests are not declared to be async then when they are run in isolation they will raise this error:
```
RuntimeError: There is no current event loop in thread 'MainThread'.
```
However, they work when running the test suite as a whole because they will get the event loop created by previous tests. Either the tests should be converted to async functions, or the classes causing the problem need to be refactored so that they can be constructed without an event loop existing.

Currently at least these tests have the problem:
- Everything in `tests/unit_tests/test_matchmaker_algorithm_bucket_teams.py`

The error can be reproduced by running tests like this `pipenv run tests tests/unit_tests/test_matchmaker_algorithm_bucket_teams.py::test_make_teams_single_correct_size`

Gatsik:
Resolves #894 

Also, fix one flaky test (`test_game_matchmaking_start`) by fully waiting for the game to start.

I took a look at other flaky tests, but I couldn't reproduce their flakiness: they had been continuing to pass.

The only flaky one which sometimes failed is [`test_game_matchmaking_close_fa_and_requeue`](https://github.com/FAForever/server/blob/181bee80bd79a5507219999a811aa71515e9aeb3/tests/integration_tests/test_matchmaker.py#L327), and I believe this is happening due to `@fast_forward` not working as wished with `loop.run_in_executor` futures.

Removing `@fast_forward` decorator fixes this test's flakiness, but increases runtime (from ~1 to about ~2.5 seconds), so I didn't touch that.

### DIFF
--- a/server/games/game.py
+++ b/server/games/game.py
@@ -6,6 +6,7 @@ import pathlib
 import time
 from collections import defaultdict
 from datetime import datetime
+from functools import cached_property
 from typing import TYPE_CHECKING, Any, Awaitable, Callable, Iterable, Optional
 
 from sqlalchemy import and_, bindparam
@@ -139,11 +140,14 @@ class Game:
         self.game_options.add_callback("Title", self.on_title_changed)
 
         self.mods: dict[str, str] = {}
-        self._hosted_future: asyncio.Future[None] = asyncio.Future()
         self._finish_lock = asyncio.Lock()
 
         self._logger.debug("%s created", self)
         asyncio.get_event_loop().create_task(self.timeout_game(setup_timeout))
+
+    @cached_property
+    def _hosted_future(self) -> asyncio.Future:
+        return asyncio.get_running_loop().create_future()
 
     async def timeout_game(self, timeout: int = 60):
         await asyncio.sleep(timeout)
--- a/tests/integration_tests/test_matchmaker.py
+++ b/tests/integration_tests/test_matchmaker.py
@@ -100,9 +100,8 @@ async def test_game_launch_message_game_options(lobby_server, tmp_user):
         }
 
 
-@pytest.mark.flaky
 @fast_forward(15)
-async def test_game_matchmaking_start(lobby_server, database):
+async def test_game_matchmaking_start(lobby_server, database, game_service):
     host_id, host, guest_id, guest = await queue_players_for_matchmaking(lobby_server)
 
     # The player that queued last will be the host
@@ -139,8 +138,7 @@ async def test_game_matchmaking_start(lobby_server, database):
     })
 
     # Wait for db to be updated
-    await read_until_launched(host, game_id)
-    await read_until_launched(guest, game_id)
+    await game_service[game_id].wait_launched(None)
 
     async with database.acquire() as conn:
         result = await conn.execute(select(
--- a/server/games/ladder_game.py
+++ b/server/games/ladder_game.py
@@ -1,5 +1,6 @@
 import asyncio
 import logging
+from functools import cached_property
 from typing import Optional
 
 from server.config import config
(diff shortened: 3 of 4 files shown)


server/games/game.py
```diff
@@ -132,12 +133,15 @@ def __init__(
...
         self.mods = {}
-        self._hosted_future = asyncio.Future()
         self._finish_lock = asyncio.Lock()
 
         self._logger.debug("%s created", self)
         asyncio.get_event_loop().create_task(self.timeout_game(setup_timeout))
 
+    @cached_property
+    def _hosted_future(self) -> asyncio.Future:
+        return asyncio.get_event_loop().create_future()
```
Askaholic: Should we be using `get_running_loop` instead maybe? I'm not entirely sure what all the intricacies are but the asyncio docs for `get_event_loop` in python 3.10 say this:

> Because this function has rather complex behavior (especially when custom event loop policies are in use), using the [get_running_loop()](https://docs.python.org/3.10/library/asyncio-eventloop.html?highlight=get_event_loop#asyncio.get_running_loop) function is preferred to [get_event_loop()](https://docs.python.org/3.10/library/asyncio-eventloop.html?highlight=get_event_loop#asyncio.get_event_loop) in coroutines and callbacks.

tests/integration_tests/test_matchmaker.py
```diff
@@ -139,8 +138,7 @@ async def test_game_matchmaking_start(lobby_server, database):
     })
 
     # Wait for db to be updated
-    await read_until_launched(host, game_id)
-    await read_until_launched(guest, game_id)
+    await game_service[game_id].wait_launched(None)
```
Askaholic, line 141: I feel like this change wouldn't make any difference. In fact, as far as I can tell from the code the `read_until_launched` method should actually always wait *longer* than the `wait_launched` method. So I'm not sure if this actually fixes the flakiness or if maybe you just got lucky when running the tests. But I could be missing something.

I also want to keep these integration tests using only the protocol messages as much as possible. So if we can use a protocol message to determine that the game launched instead of having to reach into the internal state of a service, that would be much better. That way it can reveal deficiencies in our protocol if there are some use cases that are not possible using only the existing messages.

Askaholic (commented):
Sorry it took me a while to review. I wasn't sure about the approach at first, but after thinking about it for a while, I think I like it! I have just one question about the implementation, and one comment about the test change.

tests/integration_tests/test_matchmaker.py
Gatsik: Hey! Yes, you are missing a few things here:
1. Setting player options marks game as dirty
2. `read_until_launched` method has `launched_at` as a criteria of game launch, which it receives from server message
3. Sending `launched_at` is not event-based -- broadcast service reports dirties every (more or less) fixed interval (currently it is 1 second)
4. `launched_at` is set _before_ updates to the database are made (they happen in `on_game_launched` method):
https://github.com/FAForever/server/blob/9d6ba68e24f9bd60fbd83836e45820a8fb2aa4c0/server/games/game.py#L727-L739
5. `await` statements tell event loop that it can process some other scheduled tasks

`fast_forward` just exaggerates timing discrepancies -- `report_dirties` does not wait full second between reports -- therefore reporting happens _during_ database update more often.

But the test can fail even without `fast_forward` -- just tweak the sleep time here (0.01 and 0.99 worked for me):
https://github.com/FAForever/server/blob/9d6ba68e24f9bd60fbd83836e45820a8fb2aa4c0/tests/integration_tests/test_matchmaker.py#L126-L133
(I also don't completely understand the purpose of this sleep, I suppose it 'simulates' launching fa process on the client side). Anyway, some change to broadcast interval can break this test

On the other hand, `wait_launched` method of `ladder_game` is event-based -- it _always_ sets its `_launch_future` after database update (which happens in super class):
https://github.com/FAForever/server/blob/9d6ba68e24f9bd60fbd83836e45820a8fb2aa4c0/server/games/ladder_game.py#L46-L48

If you wanted to find deficiencies in protocol -- it is the one.

There's also another thing: at least in `integration_tests` services are initialized twice -- first time in their corresponding fixtures, and the second time here:
https://github.com/FAForever/server/blob/9d6ba68e24f9bd60fbd83836e45820a8fb2aa4c0/tests/integration_tests/conftest.py#L161
because `instance.started` is `False`

I split this PR in 2 commits, so if you insist on leaving flaky test as is you can drop the fix

PR #1039, opened 2025-01-26
Ensure all generated maps are ranked

Sheikah45:
The regex for determining if something was a generated map is too restrictive and accidentally excludes some generated maptypes.


Askaholic:
Can we add a unit test with an example of a real map name that this was failing for? Just testing `is_neroxis_map` should be sufficient.

Askaholic (approved):

PR #1045, opened 2025-02-09
feat(messaging): game join failed command

### LINKED ISSUE #1044: Add an event object for handling negative game joins
Related to fix for https://github.com/FAForever/downlords-faf-client/issues/3301

Ivan-Shaml:
When player fails to join a game, instead of `notice`, a new `game_join_failed` event is being sent, so the consumer can programmatically handle the exceptional behavior

This will be used in `faf-java-commons` for handling this situations, and then finally in the client

Closes #1044


server/exceptions.py
```diff
@@ -65,3 +65,16 @@ class DisabledError(Exception):
...
+class GameJoinError(Exception):
+    """
+    Represents an error occurring when a client tries to join a game.
+
+    uid is an integer representing the requested game ID.
+    """
+
+    def __init__(self, message, uid, *args, **kwargs):
+        super().__init__(*args, **kwargs)
+        self.message = message
```
Askaholic: I think we don't need to include the english messages. We should use fixed error codes instead like `invalid_uid`, `invalid_state`, `invalid_password` etc.

server/lobbyconnection.py
```diff
@@ -202,6 +203,18 @@ async def on_message_received(self, message):
...
+                "GameJoinError[%s]: %s",
+                self.user_agent,
+                e.message,
+            )
+            await self.send({
+                "command": "game_join_failed",
+                "style": "error",
+                "text": e.message,
+                "uid": e.uid
+            })
```
Askaholic: I think probably the error class isn't necessary. We can just send the messages directly

server/lobbyconnection.py
```diff
@@ -965,32 +978,35 @@ async def command_game_join(self, message):
...
                 "style": "info",
-                "text": "The host has left the game."
+                "text": "The host has left the game.",
+                "uid": uuid
             })
             return
 
         if self.player.id in game.host.foes:
-            raise ClientError("You cannot join games hosted by this player.")
+            raise GameJoinError("You cannot join games hosted by this player.", uuid)
```
Askaholic: These client errors we need to think about if they should actually be converted to `game_join_failed` messages or if they should remain `ClientErrors`. A `ClientError` class should be used anywhere that we are expecting the client to be enforcing some sort of condition because it has all the necessary data to do so. For instance, if a player is foed by the host, then the server will not send `game_info` messages for the game to that player. So there should be no way for the foed player to even know the game uid in order to initiate a game join. So if for some reason that does happen, we treat that as a `ClientError` to signal a possible bug in the client. (It could technically also be a bug in the server if it isn't sending the necessary state info, but either way it's a bug and not an expected scenario).

server/lobbyconnection.py
```diff
@@ -965,32 +978,35 @@ async def command_game_join(self, message):
...
                 "style": "info",
-                "text": "The game you are trying to join is not ready."
+                "text": "The game you are trying to join is not ready.",
+                "uid": uuid
             })
             return
 
         if game.init_mode != InitMode.NORMAL_LOBBY:
-            raise ClientError("The game cannot be joined in this way.")
+            raise GameJoinError("The game cannot be joined in this way.", uuid)
```
Askaholic: Similar thing here, I think `ClientError` is correct since this would mean that the client is trying to join a matchmaker game as if it were a custom game, which means there is a bug in the client. The client should never have to handle this error message because it should never be acting in such a way that it gets triggered in the first place.

server/lobbyconnection.py
```diff
@@ -965,32 +978,35 @@ async def command_game_join(self, message):
             game = self.game_service[uuid]
         except KeyError:
             await self.send({
-                "command": "notice",
+                "command": "game_join_failed",
```
Askaholic, line 961: For the rest of these, I think we will need to continue sending the notice messages in addition to the `game_join_failed` messages for protocol backwards compatibility. We can mark them as deprecated with a comment though and remove them with a breaking change (protocol version 2).

Askaholic (commented):
I think this is a great idea, and much needed addition honestly. I would have done it ages ago if I had someone to do the client side work for it, so happy to see someone is able to do that!

Some of the tests are flaky and will pass when we re-run them, although one of them is failing because the `notice` messages were removed and the test is expecting them to be there. That's why we need to keep sending the notice messages even with the addition of the new `game_join_failed` messages.

server/exceptions.py
Ivan-Shaml: yes, I need to think about a structure for the object that  we are gonna send, that will be useful for the client. With the new version check we have the ability to do this breaking change now.

server/lobbyconnection.py
Ivan-Shaml: you are correct. fixed

server/lobbyconnection.py
Ivan-Shaml: Got it, then we leave them as `ClientError` for now. Will see if this can be checked on the client side.

server/lobbyconnection.py
Ivan-Shaml: Yes, I have developed a client version check mechanism, so we can remain backwards-compatible.

Ivan-Shaml:
So as discussed, the focus mainly in this PR is to fix the java client bug (https://github.com/faforever/downlords-faf-client/issues/3301) and the overhaul of the server notice messaging implementation will be reworked in another PR

server/exceptions.py
Ivan-Shaml: will be done in another PR, not in the scope of this one

server/lobbyconnection.py
```diff
@@ -996,6 +1008,11 @@ async def command_game_join(self, message):
 
         await self.launch_game(game, is_host=False)
 
+    async def send_game_join_failed_response(self, new_message, legacy_message):
+        await self.send(new_message)
+        # For backwards compatibility
+        await self.send(legacy_message)
```
Askaholic: I think I would not make a new method for this. Just call the `self.send` twice in each place. We should also add a comment that says `# DEPRECATED: use 'game_join_failed' instead` above each of the notice messages. That will help me to remember to remove those on the v2 branch because I can just search for `DEPRECATED` in the whole project.

server/lobbyconnection.py
```diff
@@ -976,7 +980,11 @@ async def command_game_join(self, message):
 
         if not game or game.state is not GameState.LOBBY:
             self._logger.debug("Game not in lobby state: %s state %s", game, game.state)
-            await self.send({
+            await self.send_game_join_failed_response({
```
Askaholic: We should probably add an integration test for this case, I'm actually a little surprised there isn't one yet. We should also update any integration tests that are currently checking for the notice messages to check for the `game_join_failed` messages as well.

Askaholic (commented):
Looks good, I think this is a good first step to take. I would be happy to make the necessary unit test adjustments if you want. I'm guessing I probably have permissions to push to your branch.

server/lobbyconnection.py
```diff
@@ -964,6 +964,12 @@ async def command_game_join(self, message):
         try:
             game = self.game_service[uuid]
         except KeyError:
+            await self.send({
+                "command": "game_join_failed",
+                "reason": "HOST_LEFT_GAME",
```
Askaholic: The only other thing I'm thinking now is if maybe these codes should be lowercase. I think everything else in the protocol is generally lowercase.

Askaholic (approved):

PR #1048, opened 2025-02-18
Log every attempt to join a game

Brutus5000:
needs db v137

### DIFF
--- a/.github/workflows/test.yml
+++ b/.github/workflows/test.yml
@@ -11,7 +11,7 @@ on:
     - cron: '0 0 * * *'
 
 env:
-  FAF_DB_VERSION: v136
+  FAF_DB_VERSION: v137-RC1
   FLYWAY_VERSION: 7.5.4
 
 jobs:
--- a/.github/workflows/test.yml
+++ b/.github/workflows/test.yml
@@ -11,7 +11,7 @@ on:
     - cron: '0 0 * * *'
 
 env:
-  FAF_DB_VERSION: v137-RC1
+  FAF_DB_VERSION: v138
   FLYWAY_VERSION: 7.5.4
 
 jobs:
--- a/server/db/models.py
+++ b/server/db/models.py
@@ -1,6 +1,7 @@
 from sqlalchemy import (
     TIME,
     TIMESTAMP,
+    BigInteger,
     Boolean,
     Column,
     DateTime,
@@ -111,6 +112,13 @@ game_featuredMods = Table(
     Column("git_branch",     String),
     Column("file_extension", String),
     Column("allow_override", Boolean)
+)
+
+game_join_log = Table(
+    "game_join_log", metadata,
+    Column("id",         BigInteger, primary_key=True),
+    Column("player_id",  Integer,    ForeignKey("login.id"), nullable=False),
+    Column("game_id",    Integer,    nullable=False),
 )
 
 game_player_stats = Table(
--- a/server/lobbyconnection.py
+++ b/server/lobbyconnection.py
@@ -27,6 +27,7 @@ from .db.models import (
     ban,
     coop_map,
     friends_and_foes,
+    game_join_log,
     lobby_ban
 )
 from .db.models import login as t_login
@@ -952,6 +953,17 @@ class LobbyConnection:
 
         uuid = int(message["uid"])
         password = message.get("password")
+
+        async with self._db.acquire() as conn:
+            try:
+                await conn.execute(
+                    game_join_log.insert().values(
+                        player_id=player_id,
+                        game_id=uuid,
+                    )
+                )
+            except DBAPIError:
+                self._logger.exception("writing to game join log failed")
 
         self._logger.debug("joining: %d with pw: %s", uuid, password)
         try:
--- a/server/lobbyconnection.py
+++ b/server/lobbyconnection.py
@@ -958,7 +958,7 @@ class LobbyConnection:
             try:
                 await conn.execute(
                     game_join_log.insert().values(
-                        player_id=player_id,
+                        player_id=self.player.id,
                         game_id=uuid,
                     )
                 )
--- a/tests/integration_tests/test_game.py
+++ b/tests/integration_tests/test_game.py
@@ -7,9 +7,9 @@ from datetime import datetime
 from unittest import mock
 
 import pytest
-from sqlalchemy import select
+from sqlalchemy import select, and_
 
-from server.db.models import game_player_stats
+from server.db.models import game_player_stats, game_join_log
 from server.games.game_results import GameOutcome
 from server.protocol import Protocol
 from server.timing import datetime_now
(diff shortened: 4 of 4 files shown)


server/db/models.py
```diff
@@ -113,6 +115,14 @@
     Column("allow_override", Boolean)
 )
 
+game_join_log = Table(
+    "game_join_log", metadata,
+    Column("bigint",        BigInteger, primary_key=True),
```
Sheikah45:
```suggestion
    Column("id",        BigInteger, primary_key=True),
```

server/lobbyconnection.py
```diff
@@ -76,7 +77,7 @@ def __init__(
         rating_service: RatingService,
         oauth_service: OAuthService,
     ):
-        self._db = database
+        created_at = database
```
Sheikah45: ?

Brutus5000: Late night copy&paste misclick I assume :D

server/db/models.py
```diff
@@ -113,6 +115,14 @@
     Column("allow_override", Boolean)
 )
 
+game_join_log = Table(
+    "game_join_log", metadata,
+    Column("id",         BigInteger, primary_key=True),
+    Column("player_id",  Integer,    ForeignKey("login.id"),      nullable=False),
+    Column("game_id",    Integer,    ForeignKey("game_stats.id"), nullable=False),
```
Askaholic: Shouldn't be a foreign key on `game_id`.

server/db/models.py
```diff
@@ -113,6 +115,14 @@
...
+    Column("created_at", TIMESTAMP,  nullable=False, server_default=func.now()),
```
Askaholic: I think the `server_default` is actually irrelevant since we're never using sqlalchemy to actually create the database tables. Looks like in our other models we've been just doing this:

```
    Column("create_time",   TIMESTAMP,  nullable=False),
    Column("update_time", TIMESTAMP, nullable=False),
```

You can also always just leave the field off completely if we don't need to ever read from it or include it in an insert statement.

Do we really want to use `created_at` rather than `create_time` like the rest of our tables?

server/lobbyconnection.py
```diff
@@ -960,6 +961,17 @@ async def command_game_join(self, message):
...
+        async with self._db.acquire() as conn:
+            try:
+                await conn.execute(
+                    game_join_log.insert().values(
+                        player_id=player_id,
+                        game_id=uuid,
+                    )
+                )
+            except DBAPIError as e:
+                self._logger.error("writing to game join log failed")
```
Askaholic: Using `.exception` from within an exception handler will automatically print the stack trace. No need to capture the exception with an `as e` syntax even.
```suggestion
                self._logger.exception("writing to game join log failed")
```

Askaholic (commented):
What is the motivation for this table? I'd also ask, is our use case solvable with a prometheus metric? Because that would be simpler.

server/db/models.py
Brutus5000: I don't know how SQLalchemy handles it. In Hibernate I have to declare it as non-insertable so it doesn't try to fill null on insert.

Will re-check on the column name.

Sheikah45:
@Askaholic don't know if brutus ever answered but we want to be able to track which users are joining more than normal games in a specific amount of time and doing that as a prometheus metric would be pretty high cardinality I would think

server/db/models.py
```diff
@@ -113,6 +114,13 @@
     Column("allow_override", Boolean)
 )
 
+game_join_log = Table(
+    "game_join_log", metadata,
+    Column("id",         BigInteger, primary_key=True),
```
Askaholic, line 119: I think `Integer` should suffice for reading purposes? We don't need to support table creation IMO since we'll never need that and it keeps the models a bit simpler.

tests/integration_tests/test_game.py
```diff
@@ -24,12 +24,12 @@
...
-    mod: str = "faf",
-    visibility: str = "public",
-    game_options: dict = {},
-    **kwargs
+        proto: Protocol,
+        *,
+        mod: str = "faf",
+        visibility: str = "public",
+        game_options: dict = {},
+        **kwargs
```
Askaholic: These lines are over-indented. I think this is a PyCharm / IntelliJ thing, check your autoformatter settings.

tests/integration_tests/test_game.py
```diff
@@ -391,6 +392,55 @@ async def test_game_with_foed_player(lobby_server):
...
+    assert msg["hosted_at"] is None
+    assert msg["launched_at"] is None
+
+    await open_fa(host_proto)
+    await read_until_command(host_proto, "HostGame", target="game")
+
+    msg = await read_until_command(host_proto, "game_info")
+    hosted_at = msg["hosted_at"]
+    assert datetime.fromisoformat(hosted_at) <= datetime_now()
+    assert msg["launched_at"] is None
```
Askaholic: Since this test is about testing the game join log, not the game hosting messages, I think it would be better to use the `host_game` helper here as it will simplify things a bit.
```suggestion
    game_id = await host_game(host_proto)
```

if some of these intermediate assertions are missing from the game host tests we should add them there.

tests/integration_tests/test_game.py
```diff
@@ -391,6 +392,55 @@ async def test_game_with_foed_player(lobby_server):
...
+        result = await conn.execute(
+            select(game_join_log).where(
+                and_(
+                    game_join_log.c.game_id == game_id,
+                    game_join_log.c.player_id == guest_id
+                )
+            )
+        )
+        row = result.one()
+        assert row is not None
```
Askaholic, line 426: What about when the same player joins the same game multiple times? I believe that is specifically the motivation for these changes so it would probably be a good idea to simulate that in the tests here. I would probably just tack that on to the end of this test although you could make a separate test if you really wanted (IMO a bit overkill tho).

You can signal a 'disconnect' from the game by sending `GameState: Ended` GPGnet message:

```python
# Leave and re-join a few times?
await guest_proto.send_message({
    "target": "game",
    "command": "GameState",
    "args": ["Ended"]
})
await join_game(guest_proto, game_id)
```

.github/workflows/test.yml
```diff
@@ -11,7 +11,7 @@ on:
     - cron: '0 0 * * *'
 
 env:
-  FAF_DB_VERSION: v136
+  FAF_DB_VERSION: v137-RC1
```
Askaholic: You will also want to bump the minimum required version in `server/db/__init__.py` otherwise the version check at startup will still be looking for the old version.

Askaholic: And ideally we have a full release before merging this, although I'm not sure what your plan is in terms of the RC.

Brutus5000: if the PR is approved we can release v137 before merging

Brutus5000:
Easier to track in the icebreaker with the pioneer.

.github/workflows/test.yml
BlackYps: v138 that contains the game log table has been released yesterday

PR #1065, opened 2025-09-08
Issue/#1015 Fix pop timer rate calculation

### LINKED ISSUE #1015: Queue pop timer rate calculation is wrong
See #970 

The queue time is supposed to be calculated by looking at the current rate at which players are joining the queue and extrapolating that into the future to predict how much time needs to pass until the desired number of players are in the queue. However, the calculation for the current queue rate is too high as it assumes players are leaving the queue and re-joining between each pop. The code should be rewritten to look at the difference between number of players in queue from when the last pop happened.

AlexChaloner:
This fixes #1015 which notes that the queue pop timer rate is calculated incorrectly.

Fixed by finding the number of new players who have joined the queue - turned out to be pretty simple thanks to some existing logic.

Unit tests:

`pipenv run tests tests/unit_tests -k pop`
image

`pipenv run tests tests/unit_tests`
image

Closes #1015 

---

Some notes on the consequences of this change:

1. This **will** have negative effects on queue waiting times:
    - The pop timer will be given *strictly fewer* people (since it no longer sees people unmatched from the previous pop). This results in longer pop times.
    - The queue matching logic becomes more relaxed the more pops have happened for a person, so individually longer pop times can result in multiplicatively longer waiting times for unmatched players.
    - Combined, this will result in longer wait times for players to be matched. 
        - Two solutions are (a) Lower the maximum time, perhaps from 90 to 60 seconds, or (b) Add extra calculation to take into account the waiting players. e.g. assume they came in at a flat rate over the maximum pop time.
2. This was previously mitigated by adding a minimum pop time of 15 seconds (#1014). As the calculations should now be more stable, we could decrease the minimum pop time further to e.g. 10 or 5 seconds. At peak times, this might be beneficial.

### DIFF
--- a/server/matchmaker/matchmaker_queue.py
+++ b/server/matchmaker/matchmaker_queue.py
@@ -108,6 +108,15 @@ class MatchmakerQueue:
     def num_players(self) -> int:
         return sum(len(search.players) for search in self._queue.keys())
 
+    @property
+    def num_new_queued_players(self) -> int:
+        """ Get the number of players who have joined the queue and not yet had a failed matching attempt in the last pop.
+
+        Returns:
+            int: Number of players who have joined the queue since last matching attempt.
+        """
+        return sum(len(search.players) for search in self._queue.keys() if search.failed_matching_attempts == 0)
+
     async def queue_pop_timer(self) -> None:
         """ Periodically tries to match all Searches in the queue. The amount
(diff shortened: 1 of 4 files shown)


ssiccha:
Hi Alex, are you still looking into this?

PR #1066, opened 2025-09-19
Update Python to 3.13 and dependencies

Brutus5000:


.flake8
Gatsik, line 1: This is not needed, there's already a dedicated section for `flake8` in the `setup.cfg` file

.pre-commit-config.yaml
```diff
@@ -1,19 +1,20 @@
...
     -   id: end-of-file-fixer
     -   id: check-docstring-first
 -   repo: https://github.com/PyCQA/flake8
-    rev: 3.9.2
+    rev: 7.3.0
     hooks:
     -   id: flake8
--   repo: https://github.com/pre-commit/mirrors-autopep8
-    rev: v1.5.7
+        args: ["--config=.flake8"]
```
Gatsik: This is also not needed

.pre-commit-config.yaml
```diff
@@ -1,19 +1,20 @@
...
+-   repo: https://github.com/hhatto/autopep8
```
Gatsik: I suggest to drop this hook entirely, because it lacks some features, including supporting `per-file-ignores` specified in the `flake8` section in `setup.cfg`

Pipfile
```diff
@@ -28,17 +28,18 @@ sortedcontainers = "*"
 sqlalchemy = ">=2.0.0"
 trueskill = "*"
 uvloop = {version = "*", markers = "sys_platform != 'win32'"}
+tzdata = "*"
```
Gatsik, line 30: Insert it in the alphabetical order?

Pipfile
```diff
@@ -28,17 +28,18 @@ sortedcontainers = "*"
...
 
 [dev-packages]
 flaky = "*"
 hypothesis = "*"  # Versions between 6.47.1 and 6.56.4 added a prerelease dependency. See https://github.com/pypa/pipenv/issues/1760
 pdoc3 = "*"
+vulture = "*"
+websockets = "*"
+pytest-asyncio = "*"
```
Gatsik: Again, alphabetical order?

server/gameconnection.py
```diff
@@ -637,26 +637,26 @@ def __str__(self):
...
-    "GameState":            GameConnection.handle_game_state,
-    "IceMsg":               GameConnection.handle_ice_message,
-    "JsonStats":            GameConnection.handle_json_stats,
-    "LaunchStatus":         GameConnection.handle_launch_status,
-    "OperationComplete":    GameConnection.handle_operation_complete,
-    "PlayerOption":         GameConnection.handle_player_option,
-    "Rehost":               GameConnection.handle_rehost,
-    "TeamkillHappened":     GameConnection.handle_teamkill_happened,
-    "TeamkillReport":       GameConnection.handle_teamkill_report,
+    "AIOption": GameConnection.handle_ai_option,
```
Gatsik: I guess this was fixed by `autopep8`, which should be dropped, because `flake8` is being told to ignore this in `setup.cfg`, therefore this change should be reverted

server/protocol/gpgnet.py
```diff
@@ -6,6 +6,7 @@ class GpgNetServerProtocol(metaclass=ABCMeta):
     """
     Defines an interface for the server side GPGNet protocol
     """
+
```
Gatsik: ?

Gatsik (commented):
Since there's no one to stop you anymore, I'm afraid you'll just merge it after bruteforcing the CI, but there are a few things which look a little bit off

Pipfile
Sheikah45: yeah I was just updating packages and was surprised it changed the order of them using basic pipenv

.pre-commit-config.yaml
Brutus5000: I tried to suppress one of the flake8 rules because it gave 100% wrong errors on strings with colons inside.

PR #1070, opened 2025-10-09
Punish the host with a matchmaker violation if game setup times out

BlackYps:
The host might be responsible for other players failing to connect to the game. The host will now receive a matchmaker violation every time the game setup times out. This might punish innocent hosts, but to prevent abuse it is important to acknowledge the possible contribution of the host as well.

## Summary by CodeRabbit

* **Bug Fixes**
  * Improved match launch timeout handling by ensuring all players, including the host, are properly detected as disconnected when a match fails to launch within the timeout window. This enables more accurate failure handling and reporting.

### DIFF
--- a/server/ladder_service/ladder_service.py
+++ b/server/ladder_service/ladder_service.py
@@ -683,7 +683,10 @@ class LadderService(Service):
             raise NotConnectedError([
                 player for player in guests
                 if player not in connected_players
-            ])
+            # The host is included here, because he might be responsible for
+            # a guest being unable to connect, e.g. because they didn't
+            # receive his connect message.
+            ] + [host])
 
     async def get_game_history(
         self,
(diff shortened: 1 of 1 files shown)


Gatsik:
Here's an idea: punish everyone, because everybody can be the cause. This might punish innocent players, but will 100% punish guilty ones.

BlackYps:
Are you unhappy with the original proposal?

Gatsik:
Maybe I'm somewhat unhappy, there are already some weird 'innocent' bans when players seemed to be connected, and this will just add more to that

Like at least checking the number of connected players before including host would somewhat reduce false positives -- if no one else were able to connect (or only few were able) then it is more likely that host is to blame

BlackYps:
I coded this because there was suspicion that some players would maliciously send incorrect packets to other players if they are the host, preventing the game from starting. In this case, checking if the majority of players are connected would not help, because they could send invalid packets to just one player.
Luckily, it turned out that this wasn't actually the case. So it seems this functionality is not needed at the moment. We can add it at a later date, if we ever suspect again that players are abusing the host position.

PR #1071, opened 2025-10-29
Feature/#1032 working veto system

K-ETFreeman:
the work is done, it seems

## Summary by CodeRabbit

* **New Features**
  * Player veto system to let players exclude maps during matchmaking
  * Map versioning added so launched games reference specific map-pool map versions

* **Improvements**
  * Map selection now uses version-aware weights and initial veto-driven weighting
  * Anti-repetition algorithm improved to reduce repeated maps and respect token limits

* **Configuration Updates**
  * Tuned anti-repetition thresholds and repeat weighting for better map variety


server/types.py
```diff
@@ -29,11 +29,22 @@ class GameLaunchOptions(NamedTuple):
...
     game_options: Optional[dict[str, Any]] = None
+    map_pool_map_version_id: Optional[int] = None
+
+
+class MatchmakerQueueMapPoolVetoData(NamedTuple):
+    matchmaker_queue_map_pool_id: int
+    map_pool_map_version_ids: list[int]
+    veto_tokens_per_player: int
+    max_tokens_per_map: float
+    minimum_maps_after_veto: float
```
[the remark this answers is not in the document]
K-ETFreeman: no, because max_tokens_per_map  can be set to float value with calculate_dynamic_max_tokens_per_map

ssiccha:
Hi @K-ETFreeman, I can have a go at reviewing your CR this weekend. I'm a bit busy and might forget about this. If this happens please do ping me on discord (@GuteAlteKanone) next week.

Since I'm new to this repository could you do the following things until the weekend?
- write a high level summary (4 - 5 sentences) of what this PR attempts to do.
  - where in the test cases is the intended behavior tested?
- review the sequence diagram that coderabbit created
  - what is accurate in the sequence diagram, what is irrelevant, is something missing?

K-ETFreeman:
> Hi @K-ETFreeman, I can have a go at reviewing your CR this weekend. I'm a bit busy and might forget about this. If this happens please do ping me on discord (@GuteAlteKanone) next week.
> 
> Since I'm new to this repository could you do the following things until the weekend?
> 
> * write a high level summary (4 - 5 sentences) of what this PR attempts to do.
>   
>   * where in the test cases is the intended behavior tested?
> * review the sequence diagram that coderabbit created
>   
>   * what is accurate in the sequence diagram, what is irrelevant, is something missing?

This PR adds veto system to the server. 
What it does: players now can ban maps they dont like in a matchmaker mappool.
The system is flexible and provides features like partial vetoes or dynamic max_tokens_for_map for each particular match.

i attach the guide to the matchmaker team (its a bit outdated in anti-repetition part tho), so you can better to understand the context
[veto-system-tutorial.html](https://github.com/user-attachments/files/23255986/veto-system-tutorial.html)

tests are written in C:\FreemanProduction\faf\server\tests\integration_tests\test_matchmaker_vetoes.py
and C:\FreemanProduction\faf\server\tests\unit_tests\test_veto_system

diagram looks fine. 
TL;DR; logic in files: 
1) lobbyconnection + players.py -> players set vetoes for themselves
2) ladder_service
       -> server fetches the mappool veto params from servers in update_data, cancels search of the players if some of their veto tokens were removed due to how settings of the mappool was changed
       -> generates initial map weights in _start_game
3) map_pool.py -> takes the initial map weights, applies antirepetition adjustment to it (so players dont play same map twice when it possible / makes sense) -> applies map inner weight (not related to the veto system, map could have inner weights even before this pr) -> chooses the map at random

all the inner logic is hidden in veto_system.py

K-ETFreeman:
Also about # TODO comments (by Askaholic):
He talks about can-we-not-adjust-player-vetoes on the server
I think the answer is NO: client can be modified easily so client checks isnt enough
we just have to have some sorta server validation isnt it

even if we try to remove immediate check (and only check when match found) that will lead to some confusion:
- client fetches the map pool params each time maplist is open
- server fetches the map pool params only each 10 minutes

so if we start to only checking vetoes on match start, this will lead to:
mm team changes mappool settings -> client sees new settings and bans some map -> client finds a match -> his vetoes are cancelled by server (cuz server still has old veto settings) and he plays a map he just banned (as he saw it)

so i think current approach is correct. if you agree with me, free to delete those TODO comments

server/matchmaker/map_pool.py
```diff
@@ -24,39 +26,96 @@ def __init__(
...
+        self._logger.debug("______map_list________________: %s", map_list)
+        final_weights = [adjusted_weights.get(mp_mv_id, 0) * m.weight for mp_mv_id, m in map_list]
+        self._logger.debug("______final_weights________________: %s", final_weights)
+        return random.choices([map for _, map in map_list], weights=final_weights, k=1)[0].get_map()  # nosec B311
 
     def __repr__(self) -> str:
         return f"MapPool({self.id}, {self.name}, {list(self.maps.values())})"
+
+
+class MatchmakerQueueMapPool(NamedTuple):
```
ssiccha, line 114: It's not obvious to me what this class is supposed to do. I think to understand that I'd have to read the full PR first.

It seems to connect a map pool with information that the veto system needs. Why do we need this class? Why is this not either a part of the MapPool or of the MatchmakerQueue itself? 

_edit: I just saw that `matchmaker_queue_map_pool` in `db/models.py` already existed before your PR._

If we think we need this class, then please improve:
- The naming of the class. The current name concatenates the names of two existing classes. But how does this relate to MatchmakerQueue-s? That's not obvious to me. edit: _if this stays the only finding of my review, let's not change the name. If I ask you to change other parts of the CR too and you need to test that anyways, then I'd say let's also rename this thing._
- give it a docstring

ssiccha: I've pasted some parts of your PR into Claude sonnet and asked it to come up with more concrete suggestions. Here's the output. Please double-check whether this even makes sense and is not completely hallucinated. The LLMs are good at generating stuff that sounds convincing but might be subtly or even fundamentally wrong. Often they're subtly wrong and it's easy to miss that.

```suggestion
class QueueMapPoolConfig(NamedTuple):
    """
    Configuration for using a map pool in a matchmaker queue.
    
    Defines rating-based restrictions and veto system parameters for how
    a map pool should be used within a specific queue. A single MapPool
    can be reused across multiple queues with different configurations.
    
    This corresponds to a row in the `matchmaker_queue_map_pool` database table.
    
    Attributes:
        id: Database ID of this configuration (matchmaker_queue_map_pool.id)
        map_pool: The map pool to use
        min_rating: Minimum displayed rating to use this map pool (None = no limit)
        max_rating: Maximum displayed rating to use this map pool (None = no limit)
        veto_tokens_per_player: Number of maps each player can veto
        max_tokens_per_map: Maximum veto tokens that can be applied to any single map
        minimum_maps_after_veto: Minimum number of maps that must remain selectable after vetoes
    """
```

server/db/models.py
```diff
@@ -295,10 +295,14 @@
 
 matchmaker_queue_map_pool = Table(
     "matchmaker_queue_map_pool", metadata,
-    Column("matchmaker_queue_id",   Integer,    ForeignKey("matchmaker_queue.id"),  nullable=False),
-    Column("map_pool_id",           Integer,    ForeignKey("map_pool.id"),          nullable=False),
-    Column("min_rating",            Integer),
-    Column("max_rating",            Integer),
+    Column("id",                       Integer, primary_key=True),
```
ssiccha, line 298: Regarding my earlier comment. I just saw that this was not introduced by your PR but instead already existed in the code base.

I pasted your version of this definition into claude sonnet and prompted to generate a comment for the review, based on my earlier comment. This is its output:

> The table name matchmaker_queue_map_pool makes it look like a simple join table, but it's actually a first-class entity with its own attributes (rating ranges, veto configuration). This deserves a more descriptive name that reflects what it represents in the domain: a configured assignment of a map pool to a queue.
> Consider renaming to something like queue_map_pool_config or map_pool_assignment to better communicate its purpose. At minimum, please add a comment explaining what this table represents.

Code Suggestion:
```
# Configures how a map pool is used within a matchmaker queue.
# Each row represents an assignment of a map pool to a queue with specific
# parameters for rating eligibility and the veto system. A queue can have
# multiple assignments (e.g., different map pools for different rating ranges),
# and a map pool can be assigned to multiple queues with different configurations.
matchmaker_queue_map_pool = Table(
    "matchmaker_queue_map_pool", metadata,
    Column("id",                       Integer, primary_key=True),
    Column("matchmaker_queue_id",      Integer, ForeignKey("matchmaker_queue.id"), nullable=False),
    Column("map_pool_id",              Integer, ForeignKey("map_pool.id"),         nullable=False),
    Column("min_rating",               Integer),  # Minimum rating to use this map pool assignment
    Column("max_rating",               Integer),  # Maximum rating to use this map pool assignment
    Column("veto_tokens_per_player",   Integer, nullable=False),  # Number of maps each player can veto
    Column("max_tokens_per_map",       Integer, nullable=False),  # Max veto tokens applicable to one map
    Column("minimum_maps_after_veto",  Float,   nullable=False),  # Minimum maps remaining after vetoes
)
```

ssiccha: Since this already existed before your PR I'd say let's not rename this, but definitely do add the source comment. First check whether it makes sense though. The LLMs are good at generating stuff that sounds convincing but might be subtly or even fundamentally wrong.

server/matchmaker/map_pool.py
```diff
@@ -24,39 +26,96 @@ def __init__(
...
+    id: int
+    map_pool: MapPool
+    min_rating: Optional[int]
+    max_rating: Optional[int]
+    veto_tokens_per_player: int = 0
+    max_tokens_per_map: float = 0
+    minimum_maps_after_veto: float = 1
```
ssiccha, line 121: This is duplicated with
```
class MatchmakerQueueMapPoolVetoData(NamedTuple):
...
    veto_tokens_per_player: int
    max_tokens_per_map: float
    minimum_maps_after_veto: float
```

Can we get rid of one of them?

ssiccha: First thought: Maybe remove this from MatchmakerQueueMapPoolVetoData and there only keep 
matchmaker_queue_map_pool_id and    map_pool_map_version_ids?

server/ladder_service/veto_system.py
```diff
@@ -0,0 +1,335 @@
...
+from server.exceptions import ClientError
+from server.matchmaker import MatchmakerQueue, MatchmakerQueueMapPool
+from server.player_service import PlayerService
+from server.players import Player
+from server.types import MatchmakerQueueMapPoolVetoData
+
+BracketID = int
+MapPoolMapVersionId = int
+VetoTokensApplied = int
+VetosMap = dict[MapPoolMapVersionId, VetoTokensApplied]
```
ssiccha: nit: consider turning these into classes such that type checking can distinguish between them. Since e.g. the first three are `int`, I assume `mypy` can't distinguish them.

server/ladder_service/veto_system.py
```diff
@@ -0,0 +1,335 @@
...
+
+from server.core import Service
+from server.decorators import with_logger
+from server.exceptions import ClientError
+from server.matchmaker import MatchmakerQueue, MatchmakerQueueMapPool
+from server.player_service import PlayerService
+from server.players import Player
+from server.types import MatchmakerQueueMapPoolVetoData
+
+BracketID = int
```
ssiccha, line 13: what is a bracket? docstring please

server/ladder_service/veto_system.py
```diff
@@ -0,0 +1,335 @@
...
+    def get_vetoes_for_bracket(self, bracket_id: BracketID) -> VetosMap:
+        return self._vetoes.get(bracket_id, {})
+
+    def to_dict(self) -> dict:
+        return {
+            "vetoes": [
+                {
+                    "map_pool_map_version_id": map_id,
+                    "veto_tokens_applied": tokens,
+                    "matchmaker_queue_map_pool_id": bracket_id
```
ssiccha, line 32: so a bracket is an entry in `matchmaker_queue_map_pool`. Document this somewhere please.

server/ladder_service/veto_system.py
```diff
@@ -0,0 +1,335 @@
+import logging
```
ssiccha, line 1: Please add a reference to https://github.com/user-attachments/files/23255986/veto-system-tutorial.html as a module docstring. Also from how I read that document, what we're implementing actually isn't a veto system. A veto is an "unilaterally stopping an action". We don't do that here. Naming it a "veto system" is confusing, I'd change that in the announcement. Ideally also in the code.

This is more of a preference system were players can influence which maps will be made less likely. But they don't have an actual veto vote. Which totally makes sense, given that our matchmaking queues don't have hundreds of players.

server/db/models.py
K-ETFreeman: https://faforever.github.io/db/tables/matchmaker_queue_map_pool.html

this is just database entity
i am not sure why any comment is needed here because no other table has this sorta comments either

server/matchmaker/map_pool.py
K-ETFreeman: yes as you saw later this is just representation of the DB entity
https://faforever.github.io/db/tables/matchmaker_queue_map_pool.html

map_pool too
https://faforever.github.io/db/tables/map_pool.html

the name is intentionally the same to link them, i dont understand why it should be changed
MapPool has the same name as its DB representation, so MatchmakerQueueMapPool should be MatchmakerQueueMapPool as well, not QueueMapPoolConfig

why needed? to not have code repetition
its used in multiple places, if you remove that -> you'll need to write tuple of 7 elements instead of single word of this class

server/matchmaker/map_pool.py
K-ETFreeman: > First thought: Maybe remove this from MatchmakerQueueMapPoolVetoData and there only keep matchmaker_queue_map_pool_id and map_pool_map_version_ids?

no, this wont work
MatchmakerQueueMapPoolVetoData is a data structure that extracts only the veto-relevant information from MatchmakerQueueMapPool. It's used by the VetoService for efficient veto validation and adjustment.

you can look at line 57 in veto_system.py:
```
        pools_vetodata = self.extract_pools_veto_config(queues)

        if self.pools_veto_data != pools_vetodata:
          ... update stuff
```

pools_vetodata is list of MatchmakerQueueMapPoolVetoData
and when we do `self.pools_veto_data != pools_vetodata` we check, if anything veto-related was changed in the pool. If was changed -> we're doing some actions, if not -> we're not

These are fundamental to MatchmakerQueueMapPoolVetoData and cannot be removed. You cannot directly replace it with MatchmaerQueueMapPool either without overcomplicating everything:

> is it possible to replace it with MatchmakerQueueMapPool by any chance? if yes, what fixes that will require to preserve the same behaviour? 
>   ⎿  ⧉ Selected 1 lines from server\ladder_service\veto_system.py in Visual Studio Code
> 
> ● Looking at the code, I can see why MatchmakerQueueMapPoolVetoData exists as a separate structure. Let me analyze if replacement is feasible.
> 
> ● Search(pattern: "self\.pools_veto_data", path: "server\ladder_service\veto_system.py", output_mode: "content")
>   ⎿  Found 5 lines (ctrl+o to expand)
> 
> ● Short answer: It's technically possible but not recommended. The current design is intentional and solves several important problems.
> 
>   Key Issues with Using MatchmakerQueueMapPool
> 
>   1. Equality Comparison Breaks (line 59)
> 
>   if self.pools_veto_data != pools_vetodata:
> 
>   Problem: MatchmakerQueueMapPool contains a MapPool object, which has:
>   - A mutable dict: self.maps = {map_.id: map_ for map_ in maps}
>   - This makes equality comparison unreliable
> 
>   Impact: The change detection would fail because:
>   - Two different MapPool instances are never equal (even with same data)
>   - The condition would always be True, triggering unnecessary updates every time
>   - All players would get forced veto adjustments on every config refresh
> 
>   2. Tuple Unpacking Requires Major Refactoring (line 158)
> 
>   # Current: unpacks 5 fields in specific order
>   for bracket_id, map_ids, total_tokens, max_per_map, _ in self.pools_veto_data:
> 
>   # Would become: unpacks 7 fields, different order, need to extract map IDs
>   for pool_config in self.pools_veto_data:
>       bracket_id = pool_config.id
>       map_ids = [m.map_pool_map_version_id for m in pool_config.map_pool.maps.values()]
>       total_tokens = pool_config.veto_tokens_per_player
>       max_per_map = pool_config.max_tokens_per_map
> 
>   3. Map ID Extraction Everywhere (line 64)
> 
>   # Current: direct access
>   pool_maps_by_bracket = {
>       pool_data.matchmaker_queue_map_pool_id: set(pool_data.map_pool_map_version_ids)
>       for pool_data in self.pools_veto_data
>   }
> 
>   # Would become: repeated extraction
>   pool_maps_by_bracket = {
>       pool_data.id: set(m.map_pool_map_version_id for m in pool_data.map_pool.maps.values())
>       for pool_data in self.pools_veto_data
>   }
> 
>   4. Memory and Complexity
> 
>   - Stores unnecessary fields (min_rating, max_rating)
>   - Keeps full MapPool objects in memory (with all map data)
>   - Creates tight coupling between veto logic and map pool structure
> 
>   What Would Be Required
> 
>   If you really wanted to do this, you'd need:
> 
>   1. Replace equality check with manual field comparison:
>   def _veto_configs_equal(old_pools, new_pools):
>       if len(old_pools) != len(new_pools):
>           return False
>       for old, new in zip(old_pools, new_pools):
>           if (old.id != new.id or
>               old.veto_tokens_per_player != new.veto_tokens_per_player or
>               old.max_tokens_per_map != new.max_tokens_per_map or
>               old.minimum_maps_after_veto != new.minimum_maps_after_veto or
>               set(m.map_pool_map_version_id for m in old.map_pool.maps.values()) !=
>               set(m.map_pool_map_version_id for m in new.map_pool.maps.values())):
>               return False
>       return True
> 
>   if not _veto_configs_equal(self.pools_veto_data, pools_vetodata):
> 
>   2. Update all tuple unpacking
>   3. Extract map IDs everywhere they're needed
>   4. Update type hints and field access patterns
> 
>   Why Current Design is Better
> 
>   The MatchmakerQueueMapPoolVetoData structure is a value object pattern that:
>   - ✅ Is immutable and comparable
>   - ✅ Contains only veto-relevant data
>   - ✅ Enables reliable change detection
>   - ✅ Decouples veto logic from MapPool implementation
>   - ✅ Is lightweight for caching
> 
>   Verdict: The current design is correct. The data duplication is intentional and beneficial. Replacing it would add complexity, reduce
>   reliability, and violate separation of concerns.
>

K-ETFreeman: Basically, MatchmakerQueueMapPool exists because its representation of what we get from the DB (full info about the mappool + his matchmaker-and-veto settings), MatchmakerQueueMapPoolVetoData is distilled version of it which contains only veto-related entities which simplifies a bunch of stuff

server/ladder_service/veto_system.py
K-ETFreeman: You right that mypy can't distinguish them

But these are just for the readability purposes
Can be straight up deleted but reading dict[int, dict[int,int]] can be not very pleasant experience
Same pattern is already used in typedefs.py

turning these into classes - you mean NewType?
i dont see it used anywhere in the codebase, and looks like its an overkill here

more importantly: why its "VetosMap" when it should be "VetoesMap"? the real bug found

server/ladder_service/veto_system.py
K-ETFreeman: yea, bracket is a matchmaker_queue_map_pool because saying matchmaker_queue_map_pool is too long

K-ETFreeman: You right, its not obvious
I guess i'll just replace it with MatchmakerQueueMapPoolId

K-ETFreeman: or maybe not. bracket is simpler. ugh

K-ETFreeman: i guess i just explain the connection in the doc
its "Bracket" because MatchmakerQueueMapPoolId has range of ratings (and maplist), so like "Rating bracket" -> bracket

server/ladder_service/ladder_service.py
```diff
@@ -116,21 +133,32 @@ async def update_data(self) -> None:
...
                 )
         # Remove queues that don't exist anymore
         for queue_name in list(self.queues.keys()):
             if queue_name not in db_queues:
                 self.queues[queue_name].shutdown()
                 del self.queues[queue_name]
 
+        affected_players = self.veto_service.update_pools_veto_config(self.queues)
+        for player in affected_players:
+            self.cancel_search(player)
```
Brutus5000, line 154: Please add a comment. I assume we remove people from the TMM queues if their map vetos settings are incorrect now. Is this mandatory? It is not very user friendly to do so. Could we at least send them a notification?

server/ladder_service/veto_system.py
```diff
@@ -0,0 +1,335 @@
...
+                        "command": "vetoes_info",
+                        "forced": tokens_amount_for_some_map_was_reduced,
+                        **player.vetoes.to_dict(),
+                    })
+                    if tokens_amount_for_some_map_was_reduced:
+                        affected_players.append(player)
+
+            return affected_players
+
+        return []
```
Brutus5000: This can be simplified with 

```python
if self.pools_veto_data pyth == pools_vetodata:
  return []
```

and then the regular code without indentation.

This is called a guard statement and helps reducing complexity.

Brutus5000 (commented):
Even though I don't understand it in detail, this PR looks solid to me.

Please fix the 2 minor things, then I am willing to merge it.
In case further review findings of other reviewers, we can still fix them afterwards.

server/ladder_service/ladder_service.py
K-ETFreeman: > Please add a comment. I assume we remove people from the TMM queues if their map vetos settings are incorrect now. Is this mandatory? It is not very user friendly to do so. Could we at least send them a notification?

We do send them notification!
How this works: 
1) We remove from queue only those players who have some veto tokens removed. F.e. you had 1 token on map, now its 0 (map still in the pool). This done to prevent accidental playing the maps they banned
2) notification looks like this and can be changed in the client if necessary:
image

server/ladder_service/veto_system.py
K-ETFreeman: ok

K-ETFreeman:
[veto-system-developer-guide.html](https://github.com/user-attachments/files/23286058/veto-system-developer-guide.html)

i will put it here for future generations

server/ladder_service/ladder_service.py
K-ETFreeman: Notification happens in veto_system.py:82
Affected players are the same who gets forced: true, and the client reacts on forced: true like (i need to show the notification) (already done in the [Client's PR](https://github.com/FAForever/downlords-faf-client/pull/3428))

server/config.py
```diff
@@ -137,7 +137,9 @@ def __init__(self):
         self.GEO_IP_DATABASE_MAX_AGE_DAYS = 22
 
         self.LADDER_1V1_OUTCOME_OVERRIDE = True
-        self.LADDER_ANTI_REPETITION_LIMIT = 2
+        self.LADDER_ANTI_REPETITION_LIMIT = 1
```
ssiccha, line 140: You said there were issues with this being `2`. Add a unit test which: enforces this being 1 until the underlying issue is fixed. Add a short description to the unit test of what the issue was.

server/db/models.py
ssiccha: Because the other database entities have descriptive names. This name is not helpful. Instead of renaming it, I'd suggest simply adding a source comment what this is for.

server/matchmaker/map_pool.py
ssiccha: Interesting. Thanks for the explanation. :)

server/matchmaker/map_pool.py
ssiccha: > why needed? to not have code repetition
its used in multiple places, if you remove that -> you'll need to write tuple of 7 elements instead of single word of this class

I meant "what purpose does it serve"? :) I think I got it now though.

server/ladder_service/veto_system.py
ssiccha: I mean something like
```
class BracketID(int):
    pass
```

The change is trivial and the advantage is as I said that `mypy` can help you find errors early.

server/ladder_service/veto_system.py
ssiccha: Yeah, I think the whole `matchmaker_queue_map_pool` thing in the DB was badly named in the beginning. As you say, it should have been `Bracket` all along. That makes way more sense. But let's not go ahead and do all kinds of renaming stuff. That's not worth it. Since the documentation you wrote/updated explains this, we're fine.

ssiccha:
> [veto-system-developer-guide.html](https://github.com/user-attachments/files/23286058/veto-system-developer-guide.html)

Do add a link to this to

`server/ladder_service/veto_system.py`. It's a lot less likely that folks will find it in the PR when they're debugging / looking at the veto system code.

PR #1073, opened 2025-11-06
Waive time requirement for ranking custom games when recalling

Hdt80bro:
The time requirement for ranking custom games is in place so that connection issues can be discovered early in the match and rehosted without penalty to rating. However, if a team unanimously agrees to recall during that time frame, it is safe to presume that it was not due to any connection issues and they actually do want to forfeit the match.

This PR introduces a check on the prevalidation step of ranking a game that skips the `TOO_SHORT` path if there wasn't anything that looks like it could have been a connection issue by introducing a `RECALL` army result.

Note that the game itself will still need to be [updated](https://github.com/FAForever/fa/pull/6960) to report such a result; this is only the preparation step to allow the server to handle recalling.

## Summary by CodeRabbit

* **New Features**
  * Added a RECALL outcome to represent early recalls/withdrawals and included its resolution behavior.

* **Bug Fixes**
  * Pre-rate validity now considers per-team outcome indicators together with elapsed time before marking games as too short, reducing false invalidations for early disconnect/recall scenarios.

* **Tests**
  * Added unit coverage for early-recall scenarios and refactored outcome-resolution tests to exercise all outcome permutations.

### DIFF
--- a/server/games/custom_game.py
+++ b/server/games/custom_game.py
@@ -34,7 +34,7 @@ class CustomGame(Game):
         # In other words, only if a team unaminously recalled would we know
         # there weren't any connection issues.
         looks_like_quitting = {ArmyOutcome.DEFEAT, ArmyOutcome.UNKNOWN, ArmyOutcome.CONFLICTING}
-        all_outcomes = {}
+        all_outcomes = set()
         for outcome in team_army_outcomes:
             all_outcomes |= outcome
         possible_conn_issues = len(all_outcomes & looks_like_quitting) > 0
--- a/server/games/custom_game.py
+++ b/server/games/custom_game.py
@@ -31,7 +31,7 @@ class CustomGame(Game):
         # As we only get extremely limited data (the army results), our lens of
         # what can look like "quitting out" is rather large (i.e. the player is
         # "defeated").
-        # In other words, only if a team unaminously recalled would we know
+        # In other words, only if a team unanimously recalled would we know
         # there weren't any connection issues.
         looks_like_quitting = {ArmyOutcome.DEFEAT, ArmyOutcome.UNKNOWN, ArmyOutcome.CONFLICTING}
         all_outcomes = set()
--- a/server/games/custom_game.py
+++ b/server/games/custom_game.py
@@ -3,6 +3,7 @@ import time
 from server.rating import RatingType
 
 from .game import Game
+from .game_results import ArmyOutcome
 from .typedefs import GameType, InitMode, ValidityState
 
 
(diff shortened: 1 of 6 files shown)


server/games/custom_game.py
```diff
@@ -18,9 +19,25 @@ def __init__(self, id, *args, **kwargs):
...
+        # As we only get extremely limited data (the army results), our lens of
+        # what can look like "quitting out" is rather large (i.e. the player is
+        # "defeated").
+        # In other words, only if a team unaminously recalled would we know
+        # there weren't any connection issues.
+        looks_like_quitting = {ArmyOutcome.DEFEAT, ArmyOutcome.UNKNOWN, ArmyOutcome.CONFLICTING}
+        all_outcomes = set()
+        for outcome in team_army_outcomes:
+            all_outcomes |= outcome
+        possible_conn_issues = len(all_outcomes & looks_like_quitting) > 0
```
ssiccha, line 38: Nit: the brackets are not needed here

server/games/custom_game.py
```diff
@@ -18,9 +19,25 @@ def __init__(self, id, *args, **kwargs):
...
-        if not self.enforce_rating and time.time() - self.launched_at < limit:
+        duration = time.time() - self.launched_at
+
+        # As we only get extremely limited data (the army results), our lens of
+        # what can look like "quitting out" is rather large (i.e. the player is
+        # "defeated").
+        # In other words, only if a team unaminously recalled would we know
+        # there weren't any connection issues.
+        looks_like_quitting = {ArmyOutcome.DEFEAT, ArmyOutcome.UNKNOWN, ArmyOutcome.CONFLICTING}
+        all_outcomes = set()
```
ssiccha: Nit: you can just do `set(army_outcomes)`

tests/unit_tests/test_game_resolution.py
```diff
@@ -8,6 +10,26 @@
...
+            with pytest.raises(GameResolutionError):
+                resolve_game(partial_outcomes)
+        else:
+            assert resolve_game(partial_outcomes) == self.resolution
+
+
+ResolveError = ResolutionTest(None)
+ResolveWin = ResolutionTest([GameOutcome.VICTORY, GameOutcome.DEFEAT])
+ResolveDraw = ResolutionTest([GameOutcome.DRAW, GameOutcome.DRAW])
+ResolveLoss = ResolutionTest([GameOutcome.DEFEAT, GameOutcome.VICTORY])
```
ssiccha: These should be snake_case since they're class instances

ssiccha (approved):

server/games/custom_game.py
Hdt80bro: I figured it made logical sense to group `possible_conn_issues ` and the duration test since they're intimately connected, but if it throws people off like it did you, I suppose it should go

Gatsik:
Why all this hassle when [`EnforceRating`](https://github.com/FAForever/fa/blob/fe799c5708b233d1012662e88b6eb39a7088d099/lua/UserSync.lua#L118-L121) exists and is checked here (`not self.enforce_rating`)?

Wouldn't it be way less work to just enforce rating on recall like it is done [when ACU is killed](https://github.com/FAForever/fa/blob/fe799c5708b233d1012662e88b6eb39a7088d099/lua/sim/units/ACUUnit.lua#L110-L111)? And it won't require any server changes at all

Hdt80bro:
Hmm, I didn't know the game code was in charge of that - I thought the server set that property. That might be easier. My question is then, from a design stand-point, should the game or the server ultimately be in charge of if a game is ranked? It seems like it fundamentally shouldn't be part of the game, but maybe the ease-of-use you're suggesting would make it preferable nonetheless.

It should be noted that, at this point, if effort is the limiting factor, then all of the work to make this happen has already been done.

Gatsik:
I guess I just chose words poorly. By 'less work' I mean 'less complex' or 'less changes/diff', and that often does not necessarily mean that there were 'less work/effort' done. In contrast, it often means that more effort were put into research to understand the problem better.
And 'less complex' is obviously better than 'more complex' with less potential problems -- and it looks like it's possible to avoid server changes entirely!

It seems that `EnforceRating` exists purely to bypass server's 'unranked' time limit (when the game is counted as `TOO_SHORT`), and your situation with recall is exactly that!

Hdt80bro:
We'll go ahead with the game-centered approach using `EnforceRating`.

PR #1075, opened 2025-11-23
Refactor/veto stuff

K-ETFreeman:
Removed map_pool_map_version_id from GameLaunchOptions and fixed the tests so they use map name instead
added additional test which checks if map names are unique since its not guaranteed but required for veto system tests

## Summary by CodeRabbit

* **Refactor**
  * Game launches no longer include map version IDs; map selection is conveyed by map name and start position, simplifying per-player launch options.

* **Tests**
  * Updated matchmaker tests to expect map names instead of version IDs.
  * Added a test ensuring each map pool has at least one map and that map names are unique within each pool.

<sub>✏️ Tip: You can customize this high-level summary in your review settings.</sub>

### DIFF
--- a/server/ladder_service/ladder_service.py
+++ b/server/ladder_service/ladder_service.py
@@ -639,7 +639,7 @@ class LadderService(Service):
                     game_options=game_options,
                     team=game.get_player_option(player.id, "Team"),
                     faction=game.get_player_option(player.id, "Faction"),
-                    map_position=game.get_player_option(player.id, "StartSpot")
+                    map_position=game.get_player_option(player.id, "StartSpot"),
                 )
 
             await self.launch_match(game, host, all_guests, make_game_options)
--- a/server/ladder_service/ladder_service.py
+++ b/server/ladder_service/ladder_service.py
@@ -639,8 +639,7 @@ class LadderService(Service):
                     game_options=game_options,
                     team=game.get_player_option(player.id, "Team"),
                     faction=game.get_player_option(player.id, "Faction"),
-                    map_position=game.get_player_option(player.id, "StartSpot"),
-                    map_pool_map_version_id=game_map.map_pool_map_version_id
+                    map_position=game.get_player_option(player.id, "StartSpot")
                 )
 
             await self.launch_match(game, host, all_guests, make_game_options)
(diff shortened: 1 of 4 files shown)


server/ladder_service/ladder_service.py
```diff
@@ -639,8 +639,7 @@ def make_game_options(player: Player) -> GameLaunchOptions:
                     game_options=game_options,
                     team=game.get_player_option(player.id, "Team"),
                     faction=game.get_player_option(player.id, "Faction"),
-                    map_position=game.get_player_option(player.id, "StartSpot"),
-                    map_pool_map_version_id=game_map.map_pool_map_version_id
+                    map_position=game.get_player_option(player.id, "StartSpot")
```
Gatsik: you don't have to remove trailing comma -- python supports that, and the diff will look cleaner

tests/integration_tests/test_matchmaker.py
```diff
@@ -59,8 +57,7 @@ async def test_game_launch_message(lobby_server):
         "team": 2,
         "faction": 1,
         "expected_players": 2,
-        "map_position": 1,
-        "map_pool_map_version_id": mpmv_id
+        "map_position": 1
```
Gatsik: removing trailing comma does nothing

tests/integration_tests/test_matchmaker_vetoes.py
```diff
@@ -10,6 +10,23 @@
 )
 
 
+async def test_used_pools_have_some_maps_and_all_map_names_are_unique(ladder_service):
+    used_map_pools = {}
+    for queue in ladder_service.queues.values():
+        for mq_map_pool in queue.map_pools.values():
+            pool_id = mq_map_pool.map_pool.id
+            if pool_id in [1, 2, 3]:
```
Gatsik, line 18: on your first attempt there were 4 pool ids, and tests failed. so you reduced their amount to 3. why? wasn't the failure legitimate?

K-ETFreeman: error was not due to that
pool 4 just is not used in any veto tests so it doesnt care about its content

PR #1086, opened 2026-05-24
Send structured banned payload for localized client handling

### LINKED ISSUE #640: Replace ban text with localizable message

ZainKazmiii:
Fixes #640

## Summary
- replace hardcoded English ban notice text with a structured `banned` payload
- send UTC ISO-8601 `expires_at` plus `reason` so clients can localize presentation
- keep ban enforcement behavior unchanged (server still aborts banned sessions)

## Validation
- `python -m py_compile server/exceptions.py server/lobbyconnection.py tests/integration_tests/test_login.py tests/integration_tests/test_server.py tests/unit_tests/test_lobbyconnection.py`
- `python -m pytest tests/unit_tests/test_lobbyconnection.py -k abort_connection_if_banned -q` (blocked in this runner: `No module named pytest`)

## Test updates
- switched ban-related integration/unit expectations from HTML `notice` text to the structured `banned` payload format

## Summary by CodeRabbit

* **Bug Fixes**
  * Ban notifications are now a structured "banned" response with an ISO‑8601 UTC expires_at and a separate reason for clearer, machine‑readable ban info.
  * Lobby connections now abort with a consistent ban-block reason and send the structured ban payload to clients.
  * Message packing now accepts additional buffer types for more robust serialization.

* **Tests**
  * Integration and unit tests updated to validate the new structured ban payload and serialization behavior.

[Review Change Stack](https://app.coderabbit.ai/change-stack/FAForever/server/pull/1086?utm_source=github_walkthrough&utm_medium=github&utm_campaign=change_stack)

### DIFF
--- a/Pipfile.lock
+++ b/Pipfile.lock
@@ -1,7 +1,7 @@
 {
     "_meta": {
         "hash": {
-            "sha256": "67adacda446396d91de50fe41c426c738226113aa0b3e3f16a00f2141ff7bc8a"
+            "sha256": "a187fb4a62ec0cc2dab78542af7c21eb559388bac49e0f84f35f656aa1e19200"
         },
         "pipfile-spec": 6,
         "requires": {
--- a/server/exceptions.py
+++ b/server/exceptions.py
@@ -4,6 +4,7 @@ Common exception definitions
 
 from datetime import timezone
 
+
 class ClientError(Exception):
     """
     Represents a protocol violation by the client.
--- a/server/exceptions.py
+++ b/server/exceptions.py
@@ -2,10 +2,7 @@
 Common exception definitions
 """
 
-import humanize
-
-from server.timing import datetime_now
-
+from datetime import timezone
 
 class ClientError(Exception):
     """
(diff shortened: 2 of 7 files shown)


ZainKazmiii:
Quick note on the latest CodeRabbit qdatastream nitpick:

I intentionally kept `pack_block` accepting `bytes | bytearray` and kept the call site unchanged because a prior bytes-conversion variant caused the `qstream` matchmaker test path to fail on this PR line.

Current head `7db962f1630329d1bbe39e015c458fa7eb998f74` is green across checks after restoring that behavior. If maintainers want a stricter immutable-bytes contract, I can do that in a follow-up that preserves runtime behavior and keeps CI green.

ZainKazmiii:
Addressed in 7e722a6. I clarified the pack_block contract in qdatastream.py: it intentionally accepts ytes | bytearray because pack_message builds a mutable ytearray buffer and passes it through. I kept behavior unchanged here because forcing ytes(msg) previously correlated with the qstream regression we just fixed.

ZainKazmiii:
Follow-up on the qdatastream immutability discussion: this is now addressed in head commit `41d82aa`.

- `pack_message` now calls `pack_block(bytes(msg))`
- `pack_block` now requires `bytes`

This keeps the contract explicit and aligns with the CodeRabbit note. Current PR checks are green on this head.

ZainKazmiii:
@coderabbitai review

Current head (41d82aa) now calls pack_block(bytes(msg)) and keeps pack_block(block: bytes), so the immutability nitpick should be resolved on latest commit.

PR #1090, opened 2026-06-11
Add localizable fields to ban notices

### LINKED ISSUE #640: Replace ban text with localizable message

2ws2khc2zh-rgb:
## Summary
- Adds structured localization fields to ban notice messages.
- Keeps the existing text fallback for backwards compatibility with current clients.
- Includes the ban reason and ISO-formatted expiry time so clients can render localized ban text themselves.

## Validation
- [x] `python -m py_compile server\exceptions.py server\lobbyconnection.py tests\integration_tests\test_login.py`
- [ ] `python -m pytest tests\integration_tests\test_login.py -k "ban"` could not be run locally because pytest is not installed in this environment.

Closes #640

## Summary by CodeRabbit

* **Bug Fixes**
  * Server ban notifications now include structured details: localization support, ban reason, and ban expiration timestamp, giving users clearer, actionable context when access is restricted.

### DIFF
--- a/server/exceptions.py
+++ b/server/exceptions.py
@@ -39,6 +39,16 @@ class BanError(Exception):
             "moderation@faforever.com</i>"
         )
 
+    def to_notice(self):
+        return {
+            "command": "notice",
+            "style": "error",
+            "text": self.message(),
+            "localization_key": "ban.error",
+            "ban_reason": self.ban_reason,
+            "ban_expires_at": self.ban_expiry.isoformat(),
+        }
+
     def _ban_duration_text(self):
         ban_duration = self.ban_expiry - datetime_now()
         if ban_duration.days > 365 * 100:
--- a/server/exceptions.py
+++ b/server/exceptions.py
@@ -40,6 +40,9 @@ class BanError(Exception):
         )
 
     def to_notice(self):
+        """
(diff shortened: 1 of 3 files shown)


2ws2khc2zh-rgb:
Thanks CodeRabbit. I added a docstring for the new `BanError.to_notice()` helper to address the docstring coverage warning.

Validation still passes with:

```text
python -m py_compile server/exceptions.py server/lobbyconnection.py tests/integration_tests/test_login.py
```

PR #1091, opened 2026-06-11
fix(tests): make integration tests pass on macOS

Brutus5000:
## Summary

On macOS with `KqueueSelector`, `asyncio.open_connection(...)` returns before the server-side accept callback has registered the new connection in `ctx.connections`. Three tests in `tests/integration_tests/test_servercontext.py` race this scheduling difference — they pass on Linux CI but fail deterministically on macOS.

- `test_unexpected_exception` — added `await exhaust_callbacks()` so the server task emits the "Exception in protocol" log before `caplog.text` is asserted.
- `test_drain_connections` — poll `ctx.connections` until populated before calling `drain_connections`. Cannot use `exhaust_callbacks` inside `@fast_forward`: the outer ticker keeps rescheduling itself and the inner exhaust never exits.
- `test_connection_broken_external` — same poll, otherwise `next(iter(ctx.connections.values()))` raised `StopIteration`.

No production code is changed. Full integration suite (`tests/integration_tests`) goes from 3 fail / 323 pass to 326 pass on macOS (Darwin 24.6.0, Python 3.13). Linux behavior is unchanged — the added waits are no-ops when the server task has already run.

## Test plan
- [ ] CI passes on Linux (no regression)
- [ ] `pipenv run py.test tests/integration_tests --mysql_database=faf` passes on macOS

🤖 Generated with [Claude Code](https://claude.com/claude-code)

## Summary by CodeRabbit

* **Tests**
  * Improved determinism and reliability of server connection handling tests by ensuring tests wait for active connections before proceeding, reducing flakiness.
  * Enhanced matchmaker-related tests to wait for expected queue information before validating boundary values, making validations more robust and stable in CI.

### DIFF
--- a/tests/integration_tests/test_servercontext.py
+++ b/tests/integration_tests/test_servercontext.py
@@ -91,6 +91,8 @@ async def test_connection_broken_external(context):
     """
     srv, ctx = context
     _, writer = await asyncio.open_connection(*srv.sockets[0].getsockname())
+    while not ctx.connections:
+        await asyncio.sleep(0)
     writer.close()
     # Need this sleep for test to work, otherwise closed protocol isn't detected
     await asyncio.sleep(0)
@@ -119,6 +121,7 @@ async def test_unexpected_exception(context, caplog, mocker):
 
     with caplog.at_level("TRACE"):
         _, writer = await asyncio.open_connection(*srv.sockets[0].getsockname())
+        await exhaust_callbacks()
 
     with closing(writer):
         assert "Exception in protocol" in caplog.text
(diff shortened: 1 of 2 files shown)


Brutus5000:
Addressed CodeRabbit's actionable comment in 5c2831e7 — both `while not ctx.connections` spin loops now go through a `wait_for_connection_registered(ctx)` helper that bounds the wait to 1000 yields and raises `AssertionError` on exhaustion, so a regression in server-side connection registration fails fast instead of hanging the suite.

The remaining lint failures are pre-existing on `develop` and unrelated to this PR:

- **mypy** — `server/protocol/qdatastream.py:95` `bytearray` vs `bytes` mismatch. This file is untouched here.
- **pipenv-verify** — `Pipfile.lock is out-of-date`. The lockfile hash drift is on `develop` itself; my branch's `Pipfile`/`Pipfile.lock` match `develop` exactly (`git diff origin/develop -- Pipfile Pipfile.lock` is empty).

Happy to address either in a follow-up if you'd like them bundled.

PR #1093, opened 2026-06-13
feat(ws): native WebSocket transport, drop raw TCP listeners

Brutus5000:
## Summary

- The lobby server now accepts WebSocket connections directly (default `:8003/`, one JSON object per text frame, `\n`-terminated for client framing compatibility) instead of raw TCP behind the `ws_bridge_rs` sidecar.
- Removes the proxy-protocol path and the legacy `LISTEN` config. New config:
  - `WS_HOST` (default `""`)
  - `WS_PORT` (default `8003`)
  - `WS_PATH` (default `"/"` — matches what `faf-commons-lobby` connects to)
  - `WS_FORWARDED_IP_HEADER` (default `"X-Real-IP"`; set to `""` to opt out, or to `"CF-Connecting-IP"` to use Cloudflare's header)
- `QDataStreamProtocol` / `SimpleJsonProtocol` modules are kept (encoding helpers / unit tests still use them) but are no longer wired into a listener.
- Motivation: the `ws_bridge_rs` → TCP hop was breaking client reconnect, and operationally we can't expose raw TCP anymore (DDoS). Native WS keeps everything on HTTPS behind Traefik / Cloudflare.

## Client compatibility

`faf-commons-lobby` (used by the Java client) connects to `wss://lobby.{base}/` and splits the inbound WS byte stream on `\n` rather than treating each frame as a message. So:

1. The default path is `/` (a `/ws` default caused aiohttp to fail routing on the upgrade).
2. Outgoing frames are terminated with `\n` to match what the client splits on. Mirrors how `SimpleJsonProtocol` framed messages before — the bridge passed those bytes through unchanged, which is why it worked before.

## Forwarded-IP header

Only honored when `WS_FORWARDED_IP_HEADER` is explicitly set; otherwise the direct TCP peer is used. Stops direct clients from spoofing their peername by sending `X-Forwarded-For` themselves.

## Follow-up (gitops-stack)

- Drop `apps/faf-ws-bridge/`.
- Point the IngressRoute for `lobby.{baseDomain}` (and `ws.{baseDomain}` for back-compat) directly at `faf-lobby-server:8003`.
- Set `WS_FORWARDED_IP_HEADER: CF-Connecting-IP` in the lobby config so Cloudflare's client IP is preserved.

## Test plan

- [x] `pipenv run pytest tests/unit_tests/` — 692 pass (1 pre-existing DB-schema deselect, unrelated)
- [x] `pipenv run pytest tests/integration_tests/ -m 'not rabbitmq'` — 154 pass
- [x] New `WebSocketProtocol` unit tests at 100% line coverage in isolation (binary frames, write_message/write_messages paths, drain edge cases, abort)
- [x] Live smoke: connect with `aiohttp.ClientSession().ws_connect()`, send/receive round-trip
- [x] Real Java client (faf-commons-lobby) connects via `wss://lobby.faforever.xyz` against the test deploy

🤖 Generated with [Claude Code](https://claude.com/claude-code)

## Release Notes

* **New Features**
  * Server now exclusively uses WebSocket protocol on port 8003.

* **Removed Features**
  * Removed legacy protocol support and proxy pass-through functionality.

* **Configuration**
  * Updated default configuration with WebSocket-specific settings including host, port, and path.

* **Infrastructure**
  * Updated Kubernetes deployment to expose only WebSocket endpoint.

### DIFF
--- a/Pipfile
+++ b/Pipfile
@@ -21,7 +21,6 @@ humanize = ">=2.6.0"
 maxminddb = "*"
 oauthlib = "*"
 prometheus_client = "*"
-proxy-protocol = "*"
 pyjwt = {version = ">=2.4.0", extras = ["crypto"]}
 pyyaml = "*"
 sortedcontainers = "*"
--- a/Pipfile.lock
+++ b/Pipfile.lock
@@ -1,7 +1,7 @@
 {
     "_meta": {
         "hash": {
-            "sha256": "459f11a21d4df5330c32da06d77ec5ad501bbd95b911fe638b1b371943bbf01c"
+            "sha256": "52d5a482382a456a1c616ce3c0cd5c18fb3d96960802477509c61cb171fa4697"
         },
         "pipfile-spec": 6,
         "requires": {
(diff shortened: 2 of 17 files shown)


Brutus5000:
I can't test a game (because setup on mac is not working) but tested the broadcast message, the websocket connection works.

PR #1104, opened 2026-09-06
Feature/#1103 Accept replay review requests and publish them to the broker

### LINKED ISSUE #1181: Replay review requests from the client, via the api and RabbitMQ
**Goal in one sentence:** a player presses "request replay review" in the client, and the request appears as a post in the training Discord's replay review forum channel, attributed to a FAF account the server vouched for.

This writes up Brutus5000's suggestion from Discord: lobby command → RabbitMQ → faf-qai consumes and posts. It replaces two earlier ideas that were rejected for good reasons — a webhook in the client, which cannot verify who is asking, and an HTTP write endpoint on faf-qai, which adds inbound attack surface to a service that is not closely maintained.

## Why this shape

- **Identity is free.** The client is already authenticated on the lobby socket. The server knows who is sending, so nothing has to trust a client-supplied player id.
- **No new inbound surface.** faf-qai consumes from the broker. It opens no port. This was the blocking objection to the HTTP variant.
- **Kill switch.** Unbind the queue and the feature is off — no client release, no code change, no restart.
- **Durability.** A qai restart does not lose requests; they wait in the queue.

## Step 1 — Lobby server (this repo)

A `command_request_replay_review` handler on `LobbyConnection`, following the convention of `command_invite_to_party` and `command_set_player_vetoes`. It:

1. Requires an authenticated connection (the existing `ensure_authenticated` gate covers this — no new mechanism).
2. Validates the payload against a fixed schema and rejects anything else.
3. **Stamps `player_id` and `login` from the connection, not from the body.** The client supplies content fields only. This is the property the whole design rests on; without it the impersonation hole would be back by another route.
4. Rate limits per player and rejects over-limit requests *before* publishing, so abuse never reaches the bus.
5. Publishes to `MQ_EXCHANGE_NAME` with routing key `request.replay_review.create`.

Payload:

| field | source | notes |
|---|---|---|
| `player_id`, `login` | **server** | from the connection |
| `replay_id` | client | id only, never a url — see below |
| `map`, `game_mode`, `faction`, `rating`, `played_at` | client | prefilled by the client, editable by the player |
| `goal` | client | required, free text |
| `struggle` | client | optional, free text |
| `requested_at` | server | ISO 8601 UTC |

Two deliberate choices in that table:

**Structured fields, not rendered Markdown.** Formatting belongs at the destination, so changing how a review post reads does not need a client release, and a second consumer can use the same request differently.

**A replay is named by id only.** The client also knows how to name a replay by link or by a local file. Neither belongs on the bus: a link would let the client decide what a review post points at, and a file nobody else can fetch is not a review request. Those two cases keep using the client's existing copy-and-paste path.

Routing key follows the `request.*` / `success.*` convention already used by `client_message_queue_service.py` and `avatar_change_queue_service.py`.

## Step 2 — Rate limit

There is no per-player rate limiter in the lobby server today, so this is new code. Proposal: one request per player per 24h, as a config constant next to the other tunables, held in memory in a small service.

In-memory means it does not survive a restart and each instance counts on its own — a spam guard rather than a quota. That looks like the right trade: the cost of a duplicate is a duplicate post, and the alternative is a schema migration plus a database write on a path that has no other reason to touch the database. Happy to be argued out of it.

## Step 3 — Broker

A queue for faf-qai bound to `request.replay_review.create` on `faf-rabbitmq`. Ops task, not a code change.

## Step 4 — faf-qai

A consumer on that queue that opens a forum post per request. DSharpPlus supports forum posts natively, so no REST workarounds. It renders the structured payload and writes the verified player name, so the author shown in Discord is one the server vouched for.

If the requesting player has linked their Discord account (`AccountLinkService`), a later version can mention them so the answer reaches them. Not in the first version.

## Step 5 — Client

One frame alongside the existing lobby commands. The existing copy-and-paste path stays as the fallback for a player who is offline, for a replay that has no id, and for the case where any part of this is switched off — so nothing regresses if the feature is disabled.

## Step 6 — Rollout

Ship the lobby side first with no consumer bound: the messages go nowhere and cost nothing. Bind qai's queue when both sides are deployed. The only hard ordering is lobby-before-client.

## Decisions needed

1. **Routing key name** — is `request.replay_review.create` right?
2. **Rate limit** — one per player per day, and is the in-memory spam guard acceptable, or does this want to be durable?
3. **Who creates the queue and binding?**
4. **Is faf-qai's maintainer on board, and does the bot have permission to create posts in that channel?** This one is outside both repos and it is what killed the Dostya attempt, so I would rather ask now than find out later.

A fifth, smaller one: the lobby currently sends nothing back on success and answers a rejection with the existing `notice` path, the same as `command_invite_to_party`. That means the client shows success optimistically. If an explicit acknowledgement is wanted instead, that is a new server→client command and worth deciding before clients are written against it.

## Not in scope

The generic passthrough — a mechanism letting any client message reach the bus without a per-command change. Worth building, but it is the harder problem: it puts unvalidated client input on a bus whose producers are trusted precisely because it is internal only (see the docstring in `client_message_queue_service.py`). One command with a fixed routing key and a validated schema is the safe increment, and it is a concrete first candidate to migrate once the general mechanism exists.

Fan-out itself already works: the topic exchange lets any number of services bind their own queue to the same key, which `ClientMessageQueueService` and `AvatarChangeQueueService` already do side by side. So "multiple services consuming" needs nothing new.

## What I am offering

I have written the lobby server side with tests, and the qai consumer. I need review and a deploy, not implementation time.

TimMasalme:
Closes FAForever/faf-java-api#1181

Adds a `request_replay_review` lobby command. It validates the content fields of a replay review request, stamps the requesting player from the authenticated connection, and publishes the result to the topic exchange with routing key `request.replay_review.create`.

The lobby neither answers the request nor knows who will. A consumer bound to that key turns it into a post wherever reviews are actually answered — the intended first one is a forum post by faf-qai, which is a separate PR and not required for this one to be safe to merge.

## Why this goes through the lobby at all

Identity. A client posting straight to a chat service can claim to be anyone, which is what sank the earlier proposals. A client on this socket has already authenticated, so `player_id` and `login` are the server's word rather than the client's. Nothing else in the feature needs the lobby.

## What is in here

- `server/replay_review_service.py` — validation, the rate limit, and the publish. The wire contract is documented in the module docstring.
- `command_request_replay_review` on `LobbyConnection`, next to the other party/veto commands.
- `REPLAY_REVIEW_COOLDOWN_SECONDS` in `config.py`.

## Three decisions worth looking at

**The client cannot name a replay by url.** The payload takes `replay_id` only. The client also knows how to name a replay by link or by a local file, but a link would let the client decide what a review post points at, and a file nobody else can fetch is not a review request. Those keep using the client's existing copy-and-paste path.

**Identity is dropped at the parser, not overwritten later.** `parse_review_request` returns content fields only, and the caller adds `player_id`/`login`. There is deliberately no ordering of operations in which a client-supplied value survives — `test_submit_ignores_identity_the_client_tried_to_set` pins that end to end.

**The rate limit is in memory.** One request per player per 24h, per lobby instance, lost on restart. That makes it a spam guard rather than a quota. The cost of a duplicate is a duplicate post, and the alternative is a schema migration plus a database write on a path that has no other reason to touch the database. Rejections happen before publishing, so an over-limit request never reaches the bus. Happy to change this if you would rather have it durable — it is contained in one service.

## Deploying this alone is a no-op

With no queue bound to the routing key, published messages go nowhere and cost nothing. That makes the rollout order lobby-first, and unbinding the queue later is the off switch — no client release, no code change.

## Not covered

The lobby sends nothing back on success and answers a rejection through the existing `notice` path, the same as `command_invite_to_party`, so a client shows success optimistically. If an explicit acknowledgement is wanted that is a new server→client command, and better decided before clients are written against it. Noted as the fifth question in FAForever/faf-java-api#1181.

## Tests

- `tests/unit_tests/test_replay_review_service.py` — parsing, the identity stamp, the rate limit, and that a rejected request never reaches the broker.
- `tests/unit_tests/test_lobbyconnection.py` — the command wiring, including that an unauthenticated connection never reaches the handler.

`flake8` and `isort` are clean on the changed files. The suite itself I could not run locally — it wants Python 3.13 plus MySQL and RabbitMQ — so CI here is the first full run.

## Summary by CodeRabbit

* **New Features**
  * Added a replay review request flow from authenticated lobby players.
  * Replay review submissions are validated, sanitized, and queued for processing.
  * Added protection against repeated submissions with a one-day per-player cooldown.
  * Server-assigned identity and timestamps help ensure request information is trustworthy.

* **Bug Fixes**
  * Invalid or unauthenticated replay review requests now receive appropriate error responses.

### DIFF
--- a/server/__init__.py
+++ b/server/__init__.py
@@ -136,6 +136,7 @@ from .oauth_service import OAuthService
 from .party_service import PartyService
 from .player_service import PlayerService
 from .rating_service.rating_service import RatingService
+from .replay_review_service import ReplayReviewService
 from .servercontext import ServerContext
 from .stats.game_stats_service import GameStatsService
 
@@ -160,6 +161,7 @@ __all__ = (
     "PlayerService",
     "RatingService",
     "RatingService",
+    "ReplayReviewService",
     "ServerInstance",
     "VetoService",
     "ViolationService",
@@ -210,6 +212,7 @@ class ServerInstance(object):
             rating_service=self.services["rating_service"],
             oauth_service=self.services["oauth_service"],
             veto_service=self.services["veto_service"],
+            replay_review_service=self.services["replay_review_service"],
         )
 
     def write_broadcast(
--- a/server/config.py
+++ b/server/config.py
@@ -102,6 +102,11 @@ class ConfigurationStore:
         # How many seconds a connection has to authenticate before being killed
         self.LOGIN_TIMEOUT = 5 * 60
 
+        # How long a player must wait between replay review requests. Kept in
+        # memory per lobby instance, so this is a spam guard rather than a
+        # quota; see `ReplayReviewService`.
+        self.REPLAY_REVIEW_COOLDOWN_SECONDS = 24 * 60 * 60
+
         self.NEWBIE_BASE_MEAN = 500
         self.NEWBIE_MIN_GAMES = 10
         self.START_RATING_MEAN = 1500
--- a/server/lobbyconnection.py
+++ b/server/lobbyconnection.py
@@ -61,6 +61,7 @@ from .players import Player, PlayerState
 from .protocol import DisconnectedError, Protocol
 from .rating import InclusiveRange, RatingType
 from .rating_service import RatingService
+from .replay_review_service import ReplayReviewService, parse_review_request
 from .types import Address, GameLaunchOptions
 
 
@@ -111,6 +112,7 @@ class LobbyConnection:
         rating_service: RatingService,
         oauth_service: OAuthService,
         veto_service: VetoService,
+        replay_review_service: ReplayReviewService,
     ):
         self._db = database
         self.geoip_service = geoip
@@ -121,6 +123,7 @@ class LobbyConnection:
         self.rating_service = rating_service
         self.oauth_service = oauth_service
         self.veto_service = veto_service
+        self.replay_review_service = replay_review_service
         self._authenticated = False
         self.player: Optional[Player] = None
         self.game_connection: Optional[GameConnection] = None
@@ -1422,6 +1425,24 @@ class LobbyConnection:
             vetoes[matchmaker_queue_map_pool][map_pool_map_version_id] = veto_tokens_applied
 
         await self.veto_service.set_player_vetoes(self.player, vetoes)
+
+    async def command_request_replay_review(self, message):
+        """
+        Ask for a replay of yours to be reviewed by someone.
+
+        The lobby does not answer the request and does not know who will. It
+        validates the content, stamps the sender from this connection, and puts
+        the result on the message queue for whichever service is bound to it.
+
+        The identity stamp is the whole reason this goes through the lobby.
+        Anywhere else a client could name any player it liked; here the socket
+        has already been authenticated, so `player_id` and `login` are the
+        server's word rather than the client's.
+        """
+        assert self.player is not None
+
+        request = parse_review_request(message)
+        await self.replay_review_service.submit(self.player, request)
 
     async def send_warning(self, message: str, fatal: bool = False):
         """
--- a/server/replay_review_service.py
+++ b/server/replay_review_service.py
@@ -1,0 +1,208 @@
+"""
+Accept replay review requests from clients and publish them to the broker.
+
+# Wire contract
+The service publishes to the `MQ_EXCHANGE_NAME` topic exchange with routing
+key `request.replay_review.create`. The body is a UTF-8 JSON object:
+
+- `player_id` (int): the requesting player. **Stamped from the authenticated
+  connection**, never taken from the client message.
+- `login` (str): that player's name, stamped from the same place.
+- `replay_id` (int): the replay to review.
+- `map`, `game_mode`, `faction`, `rating`, `played_at` (str, may be empty):
+  context the reviewer sees before opening the replay.
+- `goal` (str, non-empty): what the player wants out of the review.
+- `struggle` (str, may be empty): what went wrong from the player's view.
+- `requested_at` (str): ISO 8601 UTC timestamp of when the lobby accepted it.
+
+Consumers render this into whatever their destination wants. The lobby does
+not compose prose, so the shape of a review post can change without a client
+release, and a second consumer can use the same request differently.
+
+# Why the lobby is involved at all
+The point of the detour is identity. A client posting straight to a chat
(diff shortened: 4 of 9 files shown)


TimMasalme:
On the red Codacy check, in case it looks like something to fix: it has both `D212` and `D213` enabled, which are mutually exclusive alternatives in pydocstyle, so no multi-line docstring can pass. Reproduced locally with the same two rules selected:

```
$ printf '"""Summary on the first line.\n\nBody.\n"""\n' > a.py
$ pydocstyle --select=D212,D213 a.py
a.py:1 at module level:
        D213: Multi-line docstring summary should start at the second line

$ printf '"""\nSummary on the second line.\n\nBody.\n"""\n' > b.py
$ pydocstyle --select=D212,D213 b.py
b.py:1 at module level:
        D212: Multi-line docstring summary should start at the first line
```

The one escape is a single-line docstring, which triggers neither. I have used that where it costs nothing — the constructor and the one test that was flagged. That leaves a single annotation, on the module docstring of `replay_review_service.py`, where the body is the wire contract. Collapsing that to a comment would take it out of the generated docs, and it would stop matching `client_message_queue_service.py` and `avatar_change_queue_service.py`, which document their contracts the same way. Happy to change it if you would rather have the check green.

Every other check passes, including unit tests, mypy, flake8, isort and codecov.

TimMasalme:
Thanks — took one of the two.

**Cooldown claimed before the publish**: done in 29f3a00, with a rollback when the publish raises and two regression tests. The race itself is narrow, since `ServerContext` awaits `on_message_received` one message at a time so a single connection cannot overlap with itself, but the ordering costs nothing and the rollback is worth having on its own — a request that never reached the broker should not cost the player their turn.

**Coercing the config value to `int`**: leaving this one. `ConfigurationStore.refresh` says `# NOTE: Only works for string values!` at the point where it reads the environment, and that applies to every numeric setting in the file — `LOGIN_TIMEOUT` set through the environment would break `asyncio.sleep` the same way. Making this one constant the only one that coerces would hide a repo-wide limitation behind a single exception rather than fix it. Deployments configure through `CONFIGURATION_FILE`, and YAML keeps the type. Happy to send a separate PR that coerces numeric defaults across the board if that is wanted.

Sheikah45:
This functionality should go into the api elide model as discussed in https://github.com/FAForever/faf-java-api/issues/1181

